PR #87, opened 2014-12-24
Add BMI for stream power

mcflugen:
This is an initial implementation of a Basic Modeling Interface for the stream_power component.


gregtucker:
The code looks great to me. I hadn't known about the tempfile module before.

Github claims that there are conflicts, which I can't find. @mcflugen, do you know where the conflict is?

This example makes me realize that the BMI'ing process is not as straightforward as I had thought. To what extent can we "build in" BMI compatibility into a master class (e.g., Component or its base)?

mcflugen:
OK. Thanks. I'll fix the conflicts and merge.

A lot of this is boilerplate code. Because we have standardized a set of attributes for landlab components (`_input_var_names`, `_var_units`, etc.) we can probably move much of this code elsewhere. Probably into a base class (like the `Component` class) that all landlab BMI components are built from.

gregtucker:
@mcflugen: this pull request was assigned to me, er, over a year ago... is it still valid? Superceded?

mcflugen:
@gregtucker: Yeah. Because we now automatically wrap components with a BMI, this pull request doesn't make much sense anymore. I'm just going to close it.

PR #129, opened 2015-07-13
Fix issue 128

### LINKED ISSUE #128: Standardize naming convention for grid elements
We've decided on a new naming convention for referring to the elements of a grid. We now just need to implement the convention. 

The old way:
* cell_node

The new way:
* node_at_cell

mcflugen:
This pull request is intended to fix #128, which standardizes the naming conventions of grid elements. Things aren't complete yet but I thought I would open up a pull request that we can all comment on.


mcflugen:
One issue I've come across is that there is some confusion of element numbering. From previous conversations I think we decided that grid elements should have only one id. For instance, *active_links* don't have a different numbering from *links*. I would think this should be the same for all elements - *core_node*s don't have a different number of regular *node*s.

I think it then follows that we shouldn't have any methods of the form:

    element_at_[active|core]_element

Examples would be:
* `node_at_active_link_head`
* `node_at_core_cell`
* `cell_at_core_node`

Instead, we should do the following:
* `grid.node_at_link_head[grid.active_links]`
* `grid.node_at_cell[grid.core_cells]`
* `grid.cell_at_node[grid.core_nodes]`

SiccarPoint:
Agreed, I think.

Dan

nicgaspar:
I agree too.  Thanks for getting on this Eric!

nicole

PR #136, opened 2015-07-31
Couple climate model driver

brigidlynch:
I'm working to couple Landlab with the WRF climate model and have built a driver that incorporates the stream power, diffusion, and flow routing components and am inputing WRF output of topography and discharge. I'd like to get some input on my driver and am thinking this is the best way to do this. More specifically I want to know: if I have indeed built it correctly, if there's anything I could do to improve it and  if the edits I've made to input topography and discharge are performing as they should be. 

Thanks!


jennyknuth:
Hi Brigid, finally getting back to you on this. What is the status? Is this still your current question?

brigidlynch:
Hi, this is still sort of my question but I've made some progress. I realized that when I input my discharge data in my driver using 'water__volume_flux', it was being overwritten in the flow router component and my output discharge was being calculated with drainage area (dx*dx) instead of using the values I provided. To avoid this I edited the route_flow_dn component and my driver to only incorporate 'water__volume_flux_in' and deleted anything associated with 'water__volume_flux'. It seems to have fixed the problem but I don't know if this was an appropriate fix or just putting a bandaid on the issue.

jennyknuth:
I am going to close this pull request pending conversation 11/3.

PR #145, opened 2015-09-13
Fix failing docs build

mcflugen:
The docs are not building on readthedocs. It seems readthedocs is having trouble installing the netCDF package.


mcflugen:
The docs seem to be building again.

PR #168, opened 2015-10-21
sink_fill component

SiccarPoint:
New component fills pits in a starting topography, exploiting the lake_mapper for a lot of the donkey work.

Fill can either be flat, or sloping gently towards the outlet. Method is clever enough to handle simple terrain where "back flow" is needed to get parts of the flat connected up to the outlet, but it remains to be seen how it performs on real topographies (and, dare I say, LiDAR??)

Bunch of tests here confirm things are 100% robust on simple topographies, though I suspect there might be some interesting use cases lurking out there in the wild where lakes get big and topos get highly irregular.

### DIFF
--- a/landlab/components/flow_routing/lake_mapper.py
+++ b/landlab/components/flow_routing/lake_mapper.py
@@ -156,6 +156,11 @@ class DepressionFinderAndRouter(Component):
         # ^note this is nlakes-long
         self._link_lengths = np.ones(8, dtype=float)
         self._link_lengths[4:].fill(np.sqrt(2.))
+        
+        self.is_pit = self._grid.add_ones('node', 'is_pit', dtype=bool)
+        self.flood_status = self._grid.add_zeros('node', 'flood_status_code',
+                                                 dtype=int)
+
 
     def find_pits(self):
         """Locate local depressions ("pits") in a gridded elevation field.
@@ -189,7 +194,7 @@ class DepressionFinderAndRouter(Component):
         """
         # Create the is_pit array, with all core nodes initialized to True and
         # all boundary nodes initialized to False.
-        self.is_pit = self._grid.add_ones('node', 'is_pit', dtype=bool)
+        self.is_pit.fill(True)
         self.is_pit[self._grid.boundary_nodes] = False
 
 #        # Get a list of active links; in a raster, this (TODO: optionally)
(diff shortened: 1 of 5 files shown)


mcflugen:
I already had a quick look before I was unassigned. I noticed two small things.
* Is `assert_close` the same as the existing numpy `assert_array_almost_equal`? It looks like it. If so, probably best to use the existing numpy function.
* Docstrings need to be numpydoc format. Headings like, "Return fields" will either get lost or mangled when converted to our HTML documentation.

I love the tests!

SiccarPoint:
1. Only insofar as assert_is is the same thing as assert_array_equal…

2. How can we be numpdoc compliant when we talk about fields??

mcflugen:
1. Not sure what you mean. `assert_array_almost_equal` compares two arrays, element-by-element, and checks they are equal up to some precision. Isn't that what your `assert_close` does?
2. Good question:
  * We could hack numpydoc to recognize additional headings.
  * Don't use headings but, instead, **bold**. See the `find_pits` method of the `DepressionFinderAndRouter` for an example.

SiccarPoint:
1. I wrote that method to compare floats, basically. Seems wrongheaded to be calling assert_array_almost_equal to compare two floats. The fact it will also work on arrays is by-the-by.

2. OK, bold could work. But it looks a bit clumsy IMO. I was under the impression numpydoc was a guideline, not crucial to code parsing.

mcflugen:
1. Sorry, I thought you were comparing arrays. `assert_almost_equal` compares two floats (or arrays, actually). The `assert_array_*` functions check that additional things match, like array shape.
2. A bit clumsy looking, for sure. Maybe something else could work - but just not headings. We run numpydoc over the docstrings and it then extracts sections (identified by headings) and formats them depending on the type of section. I've hacked numpydoc before to recognize additional sections. We could definitely do that if we felt there were landlab-specific sections that we use over and over.

jadams15:
To really bring home the consistency thing: does it matter that we use sink/pit/holes interchangeably? Unless these words are used elsewhere for other purposes, it shouldn't matter. 

Also, d4 should be straightforward in this example, we would just need to add a flag to trigger the route_flow to use d4 instead of d8. I think this could be just what I need..!!

jennyknuth:
Great work, all. The more eyes the better!



—

SiccarPoint:
Charlie pointed out to me a couple of days ago that the GIS crowd tends to use “sink” when they’re talking about filling in a hole with sediment. “Pit” describes the hole itself.

i.e., a filled pit is a sink. An unfilled pit is just a pit.

jadams15:
So pits and holes are the same thing, but sinks are pits or holes filled with sediment. I get it now. Thanks.

jadams15:
my only other question: have you tested this on a DEM yet?

SiccarPoint:
Nope.

jadams15:
ok. was curious if it will be fast or slow. maybe this afternoon i'll give it a shot!

mcflugen:
Regarding pits/holes/sinks in DEMs: I'm definitely not at all part of the GIS crowd, but, for what it's worth, when I think of "holes" in DEMs, I think of missing data - not pits or sinks like we're talking about.

jadams15:
So, I guess the question is: does the component HoleFiller actually fill holes - find areas of depressions and nodata within a grid and fill them? Or does it just handle sinks/pits? Dan - I am supposed to merge this pull request, are you done adding new commits?

SiccarPoint:
It’s pits only, as indicated by the rest of the path. And the docstrings.

I was trying to avoid a path that looks like

landlab.components.sink_fill.fill_sinks import sink_filler

because that looks overly repetitive to me. Or is that OK?

mcflugen:
I agree, that's overly repetitive. How about put the following in `sink_fill/__init__.py`
``` python
from .fill_sinks import HoleFiller
```
so that the user then can do,
```python
from landlab.components.sink_filler import HoleFiller
```
I guess you could then change `HoleFiller` to `SinkFiller`, if you wanted.

SiccarPoint:
I think as a general approach it would be good to do this with all our components!

Think I’m now tending to sink_filler too.

jadams15:
Hi Dan - are you all done with this? Can I go ahead and merge?

SiccarPoint:
Hang about - I’ll change it to SinkFiller. Guess that’s clearer.

jadams15:
ok. let me know when its all good to go - I'll keep an eye on the tests and merge when I get an all clear.

SiccarPoint:
Actually, do it now. I'll change the name in a future update (imminent!). This is because I've screwed up my branch management locally, and it's going to make my life easier...

jadams15:
sounds good to me! Merging now...

jadams15:
also - we should discuss when we should "delete branch" after a merge. I do not think that the person assigned to review the code - in this case, me - should delete branches. That could be mighty dangerous.

SiccarPoint:
Yes indeed!!! Person who made branch, deletes branch.

jadams15:
100% agreed.

mcflugen:
I think it's safe to delete the branch. The requester should still have it on their local copy and we can always recover it later if we really need to (although I've never actually done this, it seems straight forward).

jadams15:
... I don't know. It still makes me very nervous.

SiccarPoint:
I am, in fact, still working on this branch right now.

Though as Eric says, pushing it back up wouldn’t be hard.

mcflugen:
Sorry. I only saw your comments after posting mine. I'm totally fine with that.

I just wanted to point out that, although it might be a bit of a pain if someone deletes a branch they shouldn't, it is reversible.

jadams15:
Very good to know! Thanks, Eric!

PR #169, opened 2015-10-21
Add gradients module

mcflugen:
This pull request adds several new modules to the `landlab.grid` subpackage. New modules include:
* *gradients.py*: functions that calculate gradients on an unstructured grid.
* *raster_gradients.py*: functions that calculate gradients on a raster grid.
* *raster_steepest_descent.py*: functions that calculate steepest descent on a raster grid.

I moved functions from `grid_funcs.py` and `raster_funcs.py` into these modules to clean things up. Each function that starts with `calculate_` in these new modules are added to either `ModelGrid` or `RasterModelGrid` as methods.

There is very little new here. It's mostly a reorganization.


mcflugen:
Could someone have a look at this before merging? Apart from the usual things, I would like to know if the functions that are now in `gradients.py`, `raster_gradients.py`, and `raster_steepest_descent.py` are still used (and useful) or if some should be removed.

Some of these functions return masked arrays. They add a lot of complexity to the code. Do people use that functionality? If not, I would like to get rid of them. If so, I wonder if there is another way to do the same thing. One option would be to leave it up to the user to mask the arrays as they see fit.

SiccarPoint:
I’m tagged to do this, but I have higher priorities for this evening…

I’m not sure how to go about assessing how well used functions are without just searching and reading all the code at each use. I’m reluctant to sink a lot of time into this. Any bright ideas from the group?

I wrote the masking code. If you can figure out a better way, have at it. But given it seems robust, and works, this doesn’t seem super high-priority?

D

PR #170, opened 2015-10-22
Move the cellular_automata subpackage

mcflugen:
This pull request moves the `cellular_automata` subpackage to `landlab.ca`. There is also some minor code clean up. I don't think it's quite ready to be merged yet but I thought I would initiate the PR so people could have a look (particularly, @gregtucker)


gregtucker:
Looks great to me --- thanks for cleaning up the formatting!
One tiny thing: the type of the prop_reset_value parameter is user-dependent. For example, if the user's prop_data contains floats, then it's a float, but it could potentially be an int, or an object, or whatever.

mcflugen:
Thanks. I updated the docstrings to reflect that *prop_reset_value* can be things other than a `float`.

PR #181, opened 2015-10-29
Fixed link-fixed value node conflict

SiccarPoint:
This pull resolves an issue where it was not possible to change a node
at the end of a fixed link to a fixed value node. It is now possible.

Note it is *still* impossible, by necessity and by design, to set a
node at the end of a fixed link to closed. You need to convert the
fixed link to a regular open (or closed) link first.

### DIFF
--- a/landlab/grid/base.py
+++ b/landlab/grid/base.py
@@ -1399,7 +1399,7 @@ class ModelGrid(ModelDataFields):
                         'Assuming your solar properties are in degrees, '
                         'but your slopes and aspects are in radians...')
                     (alt, az) = (numpy.radians(alt), numpy.radians(az))
-                    #...because it would be super easy to specify radians,
+                    # ...because it would be super easy to specify radians,
                     # but leave the default params alone...
             else:
                 raise TypeError("unit must be 'degrees' or 'radians'")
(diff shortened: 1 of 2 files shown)


SiccarPoint:
This should do it Jordan. Merge at your leisure.

jadams15:
you are my hero, @SiccarPoint. And many thanks to @mcflugen as well. Once the tests finish I’ll review and merge.

mcflugen:
@jadams15, if you have a chance, could you please add either doctests or some unit tests that replicate your problem?

jadams15:
sure. I’ll add that to this branch before merging. I’ll write up a few for fixed links.

mcflugen:
Wonderful. Thanks. If you think the docstring for the function in question could be more clear, maybe tidy that up too?

jadams15:
@mcflugen - question. if a function has a test in the docstrings, does it need a separate unit test “test_this_thing.py” file as well? I can see how it could be a good safety net but I also don’t want to reinvent the wheel. Case in point: in structured_quad/links.py, a lot of the stuff I wrote has tests in the docstrings. But there *is* a structured_quad/tests/tests_links.py. Should I add tests to that file too?

mcflugen:
It doesn't *need* one. The doctests should be short, instructive examples for our documentation. That's really their main purpose. Their secondary purpose is to serve as unit tests. The unit tests in our *test_this_thing.py* modules are our primary means of testing and should be more extensive than the doctests. So, I guess, the answer is, "yeah" (but they shouldn't duplicate a doctest).

jadams15:
Gotcha. These are good things to know! Unit tests are easy to write during Thursday Night Football!

jadams15:
@mcflugen what is going on with Python 2.7? AppVeyor tests are failing for not just this pull request, but all commits with that version on x64 and x86. Can I still merge or does this need to be addressed first?

mcflugen:
I don't know what's going on but it appears to be something on their end. Python 2.6, 3.3, and 3.4 appear to be just fine. It's just Python 2.7 and it's not our tests that fail, the build crashes simply trying to install Python. Since things are working everywhere else, I think it's safe to assume things are also working for Python 2.7 on Windows.

jadams15:
OK. Gonna merge.

PR #182, opened 2015-10-31
Fix _reset_lists_of_nodes_cells

### LINKED ISSUE #171: Core cells incorrectly updated on node-status change
I could be wrong but it seems the core cells are not being updated correctly after a node status change. For example,
```python
>>> import landlab
>>> grid = landlab.RasterModelGrid((4, 5))
>>> grid.status_at_node[7] = landlab.CLOSED_BOUNDARY
>>> grid.core_cells
array([0, 1, 2, 3, 4])
```
I think what's happening is that the old-style indexing is still being used here. That is, *core cells* have their own indexing. Probably, then, `ModelGrid._reset_lists_of_nodes_cells` needs updating. It needs updating not just for the indexing problem but also it's using old terminology (*active* nodes and cells vs. *core* nodes and cells). Truthfully, I'm not 100% sure what the correct answer for the above example is (I would guess `[0, 2, 3, 4 , 5]`) but I'm pretty sure it's not `[0, 1, 2, 3, 4]`.

mcflugen:
This pull request fixes #171. The main change is to the `_reset_lists_of_nodes_cells` method of `ModelGrid`. However, in fixing this I also removed arrays that used old terminology. These include:
- `activecell_node`
- `node_corecell`
- `corecell_node`
- `node_activecell`
- `active_cell_areas`

Many of these arrays were still being used in `VoronoiDelaunayGrid` so maybe @gregtucker you could have a look?

One bit of confusion is that I'm not sure if we have a name for a node that is either *core* or *open boundary*? It seems these used to be *active*.

### DIFF
--- a/landlab/components/detachment_ltd_sed_trp/generate_detachment_ltd_transport.py
+++ b/landlab/components/detachment_ltd_sed_trp/generate_detachment_ltd_transport.py
@@ -9,7 +9,7 @@ E = K*f(qs)*(A**m)*(S**n)
 
 """
 
-from landlab import Component, ModelParameterDictionary
+from landlab import Component, ModelParameterDictionary, CLOSED_BOUNDARY
 from landlab.components.flow_accum import flow_accumulation
 import pylab
 import numpy as np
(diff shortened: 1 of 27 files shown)


mcflugen:
OK. I think this is ready to be merged now.

gregtucker:
Feels great to get this old stuff cleaned up! The mods look good to me. My only worry concerns the calculate_flux_divergence functions. I'm not sure why "cells" keeps showing up when we're calculating divergence at nodes. Back when I first wrote ModelGrid, I sloppily used node and cell interchangeably, and maybe later on someone else made mods thinking (reasonably enough) that I meant cells instead of nodes. Anyway, probably what we need is:
* good unit tests for the flux divergence functions/methods
* good unit tests for Voronoi grid
How about we accept this pull request, and I'll add an issue for these items?

mcflugen:
That sounds great, Greg. I added some doctests to the flux divergence methods but nothing major. Yes, it definitely feels good to finally clean up this old stuff.

PR #186, opened 2015-11-02
Gtucker/cleanup

gregtucker:
This pull request is for a branch that simply cleans up some old code. The fracture_grid component now includes unit tests, and allows the possibility of passing a random seed so as to reproduce a particular configuration. voronoi.py also has a couple of old debug lines and some commented-out code removed, plus shortening of comment lines.


mcflugen:
Looks good.

I have a couple of suggestions for changes that can be done later.
- *numrows* and *numcols* should be passed as a tuple instead of separate arguments.
- Similarly, *x* and *y* should be passed as a tuple of (*y*, *x*) - same with *dx*, *dy*, etc.

This reduces the number of input arguments, matches numpy conventions, and makes it easier to extend to 1D or 3D.

PR #187, opened 2015-11-02
SiccarPoint/var doc

SiccarPoint:
This pull renames the LL standard property _var_defs to _var_doc, as agreed at the meeting.

It also updates the website documentation to reflect this.

The component introspection script is being moved to its own branch.


mcflugen:
Looks good.

I still think we need to be careful have `_input_var_names`, and `_output_var_names` be mutable. It seems like it could be dangerous. We can discuss later.

Issue #193, opened 2015-11-04
Review and clean up cell area calculation in voronoi.py


gregtucker:
Make sure calculation of voronoi cell areas is up to date with current naming usage and best practice

PR #194, opened 2015-11-05
Fix number of elements attributes

mcflugen:
We have lots of attributes like `_num_core_nodes`, which is just the length of the array `_core_nodes`, and so is redundant. This pull request mostly just cleans up our code and removed redundant `_num_*` attributes.

This cleans up some of the cell area arrays so, I think, may also address #193.

This isn't quite ready to be merged yet. There are a good number of changes (mostly minor) to `hex.py`, `radial.py`, and `voronoi.py` so @gregtucker you should probably have a look at this.


mcflugen:
This should be ready to be merged now. I've rebased it with @SiccarPoint's tests, which are now all passing. This has turned out to be a much bigger clean-up than I had originally intended.

Highlights:
* Removed redundant `_num_*` properties.
* Removed the `create_node_array_zeros` and `create_active_link_array_zeros` methods. Use either `add_zeros` or `zeros` instead.
* Added `at` keyword to `zeros`, `ones`, `empty`. So, `grid.zeros(at='node')` or `grid.add_zeros('topographic__elevation', at='node'`.
* Changed fields so that the number of elements of each field doesn't have to be determined on instantiation. This is important for fields like `active_links` whose number may not be known at the time of grid creation.
* Removed `min_active_link_length`, `max_active_link_length`. Use `np.min(grid.length_of_link[grid.active_links'])` instead.
* Added `area_of_cell` property.
* Added `face_at_link` property.
* Removed `active_cells`, `active_nodes`, `activecell_node`, `node_corecell`, `corecell_node`, `node_activecell`, `active_cell_areas`, `active_link_length`, `active_link_ids`. These use old naming conventions.

I'm probably missing a few things but those are the main changes.

gregtucker:
Good to go

PR #195, opened 2015-11-06
Add neighbors_at_node property

mcflugen:
The pull request adds a new property for getting node neighbors and renames an existing method that gets *active* neighbors.

A *neighbor* is the node on the other end of a link. An *active neighbor* is the node on the other end of an *active* link.

Added `ModelGrid.neighbors_at_node` attribute. Note that, thus far, this is only implemented for a `RasterModelGrid`. Other grids need to define their own `_neighbors_at_node` array. Until that's done, for those grids, using `neighbors_at_node` will raise an `AttributeError`.

Renamed `get_neighbor_list` method to `get_active_neighbor_at_node`.

### DIFF
--- a/landlab/components/craters/craters.py
+++ b/landlab/components/craters/craters.py
@@ -430,7 +430,7 @@ class impactor(object):
                     depth_excavated = (pre_elev-_new_z)
                     crater_vol_below_ground += depth_excavated
                     self.mass_balance_in_impact -= depth_excavated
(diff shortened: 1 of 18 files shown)


mcflugen:
I've also added a `diagonals_at_node` property to `RasterModelGrid`.

@saisiddu: does this fix the problems you were having? If so, I think it should be ready to be merged.

saisiddu:
Yes Absolutely! Thanks a lot @mcflugen

Issue #191, opened 2015-11-03
Node-link-neighbor data structures


gregtucker:
What are our current data structures for storing:
1 - the links attached to each node, and
2 - the neighbors of each node?
One set I know about, because I created it, is: node_inlink_matrix and node_outlink_matrix. There seem to be others, but I'm struggling to figure out what they are exactly. I'd like to clean this up so we have just ONE consistent set of data structures with a reasonable balance between algorithm efficiency and storage.

Specifically, I want to explore the utility of replacing our current data structures with the following, which are meant to represent all grid types:

1 - links_at_node: NxM matrix containing IDs of links that connect to each node (M=4 for raster, 6 for hex, and MAX_NUM_NEIGHBORS for Voronoi)

2 - link_signs_at_node: NxM matrix containing 1 if node is the link's head, -1 if it's the tail, and 0 if there is no link at this position

3 - active_link_signs_at_node: NxM matrix containing 1 if node is the link's head, -1 if it's the tail, and 0 if there is no link at this position OR if link is inactive

4 - neighbors_at_node: NxM matrix containing IDs of neighbor nodes (corresponding exactly to links_at_node)

If, for example, you want to mask out inactive links or neighbors on inactive links, you mask out the zero elements of the active_link_signs_at_node array.

These, or any other, data structures must be compatible with (at least) the following operations:
1) flux divergence
2) steepest descent / flow routing
3) neighbor search/processing operations
4) mapping vector values at links back onto nodes (and vice versa)
5) D-infinity flow routing (slopes of triangles)

SiccarPoint:
We already have a good subset of these - with the proviso that they are not returned in a compatible fashion, which is very, very annoying (see #156). The set is (for raster, at least...):

1. mg.node_links()  -> (4, nnodes), SWNE. Needs renaming.
2,3. We don't have these, yet.
4. mg.get_neighbor_list() -> (nnodes, 4), ENWS. Also needs renaming.

On the plus side, note that the need for renaming lets us keep these old functions a bit (thus not breaking the absolute ton of complex code I've written that already works around their indeosyncracies), while still introducing the new, correct forms.

I'd also rephrase your final paragraph a bit - we already do all these things, working with what we have, so clearly what we have already meets your criteria! The priorities for me are actually twofold: that doing these things going forward gets *easier*, but modifying what we have already doesn't become a complete nightmare.

SiccarPoint:
So, immediate, non-controversial priorities:

[...]

Issue #156, opened 2015-10-15
shape and ordering incompatibilities


SiccarPoint:
mg.node_links().shape = (4, nnodes)  # SWNE
mg.calculate_gradient_across_cell_faces(z) = (ncells, 4)  # ENWS

SO STUPID. STILL.

jadams15:
agreed! drives me crazy!

mcflugen:
Any preference on which way we should go with?

Because Python's row-major, I suspect, for most grids, it's faster to create these sorts for matrices as `(4, nnodes)`: first get all the east links, then north, west and south links and stack them on top of one another.

However, for most use-cases, it's probably better as `(nnodes, 4)`: users will want to get all the links for each node and then loop through nodes.

jadams15:
I thought the agreed upon convention was SWNE. all of the functionality I've worked on for links now returns arrays in this manner. That said, I can change it if the consensus is ENWS.

mcflugen:
Sorry. I didn't mean we should change the ordering. It was just an example. Whatever ordering we're using now, we should probably stick with.

SiccarPoint:
Consensus is definitely SWNE. It’s just a SODOIT issue. Note, pleasingly, that SWNE = ENWS[::-1], which makes converting existing uses easier (and ::-1 is a very cheap Python index operation).

The (x, nnodes) vs (nnodes, x) needs a consensus decision still. Let’s do it tomorrow.

D

PR #196, opened 2015-11-12
Fix the ordering of elements around another element

mcflugen:
This pull request is an experiment to standardize the ordering of one type of element around another type of element. For example, links around a node, and nodes around a node. Previously, node neighbors were ordered counter clockwise starting at the eastern node but links were ordered clockwise starting at the southern link.

This address, at least partly, #156 and #191.

**This is definitely not complete** so not at all ready to be merged.

Added a new property, `links_at_node`, that is an `nnodes x links_per_node` matrix ordered as ENWS. This is intended to replace the method `node_links` that returned a `links_per_node x nnodes` array of links ordered SWNE.

Changed components that used `node_links` to use the `links_at_node` property.

I found two possible bugs in:
* `stream_power.py` (line 286)
* `tl_fluvial_monodirectional.py` (line 440)

In both cases they use `node_links` as though it is an array when it's actually a method.


mcflugen:
I think this is ready to go. The ordering of one thing around another should how follow E, N, W, S. (so, neighbor nodes around a central node, links touching a node, etc.). I would like to merge this before the big merge of @gregtucker's `sort-links` branch. I hope it will make things less painful.

I know everyone's busy with AGU but, maybe (?), @jadams15's could have a look?

SiccarPoint:
Yesterday, I realised another place these changes need to happen: in the grid edge BC setters. These normally do something like (bottom_is_closed, left_is_closed, top_is_closed, right_is_closed). I *always* have to check this order when I use them, so these are a prime candidate for updating.

BUT

These are also used extensively in the various examples, including in the drivers repository. We’ll have to be careful with these.

mcflugen:
Already done (as of like a minute before your post).

I actually think they should be deprecated now. It's much easier and more readable to do something along the lines of,
```python
grid.status_at_node[:, 0] = CLOSED_BOUNDARY
```

SiccarPoint:
Disagree, at least for now. These are very helpful one-line methods in real use cases which we should retain. Any alternative requires deeper knowledge of LL for a user; e.g., your example requires the user either to know how to reshape the grid as a view, or to know where LL stores its lists of edge nodes, and also to know how the import the NODE_STATUS values. Which as of a couple of days ago, was still really opaque in the master. Iff we could sort out easy access to the edge IDs, I’d feel better about:

```
>>> mg.status_at_node[mg.nodes_at_left_edge] = CLOSED_BOUNDARY
```

Did you trawl all the examples for uses? Impressive. (And the Drivers repository…?)

Also, final point of order - I’m 80% of the way through setting up unit tests for many of the components. I’m very keen to get these committed to master before any of your or Greg’s changes get pulled. I’m suspicious things like this could easily be breaking components, and we would have zero idea until it was too late.

mcflugen:
And just for further increase the confusion to max, `set_inactive_boundaries` (which is deprecated) is SENW while its replacement, `set_closed_boundaries_at_grid_edges` is SWNE.

SiccarPoint:
Re ordering changes... yeah. That's annoying. But at least most (NOT ALL) of the existing BC uses are symmetrical.

mcflugen:
Agreed. I'm fine keeping those methods. *However*, the fact that you (and me and probably everyone) always has to go back to the docs to figure out the ordering might be a sign we need to change something to make it more clear. Sticking with ENWS helps, certainly. Maybe add keywords?
```python
def set_closed_boundaries_at_grid_edges(east=True, north=True, west=True, south=True)
```

mcflugen:
As you pointed out, my example wasn't totally correct since `status_at_node` is flat. I just meant setting the boundary conditions through the `status_at_node` attribute. So, I agree,
```python
grid.status_at_node[grid.left_edge] = CLOSED_BOUNDARY
```
would by the way to do it.

SiccarPoint:
Yes to keywords. That’s a great plan.

Let’s retain both possibilities for now. There’s little harm in having two ways to do it. (yeah, I know, not pythonic…)

Also, we can’t actually do mg.left_edge can we? Or rather, I guess, mg.nodes_at_left_edge. Would be good to have (not squirreled away in sgrid, or whatever).

mcflugen:
No, we can't do that currently. I think adding something like that would make sense, though.

jadams15:
re: nodes at left edge, we do have a function call for that, yes? mg.left_edge_node_ids() unless it has been changed... I struggle to keep up with all of this (very important, very useful) progress!

Eric - I can try to take a look tonight. Will also prioritize finishing up the link tests today. It has been super hectic this week, trying to play catch-up after our very productive two days with Gilles Brocard last week and running my AGU stuff.

SiccarPoint:
This does exist, yes, but I can never find it when I need it because it is named so oddly! Should definitely be nodes_at_left_edge, and a property not method.

Again, please hold off until my tests are in! But that should be by this evening. I’ll let you know.

D

mcflugen:
@SiccarPoint: yes I will definitely hold off until you give the go-ahead.

@jadams15: you are correct, we do have `left_edge_node_ids`. I agree that it's oddly named and should be renamed and, I think, probably be a property and an iterator (not an array). Please have a look whenever you have a chance. No rush. Tonight would be great!

I prefer the sound of just `grid.left_edge` but it might be more in line with the rest of our naming conventions to use `grid.node_at_left_edge`.

SiccarPoint:
Problem with left edge is left edge what? We’ll probably need one for e.g. links eventually.

Would a iterator still work with:

```
mg.status_at_node[mg.nodes_at_left_edge]
```

?? (I’ve never got this)

mcflugen:
I agree. Although `left_edge` might sound good, it's not clear if it's referring to nodes, links, cells, etc.

Yes, I think we should somehow get an iterator to work. For instance the following works,
```python
grid.status_at_node[xrange(3)] = CLOSED_BOUNDARY
```
and `xrange` is an iterator. In any case, I don't think we have to carry around another array.

SiccarPoint:
Sweet. As long as it works!

jadams15:
I built counterparts for links. they're separated into vertical and horizontal functionality. things like "left_edge_vertical_links()". they'll need to be renamed eventually as well!

mcflugen:
Great. I guess a better name for them would be `links_at_left_edge` (do you ever really need horizontal links at the "left" edge?). Or maybe "edge" is redundant? `links_at_left`?

Should we be using "right", "top", "left", "bottom" or "east", "north", "west", "south"?

jadams15:
yes, you'll need horizontal links at the left edge if you want to set FIXED_LINK boundary status. 

I think I prefer to have "edge" for no reason other than I like the way it sounds, and in some of the boundary status functions we use that terminology: set_closed_boundaries_at_grid_edges(). 

The latter question: I have no strong preference. right/top/left/bottom are used in many places in the code that would need to be eventually changed.

SiccarPoint:
Top/bottom etc a bit more general - think for example of the CA in use as a vertical section, where it really is left/right/up/down. But whatever.

mcflugen:
I agree. "right", "top", "left", and "bottom" seems more general and standard. I don't know of anyone who refers to the east edge of a raster image, or the south row of a matrix.

SiccarPoint:
OK guys, pull request in. From my end, everything is passing fine, so hopefully this is a quick, clean merge.

I’m probably going to be largely offline for the next few days, so if anything needs attention that a reviewer can’t just fix on the fly, it might be better to note the issue, do the merge anyway, and I’ll fix it in master next week.

Happy Thanksgiving, all.

D

mcflugen:
Alright. This is ready to go again. I've rebased to include @SiccarPoint's new tests. I modified them to use CCW ordering (for setting the node status along edges) and they are now all passing.

gregtucker:
This looks good to go to me. @mcflugen, do you want to merge or shall I?

Great that this eliminates the use of "inactive boundary" (as in set_inactive...) in base.py, but it still exists in raster.py, right? Has anyone removed it since this PR? Or should we create a new issue to get rid of it in favor of set_closed_boundary...?

mcflugen:
Thanks, @gregtucker! I'll merge it. I still would like to wait to merge this after #222, which @saisiddu is looking at, just to minimize the disruption to @lgelb.

I'm not certain, but I think the "inactive boundary" stuff you're talking about in `raster.py` *may* be taken care of with #210. If not, let's incorporate it through another issue and pull request.

saisiddu:
I have merged the pull request assigned to #222.  Everything looks fine. Thanks Eric! I might start using write netCDF immediately. That way, I will not have to deal with multiple files lying around in my 'outputs' folder. [😊]

Issue #105, opened 2015-05-14
Fields in the grid methods


SiccarPoint:
Most of our methods in the various grid classes look like:

>>> mg.do_the_thing(values)

I think this should be (or at very least, be compatible with calls that look like):

>>> mg.do_the_thing('field_name')

SiccarPoint:
Beginning branch fields-in-grid-functions to work on this

SiccarPoint:
Completed, I think, for base.py.

mcflugen:
I wonder if it might be better to create a decorator to do this. Are you done working on `base.py` for a bit? Maybe I'll give it a try and see if it makes sense to go that route.

SiccarPoint:
Yep, I’m done. I’ll hold off on anything else for a bit.

D

PR #199, opened 2015-11-13
Add value-name decorator

mcflugen:
This pull request adds a decorator that wraps a function so that it can accept either an array of a field name. This addresses #105.

The `use_field_name_or_array` decorator wraps functions whose first two arguments are a grid and an array. The first two arguments of the wrapped function become a grid followed by either a field name as a string or an array. If needed, the array is converted to a numpy array and then flattened.

Note that this does not change the docstring to reflect the change, but it should.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,9 +9,14 @@ environment:
       secure: 8om/mA+rp/SBcVXDrFC9LgyKPcEmlseSUT4WBBOgfYM=
 
   matrix:
-    - PYTHON_VERSION: "2.6"
-    - PYTHON_VERSION: "3.3"
-    - PYTHON_VERSION: "3.4"
+    - PYTHON: "C:\\Miniconda"
+      PYTHON_VERSION: "2.6"
+    - PYTHON: "C:\\Miniconda"
+      PYTHON_VERSION: "2.7"
+    - PYTHON: "C:\\Miniconda3"
+      PYTHON_VERSION: "3.3"
+    - PYTHON: "C:\\Miniconda3"
+      PYTHON_VERSION: "3.4"
 
 platform:
   - x64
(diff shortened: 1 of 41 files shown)


SiccarPoint:
This is great, Eric. Thanks!

mcflugen:
Thanks, Dan. Sorry it took so long. Since this was your idea, I'm going to assign you to review this. It's just about ready to be merged - but not quite. I'll let you know.

SiccarPoint:
Sounds good. I’ll try to hit the docstrings too, if you haven’t got there.

D

mcflugen:
Thanks. I think the docstring issue should probably be another pull request. What I'm thinking is that the new decorator should update the docstring of the function it's wrapping to reflect that the values can now be either an array or a field name.

SiccarPoint:
That would indeed be cool - though manually reading for sense will probably still be needed, I guess.

mcflugen:
@SiccarPoint, I think this is ready to go. If you could have a look before merging, that would be great. Thanks! If we like how this works, we could decorate many more functions but right now I've only applied it to the `calculate_gradients_*` functions for raster grids.

SiccarPoint:
I’ll get to it on Monday!

D

SiccarPoint:
This looks great to me. I guess it falls on later users to incorporate this on other functions.

Only immediate action item is to add a little documentation *somewhere* in the web docs about this, or it will end up being yet more functionality we have but nobody knows about.

A potential future action item will be an enhanced version of this where the field doesn't have to be the second param. i.e., a syntax like:

def myfunc(mg, nodevals, linkvals, **kwds):
    """
    This is the func to get wrapped.
    """"

def wrapper(object):
    def __init__(self, param-element-dict):
        """
        Here, param-element-dict is passed as:
        param-element-dict = {nodevals: 'node', linkvals: 'link'}
        """
        ...

mcflugen:
Thanks!

If you're alright with this, I'll add this decorator to the rest of our functions that follow this pattern.

I don't think it would be too hard to add a keyword argument to this decorator that would give the name of the *field* argument. This would take care of cases where the field isn't the second param.
```python
@use_field_name_or_array('node', arg='node_values')
def my_func(grid, arg0, arg1, node_values):
    ...
```
I'll have a look at this tonight and, if it really isn't too hard, add it.

Where do you think we should add documentation for this in the web docs?

SiccarPoint:
What about cases where there might be more than one field? Would that work too? Two decorators??

Re docs, I’m not totally sure. Maybe where we talk about fields:
http://landlab.readthedocs.org/en/latest/model_grid_no_tutorials.html#adding-data-to-a-landlab-grid-element-using-fields <http://landlab.readthedocs.org/en/latest/model_grid_no_tutorials.html#adding-data-to-a-landlab-grid-element-using-fields>

Maybe a sentence somewhere in there.

D

SiccarPoint:
I'm merging this, on the assumption @mcflugen has in mind the need to update the docs...

PR #200, opened 2015-11-18
SiccarPoint/component introspection

SiccarPoint:
Adds a script to allow introspection of the LL component code base, to check for internal consistency, etc.


jennyknuth:
Anyone in particular you'd like this assigned to? Mcflugen? Me?

SiccarPoint:
Whoever. Should be basically trivial, as it’s not subject to our standards.

D

mcflugen:
I would like to move the contents of `introspect_components.py` into the landlab library. Possibly to `landlab/framework/inspect.py`? I think inspect is a better name since the script is inspecting components, rather than asking components to tell something about themselves. The script could still be run from the command line but also callable as a landlab library function. There's also a lot of good stuff in there that should probably be broken into smaller functions that could be reused.

If people are agreeable to this, it could be another pull request or added to this one. Either way, I'm happy to do it.

SiccarPoint:
I honestly have zero clue what “framework” means. To me that’s a junk definition, and needs renaming anyway. I’d strongly oppose putting it there while it has that name. (I have a thing about this word used in science. I hate it. It’s so vague.) Also, everything else in this folder is architectural/structural for landlab (wrappers, modifiers, LL-specific class defs, etc). That’s not this, is it? Maybe I could support breaking out into smaller functions which could live here, but with the main script left where it is.

But rename and restructure away, as long as it still does what it currently does, and you don’t mess up the documentation (which took longer than I’m really comfortable with). Remember, this is 100% a pragmatic piece of code. It doesn’t need to be beautiful. Users will basically never see this. It just needs to tell us stuff. Which it already does.

mcflugen:
The stuff in `landlab/framework/component.py` does similar stuff to this, although it's outdated. I have no problem renaming it. I'm sure we could come up with a better name than "framework".

I think users will need this sort of stuff to, for instance, ask what landlab components use or provide a particular field.

SiccarPoint:
This is what our documentation is for, surely??

mcflugen:
For sure. It might be nice to allow users to do it programmatically as well. It may not be worth it, but we're pretty much there already.

SiccarPoint:
If you want to put the time in, feel free! As long as my code still does what it already does, whatever.

mcflugen:
I won't touch your code. I might copy and paste it, but that's it.

SiccarPoint:
Sweet. Go for it (if you want…)

jennyknuth:
I'm going to go ahead and merge this. I also thought this was more user-facing than maybe it is (it could be a useful tool for people writing components, for example). All in all, an excellent tool for standardization and consistency.

Issue #203, opened 2015-11-20
Flow routing, irregular grids, and internal closed nodes


SiccarPoint:
The way the router formerly handles nodes without cells was inconsistent between regular and irregular grids.
This needs to be resolved by ensuring no discharge can ever come from a node without a cell.

A related issue is that where closed nodes are present inside the grid, they get their own area, whereas more consistent with the above would be if they got zero.

Unit tests are needed, esp. for Voronoi and internal closed nodes.

gregtucker:
I'm not sure I agree that internal closed cells should have a zero cell area. An app might want to treat them open for one process and closed for another (admittedly I'm struggling to think of an example).

PR #204, opened 2015-11-20
SiccarPoint/component tests

SiccarPoint:
This pull addresses #203.


SiccarPoint:
And, note, is very misnamed!!!

mcflugen:
I made some minor formatting changes. The biggest change is that I made `forced_cell_areas` to return a read-only array. It seems like users shouldn't be able to change areas without changing the grid itself. If a user really needs to change the array, they should have to make a copy of it.

One suggestion. How about a name change for `forced_cell_areas`? It seems like it returns the area for the cell associated with each node. If a node doesn't have a cell, the area is zero. Is that right? If so, maybe something like `cell_area_at_node`, `area_of_node`, `cell_area_of_node`?

SiccarPoint:
I actually consciously did that with forced_cell_areas. I’m actually already allowing the components that use it to modify it. I guess I could redesign this to not do this, but given it’s already a “copy” of information we have, didn’t seem worth it. If you feel strongly though, I’ll just make it zero out all closed/boundary nodes and be immutable from the get-go. In fact, I’ll just do this, I think.

On this basis, cell_area_at_core_node seems most accurate and in our style. But a real mouthful.

mcflugen:
It looked like there was only one component that was changing `forced_cell_areas`. I changed it so that it created a copy of the array and then changed the copy. That seems to work fine.

`cell_area_at_core_node` doesn't seem quite right as `forced_cell_areas` has length `n_nodes`, not `n_core_nodes`. I guess I'm not totally sure what area this is supposed to represent. Maybe `forced_cell_areas` is fine and I just don't understand what a "forced cell" is (`forced_cell_area_at_node`? Also a mouthful).

I do feel that, if this is a method of a `ModelGrid`, it should be immutable. I found it confusing trying to figure out how changing values in the array affected the grid. In the end, I don't think it changes the grid at all, which is why I think it should be read-only. It could also the problematic if two components were using the array.

If you're going to change `foced_cell_areas` so that areas for closed/boundary nodes are zero, it will also have to be updated when the status of a node is changed. Would it make more sense to have this as a function rather that an attribute of a grid? I don't know. Just some ideas.

SiccarPoint:
Good catch on the auto-update need with changing BCs if we go that route. I was trying to avoid the need to copy the whole damn array, but I guess that works too. Maybe cells_at_nodes works if we leave it how you have it now. Will take a look in a bit.

SiccarPoint:
I think this is good to go now. Retained the making of local copies as necessary, and renamed property to  cell_area_at_node.

Issue #208, opened 2015-11-25
Unified method for setting node status


mcflugen:
We have three main method in `RasterModelGrid` to set boundary conditions.
* `set_closed_boundaries_at_grid_edges`
* `set_inactive_boundaries`
* `set_fixed_value_boundaries_at_grid_edges`

First of all, if you aren't familiar with these functions take a moment to think about what you think each one does. I think it might be helpful for later on.

There are, in my opinion, a few problems with these functions.

* Each of them takes four arguments, each a `bool`, that represents a node status on a grid edge. The ordering of some is SWNE while others are SENW (yes, that's not a typo, SENW). I've changed them all to ENWS as part of #196. However, if you don't know that rule, it's still pretty confusing and hard to read. For me anyway (even knowing the convention), the following isn't all that readable,
```python
grid.set_closed_boundaries_at_grid_edges(True, False, False, True)
```

* `set_inactive_boundaries` is currently deprecated and `set_closed_boundaries_at_grid_edges` is given as its replacement. However, these two methods do **not** do the same thing. One sets edges to `CLOSED_BOUNDARY` while leaving the other edges alone and the other sets edges to `CLOSED_BOUNDARY` while setting the other edges to `FIXED_VALUE_BOUNDARY`. If you didn't know this already, did you get that from their name?

***I think we should have one unified way of setting the status of nodes that's both clear and easy to use.***

Ideas?

Some thoughts to get started.
* (I think) We should ultimately get rid of all these functions. Instead users should use the `status_at_node` property to change node status/boundary conditions. So,
```python
grid.status_at_node[grid.node_at_west_edge] = CLOSED_BOUNDARY
grid.status_at_node[grid.node_at_east_edge] = CLOSED_BOUNDARY
```
It's more typing but it's way more readable and is explicit about what's going on. This has the additional benefit that it is more uniform across grid types (certainly an unstructured grid won't have a `node_at_west_edge` attribute, but still, it would look the same). Similar uniformity could be achieved by adding methods with standardized names to each grid but that would result in a fair bit of additional code in each grid type.
* In the meantime, add keywords to the above functions so that they look like,
```python
grid.set_closed_boundaries_at_grid_edges(north=True, west=False)
```
This is considerably more readable.
* If we keep these functions, we need better names that are more standardized. One problem with keeping the names is that there are lots of different combinations of statuses and we don't (at least, I don't think) we want a function for each combination.

PR #210, opened 2015-12-02
Minor clean-up for setting node status at edges

mcflugen:
This pull request addresses, but doesn't resolve, #208.

I've added the following methods to `RasterModelGrid` that get nodes along edges:
* `nodes_at_right_edge`
* `nodes_at_top_edge`
* `nodes_at_left_edge`
* `nodes_at_bottom_edge`

These methods replaces the older `*_edge_node_ids`, which used old-style naming, were method functions (rather than properties), and returned new arrays. The new methods I've added are properties and return slices, rather than full arrays.

I've also added a new function `set_status_at_node_on_edges` that is a more general function for setting node status at edges.


mcflugen:
This is probably something for @jadams15 or @SiccarPoint to have a look at.

Looking more closely at the current use cases for `[right|top|left|bottom]_edge_node_ids`, the new `nodes_at_[right|top|left|bottom]_edge` methods aren't quite replacements. For instance, you can't do:
* `grid.nodes_at_top_edge[1:-1]` to get the top edge nodes minus the corners.
* `grid.nodes_at_top_edge - grid.shape[1]` to get the second-to-last row of nodes. Incidentally, I think this pattern is probably bad form outside of `raster.py` as it assumes an implementation detail about how we number nodes. `RasterModelGrid` should provide a better way to do this.

An alternative is to introduce a shaped array of node ids. The good is that it would be fast to get node ids along rows an columns - it would return slices rather than arrays. The bad is that it would take up more memory - we would have one more array of size `number_of_nodes`.
```python
>>> grid.nodes[0, 1:-1] # The first row of nodes minus the corners
array([0, 1, 2, 3])
>>> grid.nodes[:, -2] # The second-to-last column of nodes
array([ 2,  6, 10])
>>> grid.nodes[:, (0, -1)] # The left and right edge nodes
array([[ 0,  3],
       [ 4,  7],
       [ 8, 11]])
>>> grid.status_at_node[grid.nodes[0, :]] = CLOSED_BOUNDARY
```

SiccarPoint:
Think I disagree with you that it's bad form to assume you know the ID scheme. This is why I was so unhappy at the prospect of changing fundamental IDs in the first place. But I think I agree that it would be better to just have functionality that means you aren't tempted to do this in the first place.

Is the fact we can no longer do nodes[1:-1] related to the fact this is a generator (?) now? That one I think probably does need to be preserved. That behavior is very counterintuitive to someone who doesn't intimately understand underlying Python iterables, and is very much the kind of thing a user might try. Is there anything we can do? (Override the indexing operators??)

I'm less concerned about the other operation. Would be easier just to have a second property... nodes_inside_top_edge? or something to that effect.

SiccarPoint:
...having the node grid would also be good, obviously, but I'm not sure it would adequately replace the need/desire to do nodes_at_top_edge[1:-1]. Except then I guess we can retain these methods and the new slices will work, because now they look like:

```
def nodes_at_top_edge(self):
    return self.nodes[0, :]
```

SiccarPoint:
(I think I'm just less convinced of the need to really pare away the amount of data held inside the grid at the expense of actual used functionality/intuitiveness of use. After all, one more duplication of the node grid pales to insignificance compared to memory usage in most of the components, doesn't it?)

mcflugen:
My thinking is that renumbering element IDs should have zero impact on things outside of `ModelGrid`. It's an implementation detail that users really shouldn't worry about. Because you were upset about the changes, in my opinion, points to a poor design choice we made. We should have had functions that, for instance, returned all the vertical/horizontal links (we have them now) or the inner top edge.

Yes, you are unable to index the new methods with `[1:-1]` because they are slice objects (which, I guess, are basically iterators), not arrays. I'm not sure sure about a `nodes_inside_top_edge` property. We would also need ones for the other 3 edges and what about the next inside edge?

If we had a nodes array, the `nodes_at_top_edge` property would just return a slice of the nodes array, which is fast, rather than creating a new array. I was thinking we would change those methods to exactly what you say.

I'm not certain we need to worry about memory or not. My feeling is that we do, but I can't say for sure. Maybe one more nrows x ncols array will be down in the noise compared to the overall memory footprint of the object but, then again, maybe all these extra arrays will add up and for large grids that could be a problem. It's just something to keep in mind, I guess.

gregtucker:
Is there a simple way to have versions of this for hex grids? Currently the shape of a hex grid can be either rectangular (in which case of course it has top bottom left right) or hexagonal (in which case it has *either* left and right *or* top and bottom, plus the other 4 sides). Optional argument controls both shape and orientation (whether one axis is aligned vertically or horizontally).

SiccarPoint:
Would be generic if they were labelled nodes_at_first_edge, nodes_at_second_edge, etc, where the number reflects our shiny new standardized system:

```
   2  *  1
*           *
3           0
*           *
   4  *  5

vs

   2  *  1  *  0
*                 *
   3  *  4  *  5

```

SiccarPoint:
...And I guess nodes_at_left_edge etc as before would be available instead if your hex were rectangular.

mcflugen:
I don't see why we couldn't have something like this for hex grids. Don't all hex grids have left, right, top, and bottom edges? They may be jagged edges but they are still well defined, no?

We could also have something for radial grids except it would be something like `nodes_at_outer_ring` or something like that.

SiccarPoint:
I already built some of this functionality for radial last week. I *think* it’s in master. It does require building a data structure to hold the lists, though.

mcflugen:
This is all ready to go again.

I've added a new property to `RasterModelGrid` called `nodes` that is an array of node ids of shape `n_rows` by `n_cols`.
```python
>>> grid.nodes.shape
(3, 4)
>>> grid.nodes
array([[ 0,  1,  2,  3],
       [ 4,  5,  6,  7],
       [ 8,  9, 10, 11]])
```
The edge methods now return a view into this array rather than a stride object. This allows users to use the new `nodes_at_*_edge` properties as they would expect. It also allows for a more natural and readable way of getting other rows and columns of nodes,
```python
>>> grid.nodes[1, :] # The next-to-bottom row
array([4, 5, 6, 7])
>>> grid.nodes[-1, 1:-1] # The top row, minus corners
array([ 9, 10])
```

SiccarPoint:
Awesome Eric. Will try to get to this at some point today.

However, first thing that occurs to me - it’s not very obvious from the name what the nodes property is. How about node_raster or similar, to emphasise the 2D-ness?

mcflugen:
I thought of that too. I'm definitely open to ideas. My thinking was, if you're are using it as a property of a `RasterModelGrid` perhaps the 2D-ness would be clear and you wouldn't need to qualify it with something like `raster`. I also like the name because it's short and not grid-specific.

Still, there certainly may be a better name. I'll keep thinking.

SiccarPoint:
This all looks good, except one thing - forcing users to provide a dict (or list of tuples) to set_status_at_node_on_edges() seems a little counterintuitive. Feels like the user should be able to pass right=..., top=..., etc as independent keywords too. This seems the more numpy-ish way (and would be how I'd expect this to work by default). Would adding this be problematic? (as well as the existing methods?) So it looks like:

```
mg.set_status_at_node_on_edges(right=CLOSED_BOUNDARY, left=FIXED_GRADIENT_BOUNDARY)
```

Also, further thoughts on names for grid.nodes? Happy to leave it as it is too.

SiccarPoint:
Also, need to add mg.nodes to the voronoi, presumably returning an (unshaped) list of the nodes, and update the docstring to reflect that.

mcflugen:
In Python 3.5 (and maybe 3.4) the oder of the keywords is preserved. For every other Python version, however, `**kwds` is just a regular old dict with unsorted keys. The `set_status_at_node_on_edges` function sets the edges in the order they are provided. This allows the user to better control the status of the corner nodes. Since each corner node is part of two edges, the last edge set determines the value of its corners. If this functionality isn't worth it, I can easily change it to use regular old keywords for each edge. We could also have both, although that might be even more confusing.

I added a `nodes` property (and `_setup_nodes` method) to `ModelGrid` that `Voronoi` and other can inherit.

SiccarPoint:
I think I'd still favour being able to set the edges both ways, if possible. Can see the utility of both in different cases. 3rd opinion would be good...?

SiccarPoint:
...there's also quite a lot of documentation updates related to this merge. Need to get all that sorted too. Maybe post-AGU.

mcflugen:
How about changing the name of the `grid.nodes` property to `id_of_node`? That seems to be clearer and fits with our naming conventions.

I'll add the `left`, `right`, `top`, and `bottom` keywords back to `set_status_at_node_on_edges`. I agree that it's clearer. I'm still not sure what to do about corner nodes, though. Perhaps leaving it up to the user to set by hand is fine.

SiccarPoint:
I don't know that `id_of_node` is any clearer than just `nodes`, so maybe let's just leave it at `nodes`. It has the great advantage of brevity, after all.

I guess with the corners, most consistent with our conventions would be to associate the top right with the right, top left with the top, and so on (CCW). Though invoking symmetry would probably be a better use case (maybe, "if you set it, incorporate the corners, unless there's a conflict, in which case the last named parameter takes precedence")... but now I see that written out, I hate it.

mcflugen:
I was planning to apply the boundary conditions in the order of right, top, left, then bottom, since it follows CCW ordering starting at right. However, that would mean the corners would be the opposite of what you suggest. Instead, upper-right goes with top edge, upper-left with left edge, and bottom-left and bottom-right with bottom edge. Is that OK?

SiccarPoint:
But my scheme are the corners CCW of the named edges?

mcflugen:
Sure. No problem. I can do that just as easily. So, NE->E, NW->N, SW->W, SE->S.

SiccarPoint:
Third opinion would be good to clarify I’m not being an idiot (too much turkey?)

nicgaspar:
Here's my opinion - whatever we do it just needs to be clear.  Our convention will be what some would expect and not what others would expect, no matter what.  And, maybe I'm wrong, but at least with regular grids, the corners won't matter, right?

mcflugen:
That's a good point, @nicgaspar. I guess the corners don't really matter all that much. Maybe with diagonal links?

In any case, I've changed it so that an edge sets it's CCW neighbor. For example, the right edge controls the status of the upper-right corner, etc. I also added keywords for each of the edges rather than passing a dictionary of edge statuses. It really is more clear that way.

mcflugen:
@SiccarPoint, I think this is ready to go now. Could you please give it another look?

mcflugen:
@SiccarPoint, I just merged the most recent changes in master into this branch. Are you alright with this now?

nicgaspar:
@mcflugen Are we really getting rid of set_closed_boundaries_at_grid_edges?  It's all over the examples.  I know that's not a reason to keep it.  But, I'm trying to build that screencast and wondering if I should use it.  Part of me thinks this might be something that someone who has LEM experience might expect to see.  Did we actually come to a consensus about getting rid of it?

SiccarPoint:
I don’t think we did. This seems like a place where having more than one way to do things isn’t a bad thing. Removing these seems a bit overzealous.

jadams15:
I am perfectly fine renaming the nodes_at_*_edges methods and they look good here. If set_status_at_node_on_edges() is meant to replace things like set_closed_boundaries_at_grid_edges, set_fixed_value_boundaries_at_grid_edges, I'm not sure I'm for that. I think users will benefit from having explicit function names. That, and selfishly, I don't see how these work with fixed_links. 

Having them separate too, kind of solves the corners issue. If the user calls:

set_fixed_value_boundaries_at_grid_edges(east=True, north=False, west=True, south=False)
set_closed_boundaries_at_grid_edges(east=False, north=True, west=False, south=False),

the NE and NW corners automatically become closed, because, coding 101, the last line run overwrites what was done earlier, right? Seems like we would expect users to understand that basic fact. Asking them to set corners by calling set_status_at_node_on_edges 2x seems... confusing...

mcflugen:
@nicgaspar: I didn't remove `set_closed_boundaries_at_grid_edges`. At least not yet. I would like to deprecate it, though, and eventually remove it. I think we should promote using a more user-friendly and standardized (and standarly-named) function.

@SiccarPoint: For now, we have multiple ways of doing the same thing but, ultimately, I think there should be one, and preferably only one, obvious way to do it. We may not have that way yet, but it would be nice if we can find one.

@jadams15: Those are all good points. If possible, I would like to reduce the number of these sorts of functions, as it seems like there would be a lot of repeated code otherwise. It would be good if we didn't have a separate function for every possible type of boundary condition.

All of these points should definitely be addressed in a future pull request. I really would like to get this tidied up at some point. The main aim of this pull request was to make sure the arguments for functions like `set_closed_boundaries_at_grid_edges` were ordered as ENWS and that there was an alternative function that was more explicit in its arguments (`set_status_at_node_on_edges(top=CLOSED_BOUNDARY`), for example).

nicgaspar:
@everyone (is that a thing?)

Can we please discuss this on a call?  I’m worried I’m not absorbing all the details in this email chain and it feels important.

mcflugen:
@nicgaspar: Your confusion is my fault. I haven't been very clear. Yes, let's definitely discuss this on our next call.

This pull request really was just to reorder the arguments to the BC functions from SWNE to ENWS, and try out a couple of new candidates for setting node status at the edges. The more interesting discussion that this has turned into should probably be moved over to issue #208.

jadams15:
While several of us are unsure about the deprecation of certain boundary condition handling functions, this pull request re-orders the existing functions as ENWS and also renames some of our edge finding functions to follow Landlab naming convention. I'm pretty sure this pull request can be merged and closed. Dan is assigned to it, but I can do this (many of the conflicts are in my code). Once @mcflugen gives me the go-ahead I'll handle this.

mcflugen:
@jadams15: Thanks! Yes, please merge!

PR #212, opened 2015-12-05
Gtucker/sort links

gregtucker:
The branch with nodes and links numbered according to position is now passing all tests and should be about ready to merge. However, it would be good if folks could test their components with it first. I've tested the CellLab examples and they work fine.


jadams15:
I will test my code on Monday afternoon, and will also volunteer to review the stuff in structured_quad/links.py and mappers.py!

mcflugen:
This is all up-to-date again. Do folks need to do more testing with the branch before merging (I'm looking your way @jadams15)? I'm happy to help people with testing their stuff and fixing any problems, just let me know.

jadams15:
preliminary testing shows that i'm getting different results with this versus my branch. i need to spend more time with this.

mcflugen:
OK. That's too bad. Thanks for testing, though! Please let me know if I can help. Is there a particular script that I could have a quick look at to see what changes in this pull request may be causing you trouble?

jadams15:
I'm on it right now... think the error might have been on my end (not surprising to me in the least). Seriously, only a few more minutes and I may have tracked this down and be OK with all changes.

jadams15:
wait... something is broken and I don't think it's on my end anymore. I can't even get the analytical solution to run on Greg's branch... This stuff is all on master already, in the overland_flow folder.  if you turn on the print statement at the end of the loop, you can see it gets hung up... somewhere...

jadams15:
if you want to take a look, @mcflugen, you could start there. That's the simplest test case and (eventually) will be written as a unit test. I can come back to this either later this week, during AGU (ha!) or after...

mcflugen:
Thanks. Which script is it? I'll try to have a look at it this evening.

mcflugen:
Sorry. Got it. I think. `overland_flow_with_model_grid_deAlmeida_analyticalSolution.py`? GitHub had cut off the last part of the file name in my browser.

jadams15:
overland_flow/examples/overland_flow_driver_deAlmeida.py is the driver. Just tested a clean version of master and it worked fine. That one^ I think is just hard-coding the equation without using the component. (For a while we talked about comparing component vs non-component codes so I built both. I don't know if we are still doing that, so that may be outdated stuff...)

mcflugen:
@jadams15, I made a change to `generate_overland_flow_deAlmeida.py` that may fix this problem. Could you please have a look? The problem was how you were constructing the discharge array at links. You were concatenating discharge at the vertical links with discharge at the horizontal links to get an array of discharges at all the links. However, this doesn't jive with the new link ordering.

Issue #213, opened 2015-12-06
Improving how we import components


mcflugen:
We need to improve how we import `landlab` components. Right now, we do something like,
```python
>>> from landlab.components.flow_routing.flow_direction_DN import grid_flow_directions
```
In my opinion, that's pretty clumsy, lots of characters to type and remember, and requires the user to know something about how the flow routing component organizes its modules. We should change things so that, at least, we have
```python
>>> from landlab.components.flow_routing import grid_flow_directions # Just the function
>>> from landlab.components import flow_routing # The entire sub-package
```
I've already done this for most of the components. However, if we think it's necessary, we could go further,
```python
>>> from landlab.flow_routing import grid_flow_directions
>>> from landlab import flow_routing
```
Is this necessary? Do folks have other ideas about how they think we should import stuff like this?

SiccarPoint:
I’m kind of agnostic about which of these two styles is better. I kind of like the explicit declaration that these are components, but also, more typing.

100% agree we should be able to do function/sub-package imports as you suggest though. This is a matter of putting the full path in the correct __init__.py files, right?

mcflugen:
OK. A pull request will come soon with the `from landlab.components` form. If others prefer another alternative we can always add it later.

All that needs to be done is add a bit of code to the `__init__.py` in each component folder. For example,
```python
from .route_flow_dn import FlowRouter
from .lake_mapper import DepressionFinderAndRouter

__all__ = ['FlowRouter', 'DepressionFinderAndRouter', ]
```

PR #214, opened 2015-12-07
Improve component imports

mcflugen:
This pull request addresses #213. Mostly, it just adds import statements into component `__init__.py` files. All of the components are complete. The `landlab.ca` sub-package still needs to be done.


SiccarPoint:
Web docs need updating... I'll do it as I know where they are.

mcflugen:
Wonderful! Thank you! Yes, they definitely should be updated. The old-style of importing does still work but I don't think it's something we should promote.

SiccarPoint:
...also, SPEroder no longer makes sense as a name under this scheme, as explicit reference to "Fastscape" has disappeared from the string. Needs to be renamed "FastscapeEroder". (I'll get this too)

mcflugen:
Thank you for changing. That's probably best. I'm not sure if you noticed but I changed things like `import SPEroder as Fsc` to just `import SPEroder` and changed `Fsc` in the code to `SPEroder`. I wasn't sure if that was more confusing or less. `FastscapeEroder` is perfectly clear.

SiccarPoint:
New issue - I can't import @jadams15 's deAlmeida flow router by this new method.

mcflugen:
I had to do something special with the overland flow component. I'm not sure if it relates to your issue. The following will *not work* (it gives an `ImportError`),
```python
from landlab.components.overland_flow import OverlandFlow
```
To import the deAlmeida overland flow component, you have to do the old style import,
```python
from landlab.components.overland_flow.generate_overland_flow_deAlmeida import OverlandFlow
```
The reason for this is that both the `generate_overland_flow_deAlmeida` and `generate_overland_flow_Bates` modules define an `OverlandFlow` class. If only one of these components is used, I could just import that one. Another alternative is to rename them to have different names.

@SiccarPoint, is this at all related to the problem you were having?

SiccarPoint:
That is indeed the same problem. Seems like a rename to BatesOverlandFlow and DeAlmeidaOverlandFlow (or equivalents) would be appropriate, as this is basically the same issue as I was having with the two stream_power modules. This also seems like this style will make our naming conventions a bit clearer in general.

Pending @jadams15 's input, of course.

Otherwise, I'm hopeful this is good to go, and will merge once we have this in place/decided on. I also have updates ready for the drivers and tutorials repos.

SiccarPoint:
...or indeed, if the DeAlmeida algorithm is basically just better, leave that one as OverlandFlow, and only rename Bates.

mcflugen:
There is also a similar problem in @saisiddu's `vegetation_ca` component. Both `CA_Veg` and `CA_Veg_new` modules define a `VegCA` class.

Another option would be to break the deAlmeida and Bates solutions into two separate components (that is, two separate folders under `landlab/components`). In this way, we would have to different components that implement `OverlandFlow`. I'm not sure I'm crazy about this option - just throwing it out there for @jadams15 to consider.

SiccarPoint:
In Sai’s case, I’d guess the component not labelled …_new might just be obsolete, so this problem will go away. But let’s wait for some info! (I’m guessing the AGU Fear might be taking its toll here)

mcflugen:
Yes, I think you're right on both counts!

jadams15:
heres my thought - I think we might as well keep the Bates in case someone wants to use that exact algorithm (because, why not) but we should probably make deAlmeida the default. so I guess I'm in favor of OverlandFlowBates and OverlandFlow. Although I could be persuaded otherwise.

AGU prep has completely consumed me. sorry for the relative radio silence on my end... after next week when I have nothing to do (HAHA!) I will get back into all of the stuff I have seemingly missed.

SiccarPoint:
I'll implement this. Thanks Jordan!

gregtucker:
Actually I would vote to keep some kind of descriptive name in each component name. "Overland flow" is a process. There are 101 ways to model it. No one component is ever going to be the perfect, or arguably even default, choice for all applications. So I like: OverlandFlowBatesModel or something like that.

PR #226, opened 2015-12-22
Gt voronoi link at face

gregtucker:
This PR adds link_at_face and face_at_link data structures to Voronoi grids


gregtucker:
Ok, I think this minor PR should be good to review now, assuming it passes tests... One more small thing for the to-do list: with this PR, link_at_face is a property in base.py (so, it works that way for Voronoi), but it's a method for raster.py.

SiccarPoint:
That sounds pretty confusing. I think they should both be the same (property?) if possible before pull.

mcflugen:
I made only one additional change of significance. I split up `_setup_link_and_face_arrays` into two separate methods, `_setup_link_at_face` and `_setup_face_at_link`. I realize this will be slightly slower than doing them together but it now follows the pattern we use for other arrays. This way, these arrays will only be built if they are requested. Since both of these methods are general, I moved them out of `VoronoiDelaunayGrid` and into `ModelGrid`.

gregtucker:
Perfect. All good changes. And good example of why we do code reviews!

Issue #223, opened 2015-12-18
Standardize Component "run" method


mcflugen:
I'm creating this issue just to consolidate our discussion on the topic.

All landlab components have a "run" method. This is the method that advances the component in time or makes it do something. We would like to standardize what this method looks like for all components. From previous discussions, we, I believe, decided that it should take a duration of time as an argument (as opposed to an absolute time). Thus, a user would tell a component to, "run for 10 seconds", not, "run until a time of 100 seconds".

Another thing I've noticed is that some "run" methods still accept a grid as an argument. Since a reference to the grid is already embedded in the Component, this argument is redundant and should be removed.

Thus, a "run" method should look something like,
```python
def run(self, dt, **kwds):
    # Advance in time
```
or, for time-independent component, just
```python
def run(self, **kwds):
    # Do stuff
```

mcflugen:
I forgot to add a link to @SiccarPoint's [gist] (https://gist.github.com/SiccarPoint/1bd979e42bb48d6ad0ab#file-component_signatures-py).

@jadams15, I notice the `OverlandFlow` methods accept a grid as an argument. Is this just a hold-over from the early days, and can now be removed, or is it necessary?

mcflugen:
I think it would be nice if the name of a component's "run" method (or methods) was just a little bit standardized. For instance, maybe just that it starts with a verb? I think `run_` or `calc_` would work for most components. So, `calc_flexure`, `calc_overland_flow`, etc. (I don't like `run_` all that much but I think `calc_` might work).

SiccarPoint:
Don’t like the idea of replacing these methods ([insert Dan’s time-worn breaking-user-facing-code-unnecessarily screed here]). Happy to wrap them with *additional* standard-formatted interfaces.

mcflugen:
I guess I was thinking more of a "guideline" rather than a standard. Since class methods *do* something, they *should* start with a verb and, for a "run"  type method, maybe a guideline would be that verb be `calc_` or something along those lines. I don't think there's any reason we must conform to a standard here. I think this guideline might make it easier for users, but we don't have to go back and change existing code.

jennyknuth:
Yes, I like the idea of suggested "best practices." 




Jenny 



—

Issue #220, opened 2015-12-16
Standardize Component __init__


mcflugen:
Many of our components take the name of a file (or a file-like object) as an argument to its `__init__` method. This file is intended to be an initialization file from which the component reads input parameters. There are several problems with this (I'm sure there are more):
* The user is required to know the file format
* The user is tied to this file format
* The user must create a file before initializing their component
* Every component has code that reads parameters from a file and converts to a `dict` - violates the _Don't Repeat Yourself_ principle
* If we decide to change input file formats, we have to change every component

To address these issues, we decided to do away with initializing components through input files. Python has easy-to-use and intuitive built-in syntax for passing key-value pairs to functions. Let's use that instead. It will now be the responsibility of the application (driver) to read any input files. I guess the only question now is, do we want all parameters to be passed as keywords or as a single `dict`. The options are,
```python
def __init__(self, params):
```
or
```python
def __init__(self, param1='value', param2='another value'):
```
The second method is nice because it clearly shows default values for the parameters and also provides a list of all the possible parameters the component can use. The corresponding docstring would also make more sense, as each parameter would be listed as a separate item. A downside is that this could lead to long parameters lists and clutter the definition of the `__init__` method.

Some usage examples,
```python
mpd = ModelParameterDictionary(from_file='input.txt')
component = MyComponent(**mpd)
```
or if the input file is YAML
```python
params = yaml.load(input_file)
component = MyComponent(**params)
```
or without an input file at all
```python
component = MyComponent(dt=1., stop=10.)
```

Are people still OK with this? Is it OK if we go with the keyword style? What about regular arguments for `__init__`? Does anyone have a component that requires some regular arguments? Should we insist that everything is passed through keywords?

SiccarPoint:
I thought we had already agreed to migrate definitely to a single-dictionary style at the launch meeting. I had volunteered to do this, but have been slow. Having a hybrid, where we maintain compatibility with the old pass-a-string style, would be trivial. I'm still happy to do this.

I can see that the full written out signature looks superficially appealing, but I think at the end of the day it's both less clear if there are subtleties to the way the params are passed in (e.g., optional ones), and clutters the signature. The user needs already to be forced to RTFM for components; I kind of like the idea that they are forced to by passing a dict, not individual terms.

mcflugen:
Thanks for the reminder. You all with your "memory".

[...]

PR #231, opened 2015-12-28
Update init and run interface for Flexure and OverlandFlow

mcflugen:
This pull request updates the `__init__` and run interface for the `Flexure` and `OverlandFlow` components (#220, #223). I realize we haven't yet decided exactly what the interface for the `__init__` method will look like. For now I've used the keyword style but it would be trivial to change it to the dictionary-style. This is really just to see how it will look.

Some other highlights:
* New function `load_params` that reads parameters from a file (either mpd or yaml format) into a dictionary.
* New decorator that adds a filename argument to a component's `__init__` method to accommodate the old-style that reads parameters from a file.
* Added `vertical_links` and `horizontal_links` properties to `RasterModelGrid`.
* Cleaned up user interface and added documentation to `OverlandFlow` (@jadams15 please have a look at this).


jadams15:
I made some small adjustments to the overland flow code, and now I'm fine with these changes. I ran the tests locally with no failures. 

Does anyone else need to review these before merging? I know we need to wait for the Windows tests to finish.

mcflugen:
I believe the changes are backward compatible, so we should be OK. However, before merging, we may want to hear from everyone and decide if this is the component interface we're going with. This was only one candidate.

So long as it is backward-compatible and it doesn't break drivers or other components, we should be fine merging. @jadams15, it looks like you've made sure all the drivers that use `OverlandFlow` still work, is that right?

jadams15:
I made sure it worked with two out of three drivers. Both worked exactly as anticipated. I'll check the third one now.

jadams15:
Yea, all drivers work.

jadams15:
Ok, I'm confused - we should merge or we shouldn't merge? Because waiting for everyone else implies not merging, but if its backwards compatible, we can merge.

mcflugen:
It sounds to me like all of these changes are backward-compatible, so I'm fine if we merge this.

jadams15:
Thanks for handling the conflicts, Eric!

jadams15:
(or at least, I had conflicts I was trying to resolve on my end). Either way, the merge is done!

PR #246, opened 2016-02-18
Wrap landlab components with a Basic Modeling Interface

mcflugen:
This pull request provides the ability to automatically wrap landlab components with a BMI.

Some highlights:
* New command, `landlab`, gets information about existing landlab components (`landlab -h`, for help). Some examples are:
  - `landlab list` to get names of all the components
  - `landlab provided_by OverlandFlow` to get variables that `OverlandFlow` provides.
  - To get all unique input and output items for all components, `(landlab provided_by && landlab used_by) | sort | uniq`
  - `landlab provides topographic__elevation` gets all components that provide the given variable.
  - Validate a landlab component with `landlab validate Flexure`
* You can get landlab component directly from `landlab.components` now. For instance, `from landlab.components import Flexure` works.
* BMI-wrapped landlab components are under `landlab.bmi`. So, to import the BMI version of `Flexure`, do `from landlab.bmi.components import Flexure`.


mcflugen:
A couple more notes. 

* This wrapping relies upon the standardized landlab component interface. Components that don't expose this interface will not be properly wrapped. They may look like they're wrapped but they'll most likely fail in `initialize` or `update`.
* BMI-enabled landlab components need a standard input file. The input file needs to provide, not just component parameters, but also grid and time information. See the docstring for the `initialize` method in `bmi_bridge.py` for an example with a brief description. Here's an example in YAML format
```yaml
eet: 15e3
clock:
    start: 0
    stop: 100.
    step: 2.
grid:
    type: raster
    shape: [20, 40]
    spacing: [1000., 2000.]
```

mcflugen:
@SiccarPoint: I've assigned @gregtucker to this but you might be interested in this too - particularly the new `landlab` command to query components and how the BMI uses the component standards we're trying to settle on.

SiccarPoint:
Cool, thanks.

PR #247, opened 2016-02-19
Gt handle flooded nodes in stream power

gregtucker:
Modified FastscapeStreamPower's erode method to optionally take a list of flooded nodes, and set their alpha values to zero, which forces them to have no erosion (which would actually be reverse erosion where the receiver node is higher than the donor)


SiccarPoint:
Did you ever confirm if this is also a problem with the "regular" stream power component? Seems like it shouldn't be, right?

SiccarPoint:
Looks good. I duplicated this into the stream power code, and I also modified both the SP components to never allow positive elevation changes even if the new flooded_nodes flag isn't set.

PR #248, opened 2016-03-02
Jk copy init docstring

jennyknuth:
ca, components, and core


jennyknuth:
@mcflugen, can you point me in the right direction for this error? This PR is simply to copy the docstrings from __init__ functions to the class. Not sure why/where/how it is failing, but happy to fix. Any ideas?

mcflugen:
@jennyknuth: The error appears to be fairly benign. On Python 2.6 and 2.7 on 64-bit Windows, the shape of a numpy array is being returned as long ints (so, `(800L, )` vs. `(800, )`). The error exists in master, which is why it's only failing on the merged version.

If you're ready, I think it's OK if you merge this. I'll fix the problem in master later.

Issue #250, opened 2016-03-03
clobber for mg.add_ones, add_zeros, etc


SiccarPoint:
I think that the methods to add fields of single values to the grid should have noclobber flags as add_field does. Related: I also think the default should be True, not false as it is on add_field right now (as the main use case would seem to be accidental overwrite, which if it's accidental, you will not have set the flag).

Also, the docstrings, including for add_field, should document what the default is. It doesn't appear if you do mg.add_field?, as that func is actually a wrapper for mg['group'].add_field, and is just declared with *args, **kwds.

SiccarPoint:
Happy to add this myself, but would like input on the changes first ( @mcflugen ?)

mcflugen:
That all sounds good. Have at it if you like. Otherwise, I'm happy to do this.

It looks to me like `ModelGrid.add_field` is properly documented and shows the real signature, rather than just `*args` and `**kwds`. I might be looking in the wrong place though.

SiccarPoint:
Just try opening an interactive session, creating a grid, and getting help for add_field. It shows as `(*args, **kwds)` only. Same for the other add_... options.

mcflugen:
This is what I was doing,
```
>>> from landlab import ModelGrid
>>> help(ModelGrid.add_field)
Help on method add_field in module landlab.field.grouped:

add_field(self, *args, **kwds) unbound landlab.grid.base.ModelGrid method
    add_field(group, name, value_array, units='-', copy=False, noclobber=False)
    Add an array of values to the field.
...
```
The second like gives the actual signature. I believe that on our HTML docs the `*args` and `**kwds` do not even show up. If that's not good enough, we could change `add_field` to not have `*args` or `**kwds`. However, the reason we do it now is because we're trying to support two different signatures (for backward compatibility). Perhaps we should drop the backward compatibility? I also notice that the signature in the above example is the old-style. That should be changed.

SiccarPoint:
I had my wires crossed. You’re right for add_field, but the signature is missing for the others. I’ll add it.

D

mcflugen:
Oops. Sorry. I should have caught that.

PR #251, opened 2016-03-05
SiccarPoint/fix issue 250

SiccarPoint:
Addresses issue #250 .


SiccarPoint:
Should all be fine, nothing very controversial here. Hit a surprisingly large number of doctests though:  if you show field creation twice in one docstring, you'll now by default get an exception unless you modify the flag. I fixed them all though, and still think this is the right behavior for the user.

SiccarPoint:
Bump @mcflugen : This guy wants merging. I just rebased off Master, so the previously failing tests (unrelated to these changes) hopefully will now pass.

SiccarPoint:
Found a bug where only the add_field method responded at all to the noclobber default (i.e., not add_zeros, _ones, _empty). Kicked off a bunch more unit test fails. Should all now pass.

I've reassigned this to @jadams15, as the only tests that I hit that weren't my own were yours when I mended this bug - in deAlmeida. I resolved the issue by modifying the component itself to ingest existing fields where it finds them and overwrite them with zeros. (Just setting noclobber=False in the component wouldn't be a good option, as it would break references in any other components using the existing field). This basically exactly replicates the old behaviour.

However, a better option might be to restore the original form of the component (so errors get raised if output fields already exist, if they aren't also input fields), then modify the tests themselves to not create these fields before instantiation. This is how I have my components set up mostly, as generally you want to know about it if a field your component is trying to write to already exists...

mcflugen:
I added some tests for the `noclobber` keyword.

It made me wonder. Since `noclobber` is the default, would it make more sense to have the keyword be, instead, `clobber`?

```python
>>> grid.add_ones('z', at='node', clobber=True)
```
vs.
```python
>>> grid.add_ones('z', at='node', noclobber=False)
```
`noclobber=False` seems unnecessarily confusing and a double-negative.

Thoughts?

SiccarPoint:
Is there a numpy standard or commonly used argument for allowing overwrite? We should probably adopt that… Given you adopted `noclobber` back in the mists of time, Eric, I’ll defer to your opinion! Don’t feel strongly at all if there isn’t something numpy uses.

mcflugen:
OK. Thanks. Originally, the default was to clobber an existing field so, to suppress this behavior, one would use `noclobber=True`, which made a little more sense. In any case, I think I would like to change the keyword to `clobber`. I will look for precedent in numpy/scipy, though, and any changes will be done in another branch and another pull request.

SiccarPoint:
Sounds like a plan!

SiccarPoint:
I've been fine with it. The order is the same as any of our field interfaces in my brain - grid-then-node-then-name. E.g., mg.at_node['field']. And it's pretty well ingrained throughout the code because the docs have said to do it this way.

If you do change it, make sure to do a thorough job everywhere- tests, components, examples, wiki, tutorials... The lot. But perhaps consider if this would come under change for change's sake before committing to it. And let's make it a new issue, for sure.

mcflugen:
Oh yeah. This would be in a different branch, for sure.

nicgaspar:
@SiccarPoint @mcflugen 

In a teaching tutorial I'm having a `noclobber` problem. The code block and error are attached. I sometimes need to instantiate the `SteepnessFinder` more than once, because I might vary parameters that affect `reference_concavity`. But then I have a `no_clobber` problem because the field `channel__steepness_index` already exists. I thought maybe if I created that field first and set `noclobber` to False, then I might get around the problem, but that doesn't work. I guess there is a simple solution that I'm not seeing. Can you help?
screen shot 2018-05-21 at 2 00 32 pm

mcflugen:
@nicgaspar The easy fix is to recreate your grid. The field is attached to the grid, not the component, so re-instantiating the `SteepnessFinder` doesn't help. Would that work for you example?

The slightly more difficult solution is to modify the `SteepnesssFinder` component to deal with this case. The user could control what happens if the field already exists:
* Do nothing. Use the existing field as-is
* Initialize the field
* Raise an error (the current behavior)

SiccarPoint:
Easier way: Looks like I designed this with this use case in mind. You should find a single instantiation works, then you can pass `reference_concavity` direct to `calculate_steepnesses` as a `**kwd`; see the documentation on that function. i.e., you can override the initiation variables at any time in this component.

nicgaspar:
Thanks @SiccarPoint. That is way easier. For whatever reason it didn't occur to me to check that option.

PR #254, opened 2016-03-10
Fix AppVeyor Deployment to PyPI

mcflugen:
Fixed an issue that caused wheels not to be built and deployed when a tag was created based on the release branch.

As before, a wheel will be created and deployed to PyPI if a tag is created that starts with "v" (regardless of the branch the tag was created on).


mcflugen:
This ended up being fixed by 109ea4fb.

PR #255, opened 2016-03-10
SiccarPoint/enhance mappers

SiccarPoint:
Added miscellaneous new mappers for links to nodes and vice versa, with particular emphasis on true upwind/downwind schemes.


SiccarPoint:
I just assigned this to @gregtucker as it's pretty relevant to what you were trying to achieve in your branch, but if you're too busy today I'm sure @mcflugen could scan it. This is basically just assorted helper mapping functions which the students should find useful for your class.

SiccarPoint:
Just discovered the upwind_links_at_node stuff is a bit buggy on bigger, more complex grids. I'll take a look.

gregtucker:
This is great new functionality!

PR #256, opened 2016-03-11
SiccarPoint/stream power standard

SiccarPoint:
An initial round of modernising of component interfaces.


SiccarPoint:
Hey Eric - I assigned you as you're probably best placed to review this stuff quickly and efficiently! It should all be back-compatible, so would be appropriate for inclusion in our current release version.

SiccarPoint:
I believe this is now good to go. A timely merge would be appreciated as I'm keen to get some of this stuff embedded into another branch.

gregtucker:
Sounds good

Sent from my iPhone

On Mar 18, 2016, at 12:04 AM, Dan Hobley <notifications@github.com<mailto:notifications@github.com>> wrote:


I believe this is now good to go. A timely merge would be appreciated as I'm keen to get some of this stuff embedded into another branch.

-

mcflugen:
Could you fix-up the failing tests first? I can help if you need. Just let me know.

SiccarPoint:
I took a look at this last night, and I'm pretty confused about the failing tests. On the plus side, it's only 2 distinct issues.

Python 3 appears to have an issue where if I pass in params as an active dict (in this case, an MPD), it interprets a float as a string, and then a value assertion on this value fails.

Python 2.6 I just don't understand. Something to do with the property declarations in sed_flux_dep_incision? They looked OK to me (and passed fine in 2.7!!), but I made them 100% standard, copying the format exactly from somewhere that was working, so we'll see if that works.

If you have any ideas on why Py3 (only!) is reading my threshold in an active dict as a float, that would be great...?

SiccarPoint:
As far as I can tell, this is now 100% good to go.

mcflugen:
Looks good. I think it's ready to merge.

I have only a couple questions (that can be addressed later, if necessary). The new methods of `Component` (`optional_var_names`, `initialize_output_fields`, `initialize_optional_output_fields `): do you think these should be part of the `Component` interface? That is, would an end user of a component want to call these? It seems like they're more convenience functions for component developers. Is that right? If so, perhaps we should move them from the interface.

Rename `optional_var_names` to indicate these are *output* var names?

SiccarPoint:
Thanks Eric!

1. It’s not clear to me that developer-side things shouldn’t be part of the interface, and I see no need to move them. But I don’t feel super strongly if you're motivated to move them yourself - as long as you publicise it when you do.
2. Feel free to rename the optional_vars too. This was more for conciseness.

PR #258, opened 2016-03-13
Add func to strip grid from docstring if func is imported to grid

SiccarPoint:
Resolve documentation booboo where funcs that get automatically imported from mappers.py and gradients.py into the grid retain “grid” as the first entry in their Parameters in the docstrings.


SiccarPoint:
Assigned GT as the original provoker of this feature being added.

SiccarPoint:
Per conversation on call today, I've now added to this pull request format changes such that the construction of each of these functions.methods is formally declared, and updates as appropriate on import. i.e., for a function:
```
Construction::
    map_...(grid, *args)
```
...becomes for a the imported method:
```
Construction::
    grid.map_...(*args)
```

SiccarPoint:
Now good to go.

PR #262, opened 2016-03-14
SiccarPoint/enhance mappers

SiccarPoint:
Adds ability to map only horizontal and vertical links separately, if using a raster.

Also enables a new data structure active_link_dirs_at_node, alongside GT’s link_dis_at_node. It does the same thing, but lets you work only on active links (necessary for cases with lots of internal closed nodes). Also added a couple of mappers capable of mapping only active nodes. Lots more clearly needed.


SiccarPoint:
Merging this myself as I need this functionality RIGHT NOW.

PR #263, opened 2016-03-14
SiccarPoint/enhance mappers

SiccarPoint:
Mappers can now take arrays as well as fields. Docstrings now correct also, I think.


SiccarPoint:
I'm merging this again myself... Zero controversial here. All tests pass.

PR #265, opened 2016-03-15
Missing tests on Travis and AppVeyor

### LINKED ISSUE #264: Fewer tests are running on Travis and AppVeyor than are locally
Our CI testing on AppVeyor and Travis was not finding about 70 tests (mostly in the components folder) that are found when the tests are run locally. This most likely happens only for development versions of landlab (that is, landlab is installed in development mode with something like `python setup.py develop`).

I suspect this problem would also appear with any version of landlab installed with the usual `python setup.py install` - but I haven't actually tested that.

mcflugen:
Fixes #264.

`nose` was not finding many of landlab's unit tests when installed and run in our CI environments. In fact, this issue was most likely a problem when testing landlab anywhere where it was installed in the usual way (like with, `python setup.py install` rather than in development mode).

The problem was that several test folders did not contain `__init__.py` and so were not installed with the rest of landlab.

Changes
-------
* Added missing `__init__.py` files to various tests folders.
* Install `.txt` files that are in the tests folders. These are mostly input parameter files, which should probably be removed in the future and the tests changed to pass `dict`s to components when instantiating them.
* Unzipped and renamed `.gz` files to `.txt` files so they are also installed.
* Fixed a small error in `fracture_grid.py` that caused an error in Python 3.X.


mcflugen:
@gregtucker, will you please look at my changes to `test_fracture_grid.py`. I had to change the expected result of a unit test because I also had to change the random number generator that the fracture grid component uses.

gregtucker:
test_fracture_grid.py changes look fine to me.

mcflugen:
Thanks, @gregtucker.

@SiccarPoint, @jadams15: could one or both of you have a quick look at this since (I think) the tests are mostly yours.

SiccarPoint:
Not sure what you mean. Didn't you just add `__init__.py` files? What's to look at?

SiccarPoint:
I scanned everything, and it looks good to me. The switch to txt from gz seems to have taken. I can merge unless you have any detailed concerns...

mcflugen:
Thanks. Yeah, I guess really the only change to your stuff was changing from zipped files to text files.

PR #271, opened 2016-03-22
SiccarPoint/steepness index

SiccarPoint:
Implements a steepness index calculator for Landlab, plus some associated tests (in docstrings).


SiccarPoint:
Assigned NMG, as I feel like I've been overwhelming Greg and Eric with these recently. Also, I'm assuming the NMG who crops up in the geomorphtools matlab codes for doing this is you, Nicole!

PR #274, opened 2016-03-25
SiccarPoint/plotter modernisation

SiccarPoint:
Improve @deprecated decorator to meet our current needs.

Start using it, esp. in grid and plot.

Modernise docs in plot component.


gregtucker:
Looks great. Note to all: change in name of imshow_node_grid ==> imshow_grid_at_node (and same for cell). I'm merging this.

PR #276, opened 2016-03-28
Sed flux dep

SiccarPoint:
Merging an old branch that is starting to give persistent merge conflicts off Master (which is super annoying). 

There’s probably stuff here that wants removal from the actual LL Master, but I’m going to do it retrospectively as the conflicts are unmanageable.


SiccarPoint:
Closed, as I don't trust the random changes to docs (that turned up without my intervention!) to be behaving right.

PR #282, opened 2016-04-09
Gt fix faces at cell

gregtucker:
This branch makes raster faces_at_cell a property, by:
* changing the name of raster faces_at_cell() to make_faces_at_cell(), which is called by base faces_at_cell
* having it create grid._faces_at_cell

Also updates tests


gregtucker:
I'm going to merge this. The appveyor fail was only because one of the tests timed out (60 min max). The other appveyor tests passed.

PR #288, opened 2016-04-12
SiccarPoint/component modernisation

SiccarPoint:
This pull modernises the style of the LinearDiffuser and SedFluxDepIncision components, as well as a few functionality enhancements (e.g., diffuser now handles spatially variable diffusivity).


SiccarPoint:
@mcflugen : this bad boy has started throwing int64 related errors in the Windows builds, though I didn't touch the parts of the code where they're occurring. Hopefully they'll vanish on merge to Master, but if you think they won't, please jump in beforehand.

mcflugen:
@SiccarPoint: It doesn't look like they'll be fixed with the merge. It should be an easy fix in `lake_mapper.py`, though. `pit_node_ids` just needs to be change to be an array of `int`.

mcflugen:
@SiccarPoint: your changes should work but it's better to use `as_id_array` (`from landlab.core.utils import as_id_array`) to avoid unnecessary copies.

SiccarPoint:
I'm not familiar with how that works... Could you dive in?? (Sorry...)

SiccarPoint:
...hang on, I got this - just wrap the output, right? And it's an in-place operation, not a copy?

mcflugen:
Correct. The problem is that `astype` creates a copy by default. So the following is true,
```python
>>> x = np.arange(12, dtype=int)
>>> y = x.astype(int)
>>> y is x
False
```

mcflugen:
@SiccarPoint: Sorry. I had already pushed the change before I saw your message.

SiccarPoint:
...but, in contrast:
```
>>> x = np.arange(12, dtype=int)
>>> y = as_id_array(x)
>>> y is x
True
```
? If so, that's great to know about.

gregtucker:
Looks good, I am going to merge

SiccarPoint:
No worries. Good to know my code is verbatim same as yours!

gregtucker:
...after the tests complete

mcflugen:
Yes, `as_id_array` should just return the same array if it doesn't need to make a copy when converting to `int`.

mcflugen:
@gregtucker: Yes, please wait for the AppVeyor tests to complete. I canceled some redundant builds but there's still a bit of a queue so it might take a while.

gregtucker:
Looks as if there's a Travis failure here -- just one of the tests, and maybe it timed out on build. Will it be automatically run again?

mcflugen:
I restarted the build. Things look good. The AppVeyor builds are still running but it looks like they'll all pass.

PR #290, opened 2016-04-14
SiccarPoint/fastscape threshold

SiccarPoint:
Implements an erosion threshold as part of the FastscapeEroder. K, threshold, and rainfall_intensity can all be fields or arrays.

Some component standardisation.


jadams15:
the only fail is the known Python 3.3 fail in AppVeyor. This is fixed in the master branch so I'm going to go ahead and merge this pull request.

PR #294, opened 2016-04-22
fix diffusion.py g / qs issue

### LINKED ISSUE #292: LinearDiffuser doesn't hold on to "g" when it already exists as a field
The below shows how the error can be reproduced:

from landlab.components import LinearDiffuser

rg = RasterModelGrid((4, 5), 10.0)
rg.add_zeros('node', 'topographic__elevation')
ld1 = LinearDiffuser(rg, linear_diffusivity=0.01)
ld1.diffuse(1.0)  # no problem
ld2 = LinearDiffuser(rg, linear_diffusivity=0.001)
ld2.diffuse(1.0)  # problem: gradient field already exists

  File "reproduce_diffuser_bug.py", line 9, in <module>
    ld2.diffuse(1.0)  # problem: gradient field already exists
  File "/Users/gtucker/Dev/Landlab/Clone/landlab/landlab/components/diffusion/diffusion.py", line 249, in diffuse
    self.grid.calculate_gradients_at_active_links(z)
AttributeError: 'LinearDiffuser' object has no attribute 'g'

Happens because __init__ doesn't create self.g if a gradient field already exists. A solution is to make self.g a reference to the existing field if it exists.

gregtucker:
Fix #292

### DIFF
--- a/landlab/components/diffusion/diffusion.py
+++ b/landlab/components/diffusion/diffusion.py
@@ -153,19 +153,19 @@ class LinearDiffuser(Component):
         self.interior_cells = self.grid.node_at_core_cell
 
         self.z = self.grid.at_node[self.values_to_diffuse]
-        g = self.grid.zeros(centering='link')
-        qs = self.grid.zeros(centering='link')
+        g = self.grid.zeros(at='link')
+        qs = self.grid.zeros(at='link')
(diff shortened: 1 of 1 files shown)


SiccarPoint:
Looks good.

PR #295, opened 2016-04-23
Gt flux divergence

gregtucker:
Adds modernized flux-divergence and net flux functions, for:
* faces => cells
* links => nodes
* faces => nodes


gregtucker:
I just discovered an issue with (at least) one of the divergence fns. Hold off on PR while I fix...

gregtucker:
@mcflugen: ok, this should now be good to go (assuming tests pass, which I expect they will)

PR #303, opened 2016-05-04
Update the gradient and divergence functions

gregtucker:
We now have `calc_grad_at_link` and `calc_flux_div_at_node`
Some (not all) of the other flavors now deprecated.


gregtucker:
I'm going to merge this PR so everyone who has components or programs that use methods like `calculate_gradients_at_active_links` and `calculate_flux_divergence_at_nodes` can replace them with `calc_grad_at_link` and `calc_flux_div_at_node`

PR #304, opened 2016-05-04
Fix docs not building and docstring formatting

mcflugen:
This pull request fixes an issue where documentation for certain modules was not being generated. In addition, it fixes many (mostly minor, but some major) formatting issues with docstrings throughout the code.

As I've touched just about everybody's files, this PR should probably be reviewed by most of the team. I shouldn't have changed anything major (content-wise) in the documentation so it should only take a quick look at the generated HTML to see if things are OK. It would probably be best to have a look at the generated HTML documentation, not just the docstings.


SiccarPoint:
Looks good to me.

PR #305, opened 2016-05-04
SiccarPoint/slopes

SiccarPoint:
Mends and modernises functionality to find GIS-style spatially averaged slopes at nodes.

Note the main method has changed name, mg.node_slopes() -> mg.calc_slope_of_node().

### DIFF
--- a/landlab/components/craters/craters.py
+++ b/landlab/components/craters/craters.py
@@ -409,7 +409,7 @@ class impactor(object):
                 break
             else:
                 pre_elev = data.elev[active_node]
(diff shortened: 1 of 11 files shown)


mcflugen:
Looks good. A couple suggestions, which can be fixed (if necessary) in another pull request.

My two cents:
* Angles should be in returned in radians. I don't think there's a need for a `unit` keyword. It's easy enough for a user to convert to degrees, if needed, and adds unnecessary complexity to our code.
* The `calc_slope_*` etc. methods should be broken out into separate modules (a la `gradients.py` and `divergence.py`).

SiccarPoint:
Let's do it separately, but both these sound reasonable to me.

PR #307, opened 2016-05-06
SiccarPoint/grid trawl

SiccarPoint:
This pull attempts the bulk of name modernisation on the grid, though I have probably missed a bit. It maintains many of the most widely used old methods, but deprecates them. Some efforts have been made to update uses in the code base of the old functions, but many uses of deprecated methods remain.

The most widely called of the old deprecated methods are:
1. calculate_gradients_at_d8_active_links (used in the flow router, so called A LOT)
2. _setup_active_inlink_and_outlink_matrices (not sure where this is lurking)


SiccarPoint:
Note that no modernisation has been attempted on the diagonal structures on the grid. This will need detailed attention in another branch.

PR #309, opened 2016-05-10
method kwd; correct sign error

SiccarPoint:
1. Add method={‘patch_mean’, ‘Horn’} to mg.calc_slope_of_node().
2. Correct a sign error in the slope direction where specified; it
should correspond the gradient, not the surface dip direction.
3. Better docs.


SiccarPoint:
Bump for merge

PR #310, opened 2016-05-11
Jadams15/uniform precip changes

jadams15:
updated FireGenerator, PrecipitationDistribution and DetachmentLtdErosion.

General fixes:
Removed extra / old files
Added doc strings and doc tests
Updated the component structure to match the new style


jadams15:
@SiccarPoint, I know your plate is super full at the moment, so I apologize for adding on. You are the one who has used the PrecipitationDistribution() component the most, so you'd be most affected by the changes here. I also had to change some of your unit tests too.

@saisiddu, I'm not sure if you've used the uniform precip component in a long time, but if so you should definitely pay attention to this pull request.

SiccarPoint:
Couple of things I found so far (and fixed/think is OK):
1. All our __init__ methods should terminate with `**kwds`. This lets us pass in a dict instead of argument-by-argument.
2. I get you don't declare any component standard properties as you don't take any, but I think you should still declare null entries.
3. You didn't declare the construction of the component.
4. Do you want to declare some "guideline" values for your arguments, rather than set to 0.? Consider None for 'value not provided' rather than 0. (I didn't do this, you could)

jadams15:
Thanks @SiccarPoint. Thank you for the work on points 1-3. As far as number 4, that is something I could do, but don't know if I want to prioritize right now. If I switched to None, nothing would fundamentally change in the way the component is used (just adding some tests to make sure that a value is provided), so I feel OK moving forward with this. If you have no other concerns, feel free to merge when ready.

SiccarPoint:
...these need to be listed somewhere, still. We mustn't end up hiding this functionality just because they lack component interfaces. Where can they be listed instead?

mcflugen:
I assume, but I haven't checked, they are still listed in the docs. They are also still imported into the components package (so `from landlab.components import FireGenerator` still works). I just removed them from the `COMPONENTS` variable since everything in that list is automatically wrapped with a BMI.

SiccarPoint:
That sounds good to me.

jadams15:
so, once AppVeyor is done for this, can it be merged?

SiccarPoint:
I believe so?

PR #311, opened 2016-05-11
Standardized radiation component

saisiddu:
@SiccarPoint : Can you please check my work on /components/radiation/radiation.py (added doc tests and standardized things) and the tests folder. Thanks!


mcflugen:
@saisiddu: While you're tidying things up here, could you remove the examples folder?

nicgaspar:
@mcflugen Are we not supposed to have examples folders in the components anymore?  I guess I vaguely remember this.  I know that it's not helpful for the release version, but it is helpful for developers trying to stay organized.  I guess these are supposed to go in the drivers repo, but I find this a bit awkward.

SiccarPoint:
Agree with Nicole that there's a place for having the examples folders - with the proviso that these are not expected to be used by end users.

mcflugen:
My thought is that if they really are "examples", they should go in our tutorials repository or, perhaps, a new "examples" repository. If they are used for development, they should go in a completely separate repository under the developer's account. If they can be used for tests, they should be turned into actual tests. Note that none of the code in the examples folders are tested with the rest of the code.

Keeping it in the main repository adds a lot of code that needs to be maintained and oftentimes results in old code that probably doesn't need to be maintained at all.

I think the best thing would be to turn this code into unit tests. But, unit then, my humble opinion is that it should find its way out of the main landlab library. Just my 2 cents.

nicgaspar:
Tutorials aren't tested right?  So that doesn't exactly solve the problem.  Plus I'm not a fan of the ipython notebooks, so that is not an easy way for me to easily write code to test it.  I think the the place that they should go is the drivers repo.  But when people are developing, that is awkward. 

I suppose we need to be better about updating the drivers folder once we are ready for prime time, which I guess is now.  Unit tests are certainly awesome, but I personally like to see an example file as well.

saisiddu:
Sure..  I will remove the examples.

mcflugen:
Oh. I think the tutorials and drivers repositories should be tested as well!

I'm fine leaving it up to developers. I would just ask that the example scripts are cleaned up and maintained even after component development is complete.

I really think the best would be to have developers write tests as they are developing their components rather than example scripts. The workflow would be almost exactly as it is now and we would end up with lots more tests.

SiccarPoint:
@mcflugen Agree with all this - but the issue is what to do about the legacy examples, not new ones. Let's leave them where they are for V1; we can remove them post-hoc (as we never refer to them, and users shouldn't even be aware they're there)

SiccarPoint:
@saisiddu You have failing tests? i.e., true fails, rather than odd ones?
```
FAIL: Doctest: landlab.components.radiation.radiation
  File "/anaconda/lib/python2.7/doctest.py", line 2226, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for landlab.components.radiation.radiation
  File "/Users/danhobley/git_landlab/landlab/components/radiation/radiation.py", line 0, in radiation

File "/Users/danhobley/git_landlab/landlab/components/radiation/radiation.py", line 28, in landlab.components.radiation.radiation
Failed example:
     Radiation.var_units('topographic__elevation')
Exception raised:
      File "/anaconda/lib/python2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "<doctest landlab.components.radiation.radiation[5]>", line 1
        Radiation.var_units('topographic__elevation')
        ^
    IndentationError: unexpected indent
File "/Users/danhobley/git_landlab/landlab/components/radiation/radiation.py", line 53, in landlab.components.radiation.radiation
Failed example:
    rad.output_var_names
Exception raised:
      File "/anaconda/lib/python2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "<doctest landlab.components.radiation.radiation[8]>", line 1, in <module>
        rad.output_var_names
    NameError: name 'rad' is not defined


FAIL: test_radiation.test_field_initialized_to_zero
  File "/anaconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/danhobley/git_landlab/landlab/components/radiation/tests/test_radiation.py", line 101, in test_field_initialized_to_zero
    assert_true(np.all(field == 0.))
AssertionError: False is not true

Ran 1152 tests in 10.598s

FAILED (failures=2)
```

SiccarPoint:
...Otherwise, this looks good from a first pass. Minor thing - we're now declaring the _input and _output_var_names as tuples, not sets (I'm doing it). Also, you're not PEP8 style compliant - particularly with regards to whitespace. Don't leave trailing blanks at the ends of lines or on blank lines, if you can help it.

SiccarPoint:
Other whitespace things - never whitespace after a parenthesis, or before a colon.

SiccarPoint:
Further style things:
* Take out your newlines (`\`) where you have them implied by use of brackets or parentheses.
* End lines with math symbols, don't start with them.
* When you go to a new continuing line, mostly it's either one tab in, or tab in to the position of the bracket defining the text block (hard to explain sorry! - take a look at Eric's code, or the PEP8 standards)

mcflugen:
(By "tab" we mean "four spaces" never use real tabs!)

A couple more comments (sorry):
* For that last failing test, use `assert_array_almost_equal(field, 0.)` (`from numpy.testing import assert_array_almost_equal`).
* Add `__init__.py` to your `tests` sub-package. Without it, the tests will not be run on an installed version of landlab (like on Travis).

PR #312, opened 2016-05-11
Remove craters component

mcflugen:
This pull request removes the craters component. According the @SiccarPoint, it was old code that shouldn't be included in the main repository. I've also removed the sample component, which was only used for some unit tests.


SiccarPoint:
Looks good lo me. Waiting on tests.

mcflugen:
@SiccarPoint: Notice the increase in coverage!

SiccarPoint:
This guy is still suffering from the stupid fail we discussed earlier. If you're doing the fixes elsewhere, please do this one too then pull it yourself, @mcflugen .

PR #313, opened 2016-05-11
Move slope calculation methods to gradients module

mcflugen:
This pull request moves the slope calculation functions from `base.py` into `gradients.py`. All of the functions are still available as methods of `ModelGrid` but are now standalone functions. This is mostly just an organizational move.

Added a new function, `radians_to_degrees` that converts math-style angles (radians, measured counter-clockwise from the positive x-axis) to compass-style angles (degress, measured clockwise from north).

Question for @SiccarPoint: `calc_aspect_of_node` can return angles in degrees or radians. However, when returning radians it gives them measured clockwise from north. Is this intended? It seems like it should be the usual CCW from x, no?


SiccarPoint:
To my knowledge, aspect is always measured CW from N. I've never seen it otherwise.

SiccarPoint:
@saisiddu may also have an opinion? He's been using aspect too.

mcflugen:
Even when it's given in radians? What do you think about me changing this? Just for consistency with other angles in landlab.

SiccarPoint:
Yes, even then. "In physical geology, aspect is the compass direction that a slope faces"

https://en.wikipedia.org/wiki/Aspect_(geography)

mcflugen:
OK. I'll leave it as-is, then. Based on that description, maybe we should remove the option for radians?

This is ready to be merged. Would you like to give it a quick review?

SiccarPoint:
I'm getting there. I seem to be reviewing everyone's code right now...

SiccarPoint:
I believe you have failed to move a few equivalent methods out of raster.py. Was this deliberate?

mcflugen:
Nope. Not deliberate. I'll do it.

SiccarPoint:
damn. Just realised that #309, which predates your work here, makes modifications to the functions you've moved.It's been good to go for a couple of days, but no-one's given it a merge. How would you feel about pulling that in too?

mcflugen:
OK. No problem. I'll merge #309 later today (if @saisiddu hasn't already) and then merge those changes into this pull request.

mcflugen:
@SiccarPoint: I merged your changes to the slope and aspect calculation methods into this branch. Will you please have a look to see that I didn't miss any changes. All the tests are passing so I'm fairly confident I got everything but it would still be good for you to have a look.

SiccarPoint:
This looks great to me. I'll merge when the tests complete.

mcflugen:
I'm still going to move the slope and aspect calc methods from `raster.py` into `raster_gradients.py`. Then it will be time to merge. It should be but a moment.

PR #316, opened 2016-05-12
Change of's to at's

gregtucker:
This is ready for a quick look


gregtucker:
I'm going to merge this "of-to-at" branch so we can keep moving forward. The Travis tests passed; AppVeyor is still queued.

PR #320, opened 2016-05-13
Saisiddu version 1 final

saisiddu:
This pull request should merge in standardized 'radiation' and 'potential evapotranspiration' components back to master. Tests were passing on local machine (for these two components).

Other Ecohydrology components will be added later (after this release).


mcflugen:
This looks good. I'm going to merge it. Most of my changes were superficial. Two notable ones are:
* line 353 of `potential_evapotranspiration_field.py` to maintain the reference to one of your fields.
* updated the expected output for your doctests after the change from _3.14_ to _np.pi_ This is a little dangerous. It would be good if you could double-check that things are still working.

PR #321, opened 2016-05-13
SiccarPoint/tweak plotter

SiccarPoint:
Adds a little more functionality to the imshow plotters


mcflugen:
This look good to me.

PR #322, opened 2016-05-13
SiccarPoint/patch methods

SiccarPoint:
This PR fixes the final outstanding issues with LL-standard ordering and behaviour around the patch methods. Primarily, `mg.patches_at_node` now is a property not method and returns a (nnodes, nmaxpatches) shape array with -1s for BAD_INDEX, which was not true before.


mcflugen:
I don't link the name `patches_present_at_node` (`patches_at_node` seems more inline with usage elsewhere) but, otherwise, this seems OK. I think we might want to think about this some more. I'm going to merge this but create an issue so we can discuss if that's an appropriate name.

SiccarPoint:
That’s not what that method is!! `patches_at_node` is its own thing, and already exists. It's a mask directly equivalent to `link_dirs_at_node`. i.e.,
```
>>> mg.patches_at_node * mg.patches_present_at_node  # is equivalent to
>>> mg.links_at_node * mg.link_dirs_at_node
```

SiccarPoint:
...it's for zeroing out patches on which one would not want to do a calculation, because the patch is incomplete OR contains closed nodes.

mcflugen:
OK. Still, we should discuss elsewhere. I suspect we can come up with a better name. I'm going to merge this.

SiccarPoint:
Sounds good

PR #323, opened 2016-05-13
SiccarPoint/BAD INDEX is minus1

SiccarPoint:
Sets the BIV to -1, and deals with the (limited) fallout.


mcflugen:
Since this is passing on Travis, I'm going to merge it.

PR #324, opened 2016-05-13
SiccarPoint/last minute deprecation

SiccarPoint:
This PR goes through the most obvious grid types and ensures anything that doesn’t look like it should be exposed to users, isn’t.

Exceptions are:
1. anything with calculate_, get_, or find_ at its start (these are decorated as deprecated, but not underscored)
2. vmg.active_links_ids, which I can’t find in the inheritance structure.


mcflugen:
I'm taking @SiccarPoint's word for it that this is all passing... merging now for inclusion in version 1.0.0.

PR #327, opened 2016-05-15
Remove usages of deprecated functions

mcflugen:
This pull request removes usages of many (but not all) deprecated functions/methods. Most of the remaining usages are either private methods, in doctests for the deprecated functions themselves, or not straightforward to convert to an alternative.

Note that in this pull request I have disabled warnings for methods that start with an underscore. This can easily be turned back on.

@SiccarPoint: One problem I had was with active links. The `active_links` attribute doesn't seem to ever include active *diagonal* links. This is a problem when replacing `grid.calc_grad_at_active_link()` with `calc_grad_at_link()[grid.active_links]`.

@saisiddu: In the radiation component when I switched from,
```python
self._slope, self._aspect = grid.calculate_slope_aspect_at_nodes_burrough(vals='topographic__elevation')
```
to
```python
elevation = grid.at_node['topographic__elevation']
self._slope = grid.calc_slope_at_node(elevs=elevation)[grid.core_nodes]
self._aspect = np.radians(grid.calc_aspect_at_node(elevs=elevation)[grid.core_nodes])
```
I got slightly different results so I stuck with the original. I still would like to figure out what's going on there. Do you have any ideas?

### DIFF
--- a/landlab/components/chi_index/channel_chi.py
+++ b/landlab/components/chi_index/channel_chi.py
@@ -541,13 +541,13 @@ class ChiFinder(Component):
         """
         Returns a masked array version of the 'channel__chi_index' field.
         This enables easier plotting of the values with
-        :func:`landlab.imshow_node_grid` or similar.
+        :func:`landlab.imshow_grid_at_node` or similar.
 
         Examples
         --------
         Make a topographic map with an overlay of chi values:
 
-        >>> from landlab import imshow_node_grid
+        >>> from landlab import imshow_grid_at_node
         >>> from landlab import RasterModelGrid, CLOSED_BOUNDARY
         >>> from landlab.components import FlowRouter, FastscapeEroder
         >>> mg = RasterModelGrid((5, 5), 100.)
@@ -570,9 +570,9 @@ class ChiFinder(Component):
         >>> _ = fr.route_flow()
         >>> cf.calculate_chi()
 
-        >>> imshow_node_grid(mg, 'topographic__elevation',
-        ...                  allow_colorbar=False)
-        >>> imshow_node_grid(mg, cf.masked_chi_indices,
-        ...                  color_for_closed=None, cmap='winter')
+        >>> imshow_grid_at_node(mg, 'topographic__elevation',
+        ...                     allow_colorbar=False)
+        >>> imshow_grid_at_node(mg, cf.masked_chi_indices,
+        ...                     color_for_closed=None, cmap='winter')
         """
         return np.ma.array(self.chi_indices, mask=self.hillslope_mask)
--- a/landlab/components/diffusion/diffusion.py
+++ b/landlab/components/diffusion/diffusion.py
@@ -252,8 +252,7 @@ class LinearDiffuser(Component):
                                                self.g[self.grid.active_links])
 
             # Calculate the net deposition/erosion rate at each node
-            self.dqsds = self.grid.calculate_flux_divergence_at_nodes(
-                self.qs[self.grid.active_links])
+            self.dqsds = self.grid.calc_flux_div_at_node(self.qs)
             # Calculate the total rate of elevation change
             dzdt = - self.dqsds
 
--- a/landlab/components/diffusion/diffusion.py
+++ b/landlab/components/diffusion/diffusion.py
@@ -246,7 +246,7 @@ class LinearDiffuser(Component):
         for i in range(repeats+1):
             # Calculate the gradients and sediment fluxes
             self.g[self.grid.active_links] = \
-                self.grid.calc_grad_at_active_link(z)
+                    self.grid.calc_grad_at_link(z)[self.grid.active_links]
             # if diffusivity is an array, self.kd is already active_links-long
             self.qs[self.grid.active_links] = (-kd_activelinks *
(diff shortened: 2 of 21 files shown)


SiccarPoint:
This issue is better addressed to @gregtucker ...

gregtucker:
@mcflugen, @siccarpoint: `grid.active_links` is a really essential component. It should always exist. What are the circumstances that lead it to be missing?

gregtucker:
So I just checked. With a simple 3x4 raster, `active_links` seems fine: it's an array with links 4, 5 (bottom vertical), 7, 8, 9 (middle horiz), 11, 12 (top vert).

If I run `rg.calc_grad_at_link(z)[rg.active_links]` it produces the correct result.

Here's the code that should work:

>>> rg = RasterModelGrid((3, 4), 10.0)
>>> rg.active_links
array([ 4,  5,  7,  8,  9, 11, 12])
>>> import numpy as np
>>> z = np.arange(rg.number_of_nodes)
>>> rg.calc_grad_at_link(z)[rg.active_links]
array([ 0.4,  0.4,  0.1,  0.1,  0.1,  0.4,  0.4])

mcflugen:
Sorry. I wasn't clear. `active_links` works almost all of the time. It's just with diagonal "links" that it's not "working". If your `RasterModelGrid` contains diagonal links, they aren't included in the list. I believe they are only added to the `active_inlink` and `active_outlink` data structures. I don't know that this is a bug. It may be the expected behavior. It's just that, if you're using diagonal links, the suggested substitution from the deprecated function to the new, doesn't work.

SiccarPoint:
My efforts with the diagonals should hopefully have made it impossible for a user to get a data structure returned to them that has references to diagonals in it. Or at least that was the intent. They shouldn’t have access to active_inlink, for example (which I hope is now _active_inlink IIRC). So I believe `active_links` is working correctly (there’s a now hidden `_d8_active_links` too).

The only exception I think may be `mg.length_of_link`, which gets diagonals tacked on the end if they’re present, and maybe one or two other `X_at_Y` properties. However, this won’t be an issue as a user will always be indexing that array with IDs, so the diagonals won’t ever get extracted. We’ll just have to hope no-one goes exploring in the code and gets themselves confused.

Dealing with the diagonals needs to be a top priority for after CSDMS.

mcflugen:
You hit on the problem exactly. The problem is in `route_flow_dn.py`. If I change
```python
link_slope = - self._grid.calc_grad_of_active_link(elevs)
```
to
```python
grad_at_link = self._grid.calc_grad_at_link(elevs)
link_slope = - grad_at_link[self._grid.active_links]
```
I get an array size mismatch error in `raster_gradients.py` when doing the following,
```python
grads /= grid.length_of_link
```

SiccarPoint:
And this is *exactly* the reason why I backed off this change, which I found last week! I would strongly advise you not to attempt to remove that method until we have implemented a diagonal link grid class. This really is not a trivial change, and is probably subject to further revision once we have that class.

Just make the `calc_grad_of_active_link` method internal for now? And note this problem in the docstring?

mcflugen:
I've left `calc_grad_of_active_link` in place in `route_flow_dn.py`. I have, however, replaced it elsewhere. It's not used all that much. The only other places it's used are:
* `generate_overland_flow_Bates.py`
* `generate_overland_flow_deAlmeida.py`

mcflugen:
I guess my question (and maybe this is a question for @jadams15) is, are my changes to `generate_overland_flow_Bates.py`, and `generate_overland_flow_deAlmeida.py` OK? It seems like it should be fine as both of those modules make use of `active_links` elsewhere.

By the way, my purpose for getting rid of the deprecation warnings is not just because it doesn't look good to have our own code produce warnings but that our tests are producing so many warnings that it causes some of our Travis builds to hang when creating releases. Adding a few back in won't make a huge difference.

SiccarPoint:
Did you manage to deal with the calls to `create_node_inlink_and_outlink_matrix`? They were THE WORST

mcflugen:
Oh sure! By suppressing their warnings!

SiccarPoint:
Yesssssss

mcflugen:
@gregtucker: Could you please have a look at this? Maybe make sure your tutorials still run fine? As long as you use the new functions, it should to fine. I would like to get this merged as it should help with some of our release builds hanging.

gregtucker:
This looks great to me. In a couple of tutorials I changed `imshow_node_grid` to `imshow_grid_at_node`. Some tutorials simply use `imshow_grid`. Is this still current? I didn't see it in the code you modified.

Do you want me to merge it?

mcflugen:
@gregtucker: Thanks! I prefer `imshow_grid` rather than `imshow_grid_at_node` but either is fine. I'll merge it.

gregtucker:
Oh, one thing that's relevant to @jadams15 : in the overland flow component, calculating water_surface_gradient at one array length, then resetting it to a different length, is costly in terms of compute time (and maybe memory, depending on how Python is handling garbage collection). Suggest changing this to a one-linear.

mcflugen:
@gregtucker: good catch. That was my fault. I fixed it. (@jadams15 you might still have a look, just to double-check, but it should be fine).

PR #351, opened 2016-06-22
SiccarPoint/bc cond flag

SiccarPoint:
This pull creates a new ModelGrid property, `BC_set_code`. It is a unique float that changes each time boundary conditions are modified on the grid.

This permits automated and efficient dynamic updating of privately held methods in components that in turn depend on the grid BCs.

### DIFF
--- a/landlab/components/diffusion/diffusion.py
+++ b/landlab/components/diffusion/diffusion.py
@@ -160,6 +160,7 @@ class LinearDiffuser(Component):
     def __init__(self, grid, linear_diffusivity=None, method='simple',
                  **kwds):
         self._grid = grid
+        self._BC_set_code = self.grid.BC_set_code
         assert method in ('simple', 'resolve_on_patches', 'on_diagonals')
         if method == 'resolve_on_patches':
             assert isinstance(self.grid, RasterModelGrid)
@@ -399,6 +400,9 @@ class LinearDiffuser(Component):
         if not self._run_before:
             self.updated_boundary_conditions()  # just in case
             self._run_before = True
+        if self._BC_set_code != self.grid.BC_set_code:
+            self.updated_boundary_conditions()
+            self._BC_set_code = self.grid.BC_set_code
 
         core_nodes = self.grid.node_at_core_cell
         # do mapping of array kd here, in case it points at an updating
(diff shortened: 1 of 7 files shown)


mcflugen:
A couple suggestions:
* `BC_set_code` should be `bc_set_code`, in my opinion. All lowercase looks better (in my opinion) and follows our coding-style standards.
* Since `BC_set_code` is a float, `==` and `!=` comparisons are unsafe. Better to use something along the lines of `np.isclose`.

SiccarPoint:
Agree on the rename.

Considered not using == etc, but did this deliberately. These aren’t calculations, they’re essentially arbitrary strings, which may occasionally get copied. AFAIK, this kind of precision problem with floats can’t arise copying, right? There could equally be issues with isclose, if Landlab is looping fast.

mcflugen:
If `bc_set_code` can be a string, then `isclose` definitely won't work. However, it appears you set it either to 0. or to decimal seconds since the epoch. How about converting the time *as-a-float* to time *as-an-integer* (either to the nearest second, or millisecond, or micosecond)? Would that work? Integer comparisons would be faster and safer. As you say, though, *safer* may not actually be an issue in this particular case.

nicgaspar:
I actually liked BC_set_code because BC represents two words.  But I’m not super invested in that.

I’m confused.  I thought bc_set_code was a float, so wouldn’t isclose work?  Would there be an issues with a component that has super short time steps and for some reason is constantly updating boundary conditions?  What would that component be? Not sure but...

SiccarPoint:
I'm shifting back onto github, as that was getting confusing. @nicgaspar - per discussion occurring in tandem to this on slack, we've now migrated to a int counter scheme, which makes some of this redundant. `isclose` would probably have been OK, but if BCs were updated very rapidly, there would have been a chance this check wouldn't trigger. @mcflugen - (from Slack) It's not a true flag, as we want to be able to distinguish the sequence of changes, not just changed/unchanged. The counter scheme we have works, and I hope to leave it that way now.

nicgaspar:
OK.  Been away from my computer so missed the slack side.  I’ll stay out.

mcflugen:
@SiccarPoint: sounds good. The counter method sounds like the best way to go. What do you think of a name change, though? Something that better describes what it is. `bc_update_count`? I don't know.

PR #356, opened 2016-07-07
SiccarPoint/angle of link

SiccarPoint:
Modify code such that `angle_of_link` is a property not a function. The
ability to specify the end of the link about which to measure angle is
replaced by a new property, `angle_of_link_about_head` (`angle_of_link`
is about the tail end, by default).

Modifications were only necessary in one place to accommodate this new
style.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -114,6 +114,7 @@ Fluvial geomorphology
    :maxdepth: 4
 
    landlab.components.stream_power
+   landlab.components.detachment_ltd_erosion
 
 Flow routing
 ------------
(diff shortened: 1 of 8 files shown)


gregtucker:
This looks good to me, and I'm about to merge. But I noticed that the x64 appveyor tests are failing (it did on my lake_filler bug fix too). At first I thought this was just some one-off appveyor problem, but I'm wondering if it isn't a problem with our build? @mcflugen, any ideas about why appveyor can't find cl.exe?

SiccarPoint:
Thanks Greg!

That fail happens over and over again, but seemingly at random. Not sure what’s going on.

PR #357, opened 2016-07-09
Gt cythonize cts

gregtucker:
Assuming the tests pass, I'd like to merge this. @mcflugen, you have the most experience with cython -- could you take a look?

### DIFF
--- a/landlab/ca/celllab_cts.py
+++ b/landlab/ca/celllab_cts.py
@@ -128,8 +128,8 @@ from heapq import heappop
 import landlab
 import numpy
 import pylab as plt
-from numpy import zeros
-from .cfuncs import update_node_states_cython
+#from numpy import zeros
+from .cfuncs import update_node_states_cython, update_link_state
 
 _NEVER = 1e50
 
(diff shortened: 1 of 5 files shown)


mcflugen:
A made a couple teeny changes.

* I think you might be able to get a bit more of a speed-up if you can get rid of all the Python data types in your Cython function arguments (`node_pair`, for example). Same goes for the return values of your functions. `update_node_states` returns a tuple. Even that could be a problem.
* If possible, all the variables within your functions should be typed (`i`, `current_pair`, etc.). Having non-typed variables or variables of Python types (like tuples) in you loops are potential performance killers.
* How about dropping the `_USE_CYTHON` variable? I guess it's good to keep around for a bit so you can compare speed ups.
* It would be nice, but certainly not essential, if you could break up some of the functions in `cfuncs.pyx` as they're a little long and difficult to follow (my rule-of-thumb is to keep functions less than 20 lines).

gregtucker:
Thanks Eric. I'll see what I can do about the non-C data types. Please hold off on merging for now.

PR #359, opened 2016-07-12
EcoHyd ver 1

saisiddu:
Please verify the updates to the components. Now all ecohydrology components should be compatible with version 1 Landlab. Also, unit tests and doc tests have been added to all ecohydrology components.

Please merge this ASAP (if the tests pass). Thank you!


SiccarPoint:
Don't have time for this right now Sai, sorry...

SiccarPoint:
I'm scanning the fields right now - some of these could maybe be tweaked:

* none of your "radiation" fields include a quantity. These are fluxes, right? So probably e.g. `radiation__net_shortwave_flux`, rather than just `radiation__net_shortwave`.
* `precipitation__rain` feels pretty weird as a name. Surely either `precipitation__rate` or `rainfall__rate`, or something. At any rate, it needs some kind of measurement quantity in there (i.e., "rate").
* You have both `surface__evapotranspiration_rate` and `surface__potential_evapotranspiration_rate`. This is clearly fine if they are indeed different quantities (I don't know anything about PET), but wanted to check.
* Your `...biomass` fields aren't (for me) intuitive units. I was expecting, well, a mass, but I see the unit is g DM m^-2 d^-1. So, a. I can't actually decipher that unit, and b. As it's not simply grams, this must be a `biomass_density` of some kind, right?

I'm going to spend another few minutes scanning the actual code, but I don't foresee many actual issues.

saisiddu:
- incoming short wave radiation is 'flux' by default (jargon). Hence, it would be weird calling it flux again.
- precipitation can be rain or snow even though we are not dealing with snow in this code. hence, i gave that name.
- Yes, potential evapotranspiration is different from actual evapotranspiration.
- biomass is commonly known as biomass with the units specified in Ecohydrology community.

Can we leave the names as is for now?

SiccarPoint:
Other stuff:
* your methods should be standardised with a method called `run_one_step`, not `update`. Though given you're currently passing a `current_time`, not a `dt` in most cases, perhaps better to leave it like this for now. You can add a `run_one_step(dt)`-style signature later, at your leisure.
* I don't believe there's a test in place for the actual `update` functions for `plant_competition_ca`. Your coverage will improve a lot if you just make a tiny grid with some simple initial conditions, then update once, then check the solution. I think you do this elsewhere, but missed it here? Again, this is super minor any you could leave it for now if you wanted.

...and that looks good to me. I didn't actually check the code itself, but the style looks great.

SiccarPoint:
Uh… better to do it now before we expose it to the public. The latter two are clearly fine. But “radiation” does not imply a unit (Grays would be the actual default, I believe!?) and precipitation__rain is just horrible. Just `rainfall__rate` would be fine. “Rain” implies “precipitation”.

These will be simple search-and-replaces. It will only take a minute to switch out...
[This comment duplicated here from Slack for completeness]

saisiddu:
Yep, changed them. 

- I will work on the name of *update* function later. 

Thank you Dan!

SiccarPoint:
Looks good Sai! However, I totally just noticed `precipitation__rain` is not a rate, is it? It's a depth/amount. So maybe "flux" isn't right. Maybe `rainfall__total`, `rainfall__amount`, something like that?

erkanistan:
precipitation can be in the form of snow and rain..

SiccarPoint:
@erkanistan So it's either `precipitation` or `rainfall`. It shouldn't be both, was my point.

saisiddu:
@SiccarPoint I have changed it to 'rainfall__daily'.

SiccarPoint:
Yes! Looking good.

SiccarPoint:
Let's see at least one successful test, then I'll merge.

PR #361, opened 2016-07-12
SiccarPoint/lake mapper irreg

SiccarPoint:
This pull:
1. Removes some very old grid methods related to flow routing, which were very poor performing, used old data structures, and use out-of-date style signatures
2. Removes some of Nicole’s old component work, which was the only place these methods got used.
3. Makes `active_neighbors_at_node` a property in RasterModelGrid.
4. Creates `active_neighbors_at_node` in base.py.
5. Modifies DepressionFinderAndRouter to use this new functionality to be able to run on irregular grids.


SiccarPoint:
I assigned @nicgaspar as it was I think mainly your component work that I excised. Check I did it OK!

PR #362, opened 2016-07-12
Merging 'Landslide' component

RondaStrauch:
Please review and merge. I am seeing some issues that I am not sure how to fix. Can you please suggest edits to fix the errors? 

Ronda


SiccarPoint:
Just took a very quick scan through this at a high level. Looking pretty nice! Consider however:
1. Further standardisation of your output field names. Lower case only, and the names are probably a bit too vague. i.e., `relative wetness__mean` does not tell you what is wet, probably should be `soil__mean_relative_wetness` or something; `probability_of_failure` -> `landslide__failure_probability`? `factor_of_safety` is probably OK, but "distribution" is not very clear - do you mean a standard deviation? Remember, field names get seen without the context of the component that creates them, so the user has no clues they are relevant to land sliding unless they explicitly refer to it.
2. Docstrings go immediately under the method declaration, which isn't what you have at lines 343-356 in landslide.py.
3. `calculate_factor_of_safety` needs a more comprehensive docstring. Also needs to be under the declaration. It's not clear how one goes about running it. Consider having it be just `calculate_factor_of_safety()`, so the FoS is calculated for every node at once. (You claim `i` is optional, but I believe the method will currently fail if it isn't provided, as there's no default value declared).
4. Landlab standards suggest you should ideally call `update` `run_one_step`. This method also needs a docstring. What does it do?

If you clean up this documentation-type stuff, actually assessing the code (and resolving those test fails) will be easier.

RondaStrauch:
Thanks...these are great comments/suggestions.  I'll work on these and resubmit.
Ronda

RondaStrauch:
Thanks for review and pushing!
Ronda

PR #366, opened 2016-07-18
Gt fix lake bug

gregtucker:
This PR just adds some diagnostic output for the case when the lake-fill algorithm can't find an outlet. This can happen if, for example, the user accidentally passes it a watershed DEM with no outlet node(s) specified (i.e., surrounded by closed nodes), or if there is an isolated pocket of core node(s) surrounded by closed boundary nodes.


mcflugen:
We should standardize how we print these sort of messages in landlab and what the message contents look like. In my opinion, what we have now looks a little sloppy. For warnings, maybe use the `warnings` module? We could also use the `logging` module, which would give us more flexibility.

Anyway, that's another topic for discussion for another time. Until then, how about printing your warning messages printed to `stderr` rather than `stdout`?

Issue #333, opened 2016-05-19
gflex warning coming up when working on tutorial overland flow driver


kettner:
Just want to report a warning, indicating that gflex isn't loaded when importing libraries defined in tutorial overland flow driver. See also screendump.
Fun playing though with landlab.
cheers, Albert.

![screen shot 2016-05-19 at 12 11 32 pm](https://cloud.githubusercontent.com/assets/7075202/15404514/0fa3fafa-1dbb-11e6-9c8b-18c79162a441.png)

SiccarPoint:
Thanks @kettner !

mcflugen:
This is fixed by #367.

PR #367, opened 2016-07-19
Strip gFlex warning from launch if no gFlex present

SiccarPoint:
This pull removes the annoying warning for users who haven’t installed gFlex. (I think)


SiccarPoint:
Assigned Eric, as this seems pertinent to the Warnings discussion we just had!

mcflugen:
Looks good. This fixes #333.

PR #369, opened 2016-07-21
Gt add fn to hex

gregtucker:
This PR simply adds a new method nodes_at_column() to HexModelGrid, and replaces some string comparisons with character comparisons. @mcflugen, could you take a quick look please?

### DIFF
--- a/landlab/grid/hex.py
+++ b/landlab/grid/hex.py
@@ -563,12 +563,12 @@ class HexModelGrid(VoronoiDelaunayGrid):
             'shape must be either "hex" (default) or "rect"'
 
         # Create a set of hexagonally arranged points. These will be our nodes.
-        if orientation == 'horizontal' and shape == 'hex':
+        if orientation[0].lower() == 'h' and shape[0].lower() == 'h':
             pts = HexModelGrid._hex_points_with_horizontal_hex(
                 base_num_rows, base_num_cols, dx)
             self.orientation = 'horizontal'
             self._nrows = base_num_rows
-        elif orientation == 'horizontal' and shape == 'rect':
+        elif orientation[0].lower() == 'h' and shape[0].lower() == 'r':
             pts = HexModelGrid._hex_points_with_horizontal_rect(
                 base_num_rows, base_num_cols, dx)
             self.orientation = 'horizontal'
@@ -577,7 +577,7 @@ class HexModelGrid(VoronoiDelaunayGrid):
             self._shape = (self._nrows, self._ncols)
             self._nodegrid = numpy.arange(self._nrows * self._ncols,
                                        dtype=int).reshape(self._shape)
-        elif orientation == 'vertical' and shape == 'hex':
+        elif orientation[0].lower() == 'v' and shape[0].lower() == 'h':
             pts = HexModelGrid._hex_points_with_vertical_hex(
                 base_num_rows, base_num_cols, dx)
             self.orientation = 'vertical'
@@ -930,7 +930,7 @@ class HexModelGrid(VoronoiDelaunayGrid):
         >>> from landlab import HexModelGrid
         >>> grid = HexModelGrid(3, 4, shape='rect')
         >>> grid.nodes[grid.nodes_at_top_edge]
-        array([ 8, 9, 10, 11])
+        array([ 8,  9, 10, 11])
         """
         try:
             return self._nodegrid[-1, :]
@@ -955,6 +955,38 @@ class HexModelGrid(VoronoiDelaunayGrid):
         except AttributeError:
             raise AttributeError(
                 'Only rectangular Hex grids have defined edges.')
+
+    def nodes_at_column(self, col):
+        """Return array of node IDs in given column.
+        
+        Examples
+        --------
+        >>> import numpy as np
+        >>> from landlab import HexModelGrid
+        >>> grid = HexModelGrid(3, 5, shape='rect', orientation='vert')
+        >>> grid.nodes_at_column(3)
+        array([ 4,  9, 14])
+        >>> grid.nodes_at_column(4)
+        array([ 2,  7, 12])
+        >>> grid = HexModelGrid(3, 6, shape='rect', orientation='vert')
+        >>> grid.nodes_at_column(3)
+        array([ 4, 10, 16])
+        >>> grid.nodes_at_column(4)
+        array([ 2,  8, 14])
+        >>> grid = HexModelGrid(3, 5, shape='rect', orientation='horiz')
+        >>> grid.nodes_at_column(2)
+        array([ 2,  7, 12])
+        """
+        try:
+            (num_rows, num_cols) = self._nodegrid.shape
+        except AttributeError:
+            raise AttributeError(
+                'Only rectangular Hex grids have defined columns.')
+        if self.orientation[0] == 'v':
+            base_node = (col // 2) + (col % 2) * ((num_cols + 1) // 2)
+            return numpy.arange(base_node, self.number_of_nodes, num_cols)
+        else:
+            return self._nodegrid[:, col]
 
     def _configure_hexplot(self, data, data_label=None, color_map=None):
         """
--- a/landlab/grid/hex.py
+++ b/landlab/grid/hex.py
@@ -576,9 +576,8 @@ class HexModelGrid(VoronoiDelaunayGrid):
             self._nrows = base_num_rows
             self._ncols = base_num_cols
             self._shape = (self._nrows, self._ncols)
-            self._nodegrid = numpy.arange(self._nrows * self._ncols,
-                                          dtype=int).reshape(self._shape)
-            self._nodes = self._nodegrid
+            self._nodes = numpy.arange(self._nrows * self._ncols,
+                                       dtype=int).reshape(self._shape)
         elif orientation[0].lower() == 'v' and shape[0].lower() == 'h':
             pts = HexModelGrid._hex_points_with_vertical_hex(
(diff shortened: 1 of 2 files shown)


SiccarPoint:
I saw this go in, so took a look. I think `nodes_at_column` should be a property, in keeping with, e.g., `rmg.nodes` on the Raster. So it would handle like this:
```
>>> hmg.nodes_at_column[0]
...
```

You should probably also create the equivalent `nodes_at_row` structure too.

SiccarPoint:
It might also be worth actually reshaping `hmg.nodes` under the hood (a la RasterModelGrid) to be x-by-y, so you can do these operations just by returning a slice.

SiccarPoint:
...finally, those failing tests you have on Appveyor will go away if you use one of our id decorators or functions on the method. e.g., `as_id_array` (wrapping function), `return_readonly_id_array` (decorator).

mcflugen:
What @SiccarPoint said.

Maybe a couple additional things.

* In you comparisons for `orientation`, how about something like:
```python
if 'horizontal'.startswith(orientation):
    ...
```
I also wouldn't bother converting to lower case as it just adds extra code. Make lowercase mandatory.
* Don't use `assert` when checking for valid orientation strings. `assert` should only be used for debugging. Users can turn assert statements off so that they are completely ignored.

Just suggestions.

gregtucker:
Thanks @mcflugen and @siccarpoint. Messing with _nodes turns out to break other stuff, which I suppose I could fix if I had more time. Instead, I have tried to adopt the spirit of your suggestions by replacing my nodes_at_column with "nodegrid", which is a new property that simply returns _nodegrid (which already existed, so it isn't adding new memory overhead, though I'm not sure why we need both _nodes and _nodegrid -- did one of you create the latter?).

Also, I modified the hex constructor so that _nodegrid is now properly assembled for a vertically oriented, shape='rect' hex grid (you have to account for staggering in IDs).

Dan: sorry, not clear what you mean about Appveyor tests... maybe this update will fix whatever it was...

mcflugen:
Your tests for `node_at_column` were failing, which is what Dan was referring to. Now that you've removed it, things are passing again. You should decorate your `nodegrid` property with the `return_readonly_id_array` to ensure that it returns an array of `int`s (whatever that means on the target platform - sometimes `int64`, sometimes `int32`) and that the array is read-only (to prevent users from changing the values in `_nodegrid`).

SiccarPoint:
I just added the decorator...

SiccarPoint:
We have both _nodes and _nodegrid because I couldn't figure out a clean way (i.e., a view) of transforming _nodes into the 2d _nodegrid structure. It might be more sensible to go the other way now (i.e., _nodegrid.ravel() to give _nodes, if that is indeed a clean operation).

mcflugen:
Wouldn't,
```python
self._nodegrid = self._nodes.reshape((n_rows, n_cols))
```
work?

SiccarPoint:
Not as right now
```
>>> hmg = HexModelGrid(4, 4, shape='rect', orientation='vertical')
>>> hmg.nodegrid
array([[ 0,  2,  1,  3],
       [ 4,  6,  5,  7],
       [ 8, 10,  9, 11],
       [12, 14, 13, 15]])
```
...which I'm not actually sure I like (@gregtucker?)

SiccarPoint:
@mcflugen The other issue was that (IIRC) I didn't like that _nodegrid wasn't always defined (only for `rect` hexes) so using a view sometimes but not other times seemed like a recipe for confusion. But I'm not sure if that holds up.

SiccarPoint:
I think I'd prefer what Eric is suggesting here, where it's like I had it defined before without the alternations for a vertical grid. And indeed, `self._nodegrid = self._nodes.reshape((n_rows, n_cols))` would seem a good way to save memory.

mcflugen:
Ah. Got it. Thanks. It seems like `_nodes` isn't being created properly, then. But, as @gregtucker says, messing with it causes other problems. I don't think it should though - so maybe there is a problem somewhere.

SiccarPoint:
Those errors related to `_node` could almost certainly be fixed by replacing them where they occur with a `_node.ravel()`. It's happening because the rest of the code assumes that when a Voronoi grid asks for a node list, it will be 1d. (As long as the 2d node structure is always in ID order).

mcflugen:
Yeah, `nodegrid` should only be defined for grids whose nodes can be represented by a matrix (like structured quads, etc.).

SiccarPoint:
AFAIK `_nodegrid` only exists in this one place, in Hex. Everywhere else it's just `_nodes`, which is 2D for a raster and 1D for a voronoi.

mcflugen:
Well, if that's really the only problem, maybe we should do something like that.

SiccarPoint:
Think that would make sense - as long as @gregtucker isn't wedded to this alternating ID scheme for a vertical hex.

gregtucker:
The issue here is that the rows of a vertically oriented hex grid have staggered numbering. @siccarpoint's code snippet above is the correct numbering scheme. If you used another (simpler) scheme it would simply not be correct. This is because odd-numbered columns are staggered upward. So, for example, if you had a 3x5 grid, the bottom-most nodes in each column would be numbered: 0, 3, 1, 4, 2. I have an application where I need to march through the columns (to construct a topographic profile from the Grain Hill model).

gregtucker:
However it seems wasteful to have both _nodes and _nodegrid. Suppose we made _nodes 2D for rectangular hex, ordered as appropriate for the orientation, and flattened it where needed. Would the irregular ordering (for vertically oriented rect hex) cause problems? I suppose there's only one way to find out...

SiccarPoint:
I get this now. Makes sense. As you say, let's try just flattening it and see what happens...

gregtucker:
@siccarpoint I've now removed _nodegrid from hex, in favor of using _nodes (which always exists, so we gain simplicity and save memory).

To get it to work, had to flatten `nodes` (only) when it's 2D in voronoi._create_neighbors() (flattening it always does NOT work).

The only test fail now is in the potentiality flow router. Can you take a look at:

```File "/Users/gtucker/Dev/Landlab/Clone/landlab/landlab/components/potentiality_flowrouting/route_flow_by_boundary.py", line 94, in landlab.components.potentiality_flowrouting.route_flow_by_boundary.PotentialityFlowRouter
Failed example:
    np.allclose(mg.at_node['water__discharge'][mg.core_nodes],
                Q_at_core_nodes)
Expected:
    True
Got:
    False```

**Comment by mcflugen** [comment]
Yes, sorry. I misunderstood what `shape='rect'` meant (I think). A shaped `nodegrid` really only makes sense for a structured grid of squares. It's confusing for a hex grid, for me at least.

I'm still missing something, though. How would you make `_nodes` 2D - since even and odd numbered rows (or columns, depending on the orientation) have different numbers of nodes? I guess the answer is that what you are defining a "row" or a "column" (again, depending on the orientation) is actually serpentine (that is, they don't share a common x or y).

What we have now is,
```python
nodes = array([[ 0,  2,  1,  3],
               [ 4,  6,  5,  7],
               [ 8, 10,  9, 11],
               [12, 14, 13, 15]])
```
but maybe something like this could work (which, I think, is what @SiccarPoint was suggesting)?
```python
nodes_at_row = array([[ 0,  1],
                      [ 2,  3],
                      [ 4,  5],
                      [ 6,  7],
                      [ 8,  9],
                      [10, 11],
                      [12, 13],
                      [14, 15]])
nodes_at_column = array([[ 0,  4,  8, 12],
                         [ 2,  6, 10, 14],
                         [ 1,  5,  9, 13],
                         [ 3,  7, 11, 15]])
```
`nodes_at_column` would just be a view of `nodes` but `nodes_at_row` would be an additional array. Or, with wavy rows, `nodes_at_row` would be the same as `nodes`.

This is getting sidetracked but, what do you all think of having `nodes_at_column` and `nodes_at_row` properties for `RasterModelGrid` and similar grids. It seems less confusing in its name than just `nodes`. They would both be views of `nodes` so there wouldn't be any additional memory and `nodes` would be 1D for all grids.

SiccarPoint:
Now I’m on top of it, Greg’s current way looks the most useful way to have it. The first row is simply the first node in each column. I’ve previously looked for a way to get at these structures (hence _nodegrid existing).

I’ll take a look at the failing test.

SiccarPoint:
Update: The problem is not in the `PotentialityFlowrouter`. I believe you've broken the logic in `voronoi._create_neighbors`. Lucky we caught this - I think that place in the PotFR is the only place in the whole of LL we're actually testing a Hex in the configuration that would produce a fail. There's a lesson in here...

SiccarPoint:
Got it. Was a minor fix. I did some searching, and am fairly sure this is the only place in the grid this could be a problem.

From my perspective this merge is now good to go. I don't necessarily dislike Eric's idea here on separating these out, but it should probably be a separate thing? I also don't mind how it is now. And then we're basically back to where we were with the HexModelGrid, storing basically the same data twice for little concrete gain (at least in the rect Hex case).

gregtucker:
Thanks @siccarpoint.
@mcflugen : you're right, the shape issue is that, with a rect-shape hex, one direction has aligned nodes, and the other direction has them in a wiggly pattern. I can imagine there are cases in which you'd want either or both (my current application needs the aligned ones).
I agree that nodes_at_column and nodes_at_row would be a good idea if they return slices (views) rather than consuming their own memory. But let's make that a separate issue. I'm going to go ahead and merge. Thanks again both for looking carefully at this. It was really meant to be just a quick fix that I needed, but became a good example of the benefits of "doing it properly".

PR #370, opened 2016-07-21
adding drainage density component for real this time

cmshobe:


SiccarPoint:
I've taken a look at this. I've got a merge pending for Charlie's fork that he can take a look at once he gets back from his holidays after next week. It should get this to fully ready to merge status.

SiccarPoint:
Actually, this needs someone to take a look at the standard names it implements. Probably @mcflugen ?

mcflugen:
@cmshobe: Thank you! This looks great!

I've created a pull request to your landlab fork with some small changes to include in this pull request (if they're fine by you, of course).

A couple other things:

*  As @SiccarPoint mentioned, your field names could be improved. `channel_network` is really the only one. I think the others are already used by other components. Is that correct?
*  Because the `channel_network_name` keyword can change the name of the field for the channel network values, including `channel_network` in `_input_var_names` will be incorrect if the user chooses another name. The easiest thing would be to remove that keyword. What do you think about that?
*  I think it would be great if there was a function (not a method of `DrainageDensity`) that calculated the drainage density of a channel network. So, something like,
```python
mean_drainage_density = calc_drainage_density(grid)
```
This function would probably have to call other components (`FlowRouter`, maybe?) as well as `DrainageDensity`. Anyway, it seems like that's something users would want to use. Is this possible? It seems like it would, basically, just be wrapping your doctest into a function.

SiccarPoint:
1. Re the names, I think both `channel_network` and `distance_to_channel` need an update.
2. Being able to supply an alternative name for a key field is still functionality we have knocking about in some other components. There's no harm in leaving it if it would be useful to Charlie.
3. This might be cool, but maybe should be a separate PR? Again, we have many other analysis components that could be arranged like this, but aren't.

cmshobe:
For the names, how should I go about deciding what is better? I've skimmed the standard names list and not found anything perfectly applicable; does anyone have suggestions, or is there a specific format I should adhere to (e.g., object__quantity)?

gregtucker:
The naming pattern should be "object-attribute-value."

e.g.,

water__depth = 2.0
<object>__<attribute> = value

mcflugen:
Yes, of course. `distance_to_channel` should be updated. The CSDMS Standard Names for those two quantities would be something like,

*  `model_grid_cell_channel__mask`
*  `land_surface_point-to-channel__minimum_distance`

Your names don't have to be these but maybe it gives an idea?

SiccarPoint:
Given we already use channel__bed_shear_stress, channel__chi_index, channel__depth, channel__discharge etc `channel__mask` might work.

Maybe the other one something like `surface_to_channel__minimum_distance`??

mcflugen:
Some options regarding the `channel_network_name` keyword:
*  Remove the keyword.
*  Remove `channel_network` from `_input_var_names` (so it would only include `flow__receiver_node` and `flow__link_to_receiver_node'). The `__init__` method would then add the value of `channel_network_name` to `_input_var_names`.
* Keep `_input_var_names` as-is but have `__init__` replace `_input_var_names` with a tuple that includes `channel_network_name`'s value.

If the name can be changed when the component is instantiated, we can't accurately query the component for its input variables (`landlab uses` and `landlab used_by` wouldn't work, for example). I dont know, maybe we decide that that's not a problem.

The first thought is that every field variable should have one, and only one, name.

mcflugen:
@SiccarPoint: Based on names we already have, I think those suggestions work. Did you intentionally leave out the dashes from `surface_to_channel__minimum_distance`? Just wondering.

SiccarPoint:
We haven't talked about any use for dashes, so I just used underscores. No existing LL name has dashes, and they aren't discussed in any of our docs. Seems easier to leave them out.

SiccarPoint:
For the name overrides, the pattern we already use is like your third option. This makes sense I think; there is a standard name for whatever the field is, so that should be in the `_input_var_names`, but the user can dynamically override it later if they so wish. I guess this kind of functionality should probably be in the `Component` class though, rather than being implemented piecemeal in each component.

mcflugen:
@cmshobe: Sorry. @SiccarPoint and I have hijacked your pull request!

@SiccarPoint: we used to have something like this functionally in `Component` but removed it. Let's talk about this more elsewhere. If this is something we decide we want to allow, we should probably come up with a standard interface for it.

@cmshobe: We'll leave it up to you about the `channel_network_name` keyword for now. However, please stick to either option 1 or 3 from my list above. We may go in and change you component later, though. Or not.

cmshobe:
That sounds good; I'll make the name change based on suggestion 1 or 3! Thanks for the help!

cmshobe:
The solution that Greg and I ended up agreeing on is:

users supply an array, which can either point to an independent array or a grid field, as an optional argument. `__init__` then checks whether that array is the same length as the grid, and creates and populates a field with the standard name Eric suggested (`model_grid_cell_channel__mask`) if it doesn't already exist. This way we still allow the user freedom of naming their mask array, but we standardize the grid field. If I violated any standards in the few lines I added, @SiccarPoint or @mcflugen  please feel free to let me know.

cmshobe:
Also as a side note, if it's desirable for input fields to raise an error rather than create a blank field (as was written as a comment in this component), the component tutorial needs to reflect that. It currently teaches the "create blank field" approach.

SiccarPoint:
I like it, but I really don't like that field name (i.e., `model_grid_cell_channel__mask`). All other fields about channels simply start `channel__...`.

cmshobe:
`channel__mask` it shall be!

SiccarPoint:
Awesome.

mcflugen:
@SiccarPoint: the reason for the `model_grid_cell` prefix in the CSDMS Standard Names is to indicate that the quantity is not a physical quantity but, rather, a construct of the model itself. So something like `channel__depth` means the depth of an actual channel. However, a real-life channel doesn't have a "mask".

I'm not disagreeing with you or anything. Just throwing that out there as the reason behind its use in the CSDMS Standard Names. You are certainly correct that we have not used this pattern in landlab.

SiccarPoint:
I can see that, but still don't like the transfer into our scheme, as
1. `model_grid_cell_...` is long and ugly;
2. `..._cell_...` seems an unwise thing to include in a field name given its primary LL meaning; and
3. In this context `channel__mask` is unambiguous. Since, as you say, real channels don't have masks.

Also - just spitballing here - can we _really_ distinguish physical quantities from model constructs? After all, presence/absence of a channel within a given area is an observation that could be made about a real river, and conversely, I could argue that the calculated width of a stream using, say, Q~W^0.5 is just as much a model construct as anything else... ;-)

tl;dr This clearly makes total sense in a CSDMS BMI. But maybe not here?

mcflugen:
Oh, I'm totally fine with `channel__mask`. I just wanted to give the rationale in the CSDMS Standard Names. `model_grid_node_...` would be fine too. In fact, I'll probably use that in the landlab to CSDMS Standard Names mapper.

Another thing about the `model_grid_` prefix. Sorry. Usually - but not always - two components would not exchange values of things that start with `model_grid_` as these sorts of variables are tied to a model implementation. Since landlab components all use the same grid implementation, this makes sense but, in general, that's not the case.

The actual value of channel width calculated as `Q~W^0.5` may be a model construct, but channel width is still a real-life quantity that someone can measure. Different measurement techniques, I suppose, result in different widths but a channel does still have some width.

You are correct about the mask. Someone could go out in the field and say to themselves, "Self, is there a channel here?" and record the answer as a measurement. I had some trouble with this name as we don't have anything like it yet.

Anyway, after all that, I'm totally fine with the new names. Sorry to ramble.

SiccarPoint:
Ah OK, that makes a ton of sense for CSDMS if the rationale is to prevent inappropriate data exchange between models. Cool.

mcflugen:
@cmshobe: The way it is now, if a `channel__mask` array is passed as a keyword, but one already exists in the grid, the one in the grid is used and the passed one ignored. Did you intend for this to be the case? To me, it seems like if a user has passed an array of values his or her intent would be to override an existing array.

@gregtucker, @SiccarPoint: Do you have an opinion on this? If we use this pattern in other components, we should have a consistent way of handling this situation.

SiccarPoint:
I didn’t actually read this code, but if it’s the way you describe that’s not ideal. I agree if an array is getting passed, that ought to override. I believe it does in other cases where I’ve coded something similar. (e.g., some of the StreamPower cases sometimes do something like this).

On the other hand, if it’s well documented, I don’t see that it’s a disaster. We’re going to have a hell of a time enforcing this level of conformity amongst the components. (By which I mean, we’ve already got this level of unconformity! And we can’t change it easily now because - not to sound like a broken record this - we’ve gone past v1. I think I’m broadly OK with some level of user-facing _component_ inconsistency.)

cmshobe:
@mcflugen This should be fixed now, and should allow the user provided array to overwrite the existing channel__mask grid field. The only slight downside is that the existing grid field will be lost if it had info in it, but the user can see that fairly easily.

mcflugen:
@cmshobe: Thanks. I think it's better this way and conforms to the way other components deal with this (as far as I can tell - I only did a quick search).

I agree with you that there is the possibility of the user unknowingly overwriting existing data. However, if a user explicitly provides a channel mask they should (maybe?) realize the consequences. Perhaps a good compromise is to issue a warning if a grid is being overwritten?

PR #371, opened 2016-08-01
SiccarPoint/accel patch build

SiccarPoint:
This pull cythonises a key slow part of the grid build processes - the construction of the patch data structures. The cython implementation deliberately retains loops, as a fully vectorised version of this code uses unmanageably huge amounts of memory.

Note that this process remains slow in absolute terms (I think several minutes for grids in the 10^4-10^5 nodes range), but is faster than it was and won’t chew through all the system memory any more.


SiccarPoint:
There are fails in the 64 bit versions of the PC builds that I don't understand. They are related to type mismatches in Cython between `DTYPE_INT_t == np.int_t` and the ctype `long long`. This may need the attention of @mcflugen. The same code passes fine on all other builds.

SiccarPoint:
Thanks Eric... I see you beavering away at this!

mcflugen:
I am. I'm trying to fit it in in between doing other stuff so it's going slowly. It's an easy fix, I just have to remember what it is!

SiccarPoint:
Really enjoying that last test fail output: `ValueError: type mismatch: int32, int32`

mcflugen:
Oh yeah. That's my error message. It's just saying both input arrays were `int32`, which I think they were supposed to be.

I was thinking that one of the errors was in the call to `create_patches_at_element` but now I think it is all related to the error you mentioned before with `argwhere`. `argwhere` (along with some other numpy functions) returns an array of type `intp`, which is the data type used for indexing memory. On Windows x64 `int` is 32 bits while `intp` is 64 bits.

mcflugen:
I see `np.unique` is another. Anything that returns indices into an array.

SiccarPoint:
Yuck… Let me know if I can help fix this.

mcflugen:
By removing the declarations of the arrays you use to store the values returned by `np.unique` and `np.argwhere` in your Cython code, the errors go away. However, this may not be a real fix as your code might end up slowing down.

SiccarPoint:
Is there also a problem with argsort? I’m looking at the numpy code to see if I could sub in equivalent pure C to replace these funcs.

mcflugen:
I believe it has the same issue. You should definitely use C equivalents, if possible.

It looks like the AppVeyor tests are now passing.

SiccarPoint:
Quick scan here indicates this would not be trivial to do. As you’d expect, these indexing methods are all bundling a nontrivial amount of code, which I’m not really up for translating.

Is there any way to apply a test of the architecture? So at least type declarations could be made where they can work.

mcflugen:
OK. I think things are working again. I've added a `DTYPE_INTP_t` for the arrays returned by `argwhere` and `unique`.

SiccarPoint:
Fantastic. Sorry that took so long. At least we’ll know how to resolve this next time! I’m currently looking into some more cythonising in the flow router, so I think it might…

mcflugen:
No worries. It didn't take too long - not much of my time, anyway. Mostly just making small changes, triggering tests and waiting for them to complete.

This was actually a new problem for me so I'm glad we worked it out. I've never used numpy functions within cython code. Because of that, I'm a little concerned you're not getting the speedups you could be getting. What sort of speedups do you see?

I've also had to deal with this issue in my graph branch (I know, I know...). I would be interested in seeing if one way is faster than another.

SiccarPoint:
I didn’t actually do the speed tests. The main issue was that when I tried to vectorise, I manage to chew up all the RAM on my machine. So I reverted to looping, and since I knew this was a super slow section of code, I went straight into cython.

I suspect that *any* alternative in the graphs branch would be more efficient! I’ll be glad to eliminate this as a stopgap soon… ;-)

Incidentally, Kelly was having a problem where she was bumping up against the Python RecursionLimit running the fastscape flow router on really big grids. So I just cythonised the recursive bit of code that’s been in there since 2013, in the hope that bumping the recursive call down to C level would eliminate this issue. I don’t yet know if it did, but in the process I just shaved ~10% off our time to run tests! I think I accidentally fixed a major one of our speed bottlenecks…???

SiccarPoint:
Do you reckon this is good to go now?

SiccarPoint:
Bump. Would be good to have this in for v1.

PR #372, opened 2016-08-01
SiccarPoint/flow accum bw

SiccarPoint:
This pull adds cython for the recursive call to `_DrainageStack.add_to_stack()`, which is a vital class and method in the FLowRouter.

The intent was to suppress the triggering of Python’s recursion limit when the FlowRouter is called on giant grids (which I still need to get Kelly K to test). But regardless of that outcome, this pull appears to give the fastscape algorithms in Landlab a sizeable and very useful speed boost.

### DIFF
--- a/landlab/components/flow_accum/cfuncs.pyx
+++ b/landlab/components/flow_accum/cfuncs.pyx
@@ -1,0 +1,31 @@
+import numpy as np
+cimport numpy as np
+cimport cython
+
+
+DTYPE = np.int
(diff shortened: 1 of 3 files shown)


SiccarPoint:
If this is good, I'll also duplicate these changes directly onto the release branch.

gregtucker:
Looks good to me -- glad it fixes the recursion limit, and the speedup will be most welcome

SiccarPoint:
Well, let’s hope it fixes the limit! I’ll merge.

PR #376, opened 2016-08-02
SiccarPoint/rengers overland flow

SiccarPoint:
This pull implements two new components, largely written by Francis Rengers but fully componentized by me. The components are:

1. `SoilInfiltrationGreenAmpt`, a component to simulate the infiltration of a 1D column of water into a soil, following the Green and Ampt equation, and
2. `KinematicWave`, an explicit face centered solution to the depth varying Manning Equation, following Julien et al 1995, Mugler et al., 2011.

Both the components are described in Rengers et al. 2016.

I’m open to renaming the `KinematicWave`, as this solution is considerably more involved than the simplest possible schemes (c.f., Greg’s). Perhaps `RengersKinematicWave`? `JulienKinematicWave`? Thoughts welcome.

Note also that these components are in the `soil_moisture` and `overland_flow` folders, which already contain Sai and Jordan’s stuff. (This is invisible to users, of course.) However, the way that Sai has documented his component inside of `soil_moisture/__init__.py` makes this awkward. I think we should probably think more about how we want this set up.


SiccarPoint:
The rename I prevaricated about is now done; it's now `KinematicWaveRengers`.

gregtucker:
This looks good to me @SiccarPoint -- go ahead and merge

SiccarPoint:
I’ll hold off til @nicgaspar  gives go ahead for the other merge. But thanks for the review; I will!

mcflugen:
I think the new components should be moved into their own folders for this
pull request. Is there a reason they aren't? It seems as if they are
independent components.

SiccarPoint:
They fit where they are. They are directly relevant to soil_moisture and overland_flow.

If we want my components in separate folders to what we have, then we should rename the current folders and add new ones alongside (i.e., soil_moisture_nudurupati vs soil_moisture_greenampt, or whatever). But I’m not sure what we’d be adding by doing so.

nicgaspar:
@siccarpoint Sorry for messing you up yesterday.  I was confused and thought that the reviewer did the merge.  But I know there was another issue on top of that.  I started looking at your autodocs pull request yesterday.  It’s huge and I honestly don’t understand the details on the documentation stuff.  I also wondered when I saw “THIS DOES NOT WORK YET”.   But I liked the organization, or potential organization. Can you just give me some guidance on what to look for with this request?  It looks like it’s passing all tests, the code looks tidy, but any other direction?

SiccarPoint:
I’m not really sure. It’s basically just new functionality to let us introspect the code base, and then a little bit of helper functionality to help update the lists of methods at the top of the grid files that Sphinx scrapes.

The gist of it is that the LLCATS everywhere provide a way to scrape the various LL methods and funcs into categories. The key function doing this is `landlab.core.utils.get_categories_from_grid_methods`, which builds and returns dictionaries of method names subdivided by categories. It also returns a list of stuff that should have an LLCAT, but doesn’t (“FAILS”), for adding to the interface. At the moment this is basically only focused on the grid methods, but it would be trivial to expand it to the whole of LL later.

That’s the heart of it. However, there’s also included here some headway towards automating the building of some of our docs using this interface. (in the docs folder). However to automate fully, we’d have to restructure the way we scrape the docs onto readthedocs, which isn’t hard but might be time consuming. So this represents a halfway house, where functionality exists to make text blocks that look like what we have at the top of the grid files, but you have to cut and paste it into those files by hand still (which I did as part of this pull). Hence we have the build_rst_webpages file, which was an abortive effort to do this properly, and the build_func_lists file, which implements the halfway house solution.

I’ll take out the build_rst_webpages file entirely from this pull, and make it a new branch. That seems most sensible. Then everything else should be clean and internally consistent.

I should probably also add some web documentation on how this works. It would go in the dev documentation, but I don;t think is part of this pull.

SiccarPoint:
ETA: This is in the wrong place. I'm replicating my comment on the right pull request.

SiccarPoint:
I'm going ahead with a merge for this. We can shunt the files around the folders at our leisure later.

PR #377, opened 2016-08-04
SiccarPoint/autodocs

SiccarPoint:
This pull updates the sphinx formatting of the headers on the grid files, from which are auto generated the readthedocs.

It also adds a script to produce this text, and implements a new flagging scheme - “LLCATS” - for the building of this text.


SiccarPoint:
Note that this pull is likely to throw a heap of merge conflicts on the grid files. It's all trivial stuff, but could be a bit frustrating...

nicgaspar:
Dan - I’ll look soon.  But what do you mean by merge conflicts?  You mean when updating branches?  Or is this something I need to look for - I don’t think you mean the latter.

SiccarPoint:
I mean, any branches you have from before this gets merged will likely throw a bunch of conflicts when you update from master after it has been.

I know this will happen because I already have a whole bunch more conflicts with the master from a single push Eric did a few days ago…! I’m fixing them now.

SiccarPoint:
OK, this should now be good to go...

SiccarPoint:
(Copied across from wrong place)

I’m not really sure. It’s basically just new functionality to let us introspect the code base, and then a little bit of helper functionality to help update the lists of methods at the top of the grid files that Sphinx scrapes.

The gist of it is that the LLCATS everywhere provide a way to scrape the various LL methods and funcs into categories. The key function doing this is `landlab.core.utils.get_categories_from_grid_methods`, which builds and returns dictionaries of method names subdivided by categories. It also returns a list of stuff that should have an LLCAT, but doesn’t (“FAILS”), for adding to the interface. At the moment this is basically only focused on the grid methods, but it would be trivial to expand it to the whole of LL later.

That’s the heart of it. However, there’s also included here some headway towards automating the building of some of our docs using this interface. (in the docs folder). However to automate fully, we’d have to restructure the way we scrape the docs onto readthedocs, which isn’t hard but might be time consuming. So this represents a halfway house, where functionality exists to make text blocks that look like what we have at the top of the grid files, but you have to cut and paste it into those files by hand still (which I did as part of this pull). Hence we have the build_rst_webpages file, which was an abortive effort to do this properly, and the build_func_lists file, which implements the halfway house solution.

I’ll take out the build_rst_webpages file entirely from this pull, and make it a new branch. That seems most sensible. Then everything else should be clean and internally consistent.

I should probably also add some web documentation on how this works. It would go in the dev documentation, but I don;t think is part of this pull.

nicgaspar:
I say go for the merge.

PR #381, opened 2016-08-10
SiccarPoint/fix issue 380

SiccarPoint:
Resolves some of the issues surrounding poor performance of `imshow_grid` if called inline.


jadams15:
The tests errored, didn’t fail. And the latest run on Travis actually did pass. Merging.

PR #384, opened 2016-08-12
added return_outlet_id to set_watershed_BC

nicgaspar:


nicgaspar:
Literally all this does is add a flag to set_watershed_boundary_condition that now allows for the id of the outlet to be returned (as a numpy array of one value).

PR #386, opened 2016-08-16
Added Graph and DualGraph classes

mcflugen:
This pull request adds a new subpackage, `landlab.graph`, that contains data structures that describe the topology of various graph structures. This will unify all of our graph structures so that we can be sure, for example, that all graphs are consistent with one another and contain all the same `_at_` and `_of_` attributes. Graphs that have been added are,

*  Graph (general graph topology)
*  VoronoiGraph
*  RadialGraph
*  HexGraph
*  StructuredQuadGraph
*  RectilinearGraph
*  UniformRectilinearGraph (raster)

as well as all their dual graphs. There's lots in here but perhaps the most useful will be the new `plot_graph` function that nicely plots and labels all the graph elements so that users can see how links, nodes, faces, etc. connect with one another and what their ids are.

There are a bunch of notebooks that go with this. I'll upload those separately to the tutorials repository.

**What this pull request is not: a complete integration of these classes into the existing `ModelGrid`, `RasterModelGrid`, etc.** I've decided it will be much easier to do this in pieces rather than one big merge.

What is here is nearly, but not totally, complete. There are still a few lingering details but I would like to address those through additional pull requests - as they become necessary.

Because, for the most part, this pull request just adds a new self-contained sub-package, it can be safely merged without messing anything else up.


gregtucker:
@mcflugen : I'm going to go ahead and merge this. I've given a quick read and it looks great.

PR #387, opened 2016-08-16
SiccarPoint/OverlandFlow run one step()

SiccarPoint:
The pull implements true dynamic time stepping with an imposed dt in the OverlandFlow component. It also implements a standardised `run_one_step(dt)` run method, in addition to the method it already had.


jadams15:
@SiccarPoint all name changes for surface_water__discharge / surface_water__discharge are included in here. tests pass on my machine... let's see what happens!

PR #388, opened 2016-08-16
SiccarPoint/standard name updates

SiccarPoint:
Updates the standard names we discussed on the call.


nicgaspar:
I'm merging.  looks good.

PR #393, opened 2016-09-04
Remove active link fields

mcflugen:
This pull request completely removes field locations for `active_links`, `active_faces`, `core_nodes`, and `core_cells`.

@jadams15: The biggest change, which actually isn't all the big, is in the unit tests for your overland flow component. Maybe you could have a look at see if it's fine?


mcflugen:
@jadams15: I think this should be OK so I'm going to go ahead and merge. It's a small change.

PR #398, opened 2016-10-08
Updates to drainage density component

cmshobe:
I made the requested change that allows users to specify a channelization threshold rule of the form aA^(b)*cS^(d)>thresh, if they don't want to supply a channel mask. The component throws an error if neither approach is taken. It'd be great if @mcflugen and @gregtucker would take a look at this and then merge if all looks OK.

### DIFF
--- a/landlab/components/drainage_density/drainage_density.py
+++ b/landlab/components/drainage_density/drainage_density.py
@@ -4,7 +4,7 @@ Created on Mon Jul 11 10:00:25 2016
 
 @author: Charlie Shobe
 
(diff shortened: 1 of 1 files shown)


kbarnhart:
Hey @cmshobe a quick update. Discussed this with @mcflugen this morning and it looks like Apveyor had some issues while building the landlab module. He said he would re-run the Apveyor tests and then we can re-visit this pull request.

mcflugen:
All the tests are passing now.

PR #399, opened 2016-10-21
Three functions for cell subtriangles (slope, aspect, unit normal)

kbarnhart:
Created three new raster functions to calculate the unit normals, slope, and aspect of the eight cell sub-triangles. These are intended as building blocks to Dinfinity, Dtrig, and other multidirectional flow routing algorithms. 

Passed doctests on my laptop.


SiccarPoint:
OK, I didn't check the detail of the actual unit test calculations, but if you're happy this is working right it's ready to be merged.

Also, I note you've maintained the convention I adopted that for aspect - only, in the whole of Landlab! - that we go CW from N, not CCW from E. This was because I have never in my life seen it defined any other way than CW from N, but I agonised over it. Just be aware that it's a bit of a weird special case, sorry.

PR #400, opened 2016-10-24
Gt lake bug second try

gregtucker:
This fixes the lake-routing bug. In certain situations, a node associated with a lake would point back into the lake. The result was that two adjacent nodes would point to one another. This led the flow routing procedure to essentially ignore the entire area, leaving its nodes at unit drainage area. Fixing this required a fairly radical re-write of the routing portion of the lake code. The algorithm is not optimized for speed --- not even vectorized with numpy --- so it is likely to be rather slow, and probably doesn't scale well. But at least it seems to solve the problem.


SiccarPoint:
Is the performance worse than it was? Or just "still poor"?

Given the iterative coding style of these routines, this component looks like an obvious candidate for thorough Cythonizing.

gregtucker:
"Still poor". I tested it on a DEM with 807,000 nodes, and the whole procedure (routing, lake filling, plotting) runs in a few seconds. I agree it's ripe for Cy'ing. If someone could dream up an O(N) algorithm, that would be even better!

PR #403, opened 2016-11-11
precip gen: don't round; allow generated to sequence to end in a storm

gregtucker:
Two minor changes to the PrecipitationDistribution component:
1) storm and interstorm duration are no longer rounded to 2 decimals (which I think removes the need to test for zero durations)
2) the generator function now allows the possibility for the sequence to end in the middle of a storm (previously, if time ran out during a storm, it would still return a zero-duration interstorm period, which had the odd side effect of forcing a model to always have zero precipitation just before an output to file)


jadams15:
Merging!

PR #405, opened 2016-11-14
SiccarPoint/no get methods

SiccarPoint:
This pull removes vestigial grid methods that use the obsolete `grid.get_…` syntax. This should all be fine except a couple which need Sai to check they don’t break his (untested) code.


SiccarPoint:
@saisiddu : please check the change this makes to `plant_competition_ca.py`. It's only two lines, and only subbing in newly named properties for methods, but I don't think this stuff is tested so I can't be sure I did it right.

saisiddu:
@SiccarPoint: plant_competition_ca.py is working fine. Will start adding unit tests to the core methods.

SiccarPoint:
Bump. This pull MUST be incorporated into the new release of Landlab for paper revisions.

PR #406, opened 2016-11-15
Adding a field group "at_everywhere" to store single element arrays meant to represent values global to the grid. 

kbarnhart:
To discuss and potentially change from this pull request:

1) Name. I’ve used “at_everywhere” because to keep in line with the current code used to set/test the size of controlled field groups I needed to create a property that mirrored the existing properties grid.number_of …  grid.number_of_value_at_everywhere is imperfect, but it is the best thing that I could come up with. If anyone has a better idea for how to name this, please let me know. 

2) Value or array. At the moment,  I’ve mirrored the other controlled field groups (at_node, at_cell, etc) in that this field group must be made of arrays of length 1. One could argue that this field group should be required to be str, int, float, or bool, rather than an array of length one. I wasn’t sure which was best, so I’ve implemented the version that was closest to the current field groups, and listed this here for discussion.

### DIFF
--- a/landlab/field/field_mixin.py
+++ b/landlab/field/field_mixin.py
@@ -58,7 +58,13 @@ class ModelDataFieldsMixIn(ModelDataFields):
             group = kwds.pop('at', kwds.pop('centering', 'node'))
         else:
             group = args[0]
-
+        
+        if group=='grid':
+            raise ValueError("empty is not supported for at='grid', if you "
+                             "want to create a field at the grid, use\n"
+                             "grid.at_grid['value_name']=value\n"
+                             "instead.")        
+        
         n_elements = self.number_of_elements(group)
 
         if self[group].size is None:
@@ -107,7 +113,14 @@ class ModelDataFieldsMixIn(ModelDataFields):
             group = kwds.pop('at', kwds.pop('centering', 'node'))
         else:
             group = args[0]
-
+        
+        if group=='grid':
+            raise ValueError("ones is not supported for at='grid', if you "
+                             "want to create a field at the grid, use\n"
+                             "grid.at_grid['value_name']=value\n"
+                             "instead.\nAlternatively, if you want ones"
+                             "of the shape stored at_grid, use np.array(1).")
+                             
         n_elements = self.number_of_elements(group)
 
         if self[group].size is None:
@@ -151,7 +164,14 @@ class ModelDataFieldsMixIn(ModelDataFields):
             group = kwds.pop('at', kwds.pop('centering', 'node'))
         else:
             group = args[0]
-
+            
+        if group=='grid':
+            raise ValueError("zeros is not supported for at='grid', if you "
+                             "want to create a field at the grid, use\n"
+                             "grid.at_grid['value_name']=value\n"
+                             "instead.\nAlternatively, if you want zeros"
+                             "of the shape stored at_grid, use np.array(0).")
+                             
         n_elements = self.number_of_elements(group)
 
         if self[group].size is None:
--- a/landlab/field/grouped.py
+++ b/landlab/field/grouped.py
@@ -492,6 +492,8 @@ class ModelDataFields(object):
         Return a new array of the data field size, filled with zeros. Keyword
         arguments are the same as that for the equivalent numpy function.
 
+        This method is not valid for the group *grid*.        
+        
         See Also
         --------
         numpy.zeros : See for a description of optional keywords.
@@ -525,7 +527,9 @@ class ModelDataFields(object):
         entries, and add it to the field as *name*. The *units* keyword gives
         the units of the new fields as a string. Remaining keyword arguments
         are the same as that for the equivalent numpy function.
-
+        
+        This method is not valid for the group *grid*.
+        
         Construction::
 
             add_empty(group, name, units='-', noclobber=True)
@@ -564,8 +568,8 @@ class ModelDataFields(object):
             raise ValueError('number of arguments must be 1 or 2')
             
         if group=='grid':
-            raise ValueError("add_empty is not supported for at_grid values"
-                             "use \n grid.at_grid['value_name']=value \n"
+            raise ValueError("add_empty is not supported for at_grid values "
+                             "use\ngrid.at_grid['value_name']=value\n"
                              "instead")        
                              
         numpy_kwds = kwds.copy()
@@ -582,6 +586,8 @@ class ModelDataFields(object):
         add it to the field as *name*. The *units* keyword gives the units of
         the new fields as a string. Remaining keyword arguments are the same
         as that for the equivalent numpy function.
+        
+        This method is not valid for the group *grid*.
 
         Construction::
 
@@ -638,7 +644,7 @@ class ModelDataFields(object):
 
         if group=='grid':
             raise ValueError("add_ones is not supported for at_grid values"
-                             "use \n grid.at_grid['value_name']=value \n"
+                             " use\ngrid.at_grid['value_name']=value\n"
                              "instead")                
         
         numpy_kwds = kwds.copy()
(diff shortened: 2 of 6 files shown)


SiccarPoint:
'At_all' would be more concise, but I guess doesn't quite mean the same...

'At_grid' might be a promising option, as the value is attached to the grid as a whole as opposed to a single element type.

kbarnhart:
@mcflugen and I discussed `at_grid` yesterday and favored it at that point I ended up changing my mind to `at_everywhere` because the method by which the fields are constructed references a grid attribute with a name like `number_of ...` and I though `number_of_value_at_grid` seemed really really awful, but `number_of_value_at_everywhere` was slightly less awful. I think they are both awful and am not enamored with this method. 

I think I favor setting them as Python primitives rather than arrays of length 1 (see options and ideas listed in #407), but am not sure how to implement this within the current architecture.  In looking through `ModelDataFieldsMixIn`, `ModelDataFields`, and `ScalarDataFields` it appears to me that a field being an array is pretty entrenched.

kbarnhart:
Perhaps it makes more sense to not add these sort of global values as a field but as a new sort of dict-like part of the grid. 

My favored idea at the moment:

Create a class like `ScalarDataFields`, we'll call it `PrimitiveDataFields` that can creates a single value field that is either float, int, or bool. `PrimitiveDataFields` would live in the `field` folder. Beyond creating fields, `PrimitiveDataFields` would do things like make sure that the type of a global field value doesn't change.

Then use this class to create a set of fields called `grid.globals` that can be python primitives. 

We intentionally don't put an `at` at the beginning of the name to indicate that this is a different sort of grid field than those stored in the `at_..` dicts. 

What do you guys think of this? If you like it, I can work on implementing it.

mcflugen:
I like `at_grid` better than `at_everywhere`. It shorter and just as clear, I think.

I was thinking not of numpy arrays, but numpy scalars. (`numpy.array(2.)`). Maybe this would be a good compromise? As you say, numpy arrays are a big part of our current field implementation but we could also use primitive types for our scalar fields if that's what we want.

I just pushed a change that implements fields of numpy scalars. Please have a look. If you don't like it, not problem, just undo all my changes.

So now, you would add a scalar field with,
```python
>>> grid.at_grid['const'] = 2.
>>> grid.at_grid['const']
array(2.)
```

mcflugen:
I think I would be OK with `grid.globals`. However, to maintain consistency, I still like the `at_grid` (or `at_all` or whatever we decide) better.

It's my opinion that before we go creating new classes, we should try to work with what we have. It makes things easier to maintain and understand as well as being more consistent for users. If we just can't work with what we have, then we may need to take more drastic measures.

What do you think about using numpy scalars rather than arrays of length 1 or primitive data types?

kbarnhart:
@mcflugen sorry for delay. was eating lunch!  I like your modification. I think its the best implementation without changing too much in the underlying architecture (as I understand it). 

assuming that it passes the all the tests, I think it is good enough to merge. 

since we have each commits on this pull request, do we need to assign another person to merge the request?

mcflugen:
@kbarnhart all the tests should pass as I haven't added any new ones! Do you have time to add unit tests?

kbarnhart:
Yes, I can add unit tests. Quick question since so far I have only added doctests... 

Since this new functionality is part of the grid base.py, the tests should be in the grid test rather than the field tests, right?

However it doesn't look like there are any base grid tests, so I can put the test somewhere like inside of  `landlab\grid\tests\test_raster_grid\test_fields.py'

kbarnhart:
Yup, current version is `at_grid`.

I'm working on the unit test now

mcflugen:
@kbarnhart You should add the unit tests into the tests folder under the fields folder. Certainly, you should also add some doctests to the grids as well but more extensive unit tests should go in the `landlab/field/tests`. I just pushed a commit that adds a simple unit test, which you can use as a starting point.

Note that `grid.add_ones`, and the like do not work for `at=grid` at the moment. I don't know if that's a bug or a feature. As things stand you have to instead do `grid.at_grid['const'] = 1` to get what you want. Since the `at_grid` field is always - *not matter the grid* - a single value, it makes sense not to have `add_ones`, `add_empty`, etc. for it (I think). The counter argument is, of course, consistency with other fields.

kbarnhart:
@mcflugen thanks for the starting point! 
Will write more tests. 

I favor the argument that it is a feature. At the moment, there are two ways to make this sort of a field

`grid.at_grid['value']=2`
or
`grid.add_field('value', 2, at='grid')`

the latter method parallels the multi-element fields, and the former does not. the `add_ones`, and `add_zeros` are not really appropriate, since they are plural.

kbarnhart:
At the moment these two methods result in different outcomes will work on fixing. 

`grid = RasterModelGrid((4, 5))
grid.at_grid['value_1']=1
grid.add_field('value_2', 1, at='grid')
Out[4]: array([1])
grid.at_grid
Out[5]: {'value_1': array(1), 'value_2': array([1])}`

mcflugen:
Yes, that needs fixing. Something to keep in mind is that some fields will have only one element and should be an array of length 1 (as opposed to a scalar).

We may also want to consider deprecating `add_field` (if not for all elements, then at least for `at=grid`). `grid.at_grid['val'] = 1.` is, to me anyway, so much clearer.

kbarnhart:
OK. I'll add the depreciate warning to that part of `scalar_data_fields.py` for at=grid.

mcflugen:
Hold off on the deprecation warning for now. scalar_data_fields.py is
probably not the place to add it.

kbarnhart:
Thanks, was realizing that as I was starting to make it....

Best think I could come up for is to add something inside of `add_field` in `grouped.py` that says:

`if group=='grid':`
`            raise ValueError("add_field is not supported for at_grid values"`
`                            "use \n grid.at_grid['value_name']=value \n"`
`                             "instead")        `

I'll push it so you can see what I was thinking. Feel free to remove it (or ask me to remove it).

kbarnhart:
@mcflugen The changes you made are much better than my approach! 

An outstanding question: I set up ValueErrors for `add_ones`, `add_zeros`, and `add_empty` in addition to `add_field`. Should we remove them and make these methods return what a user might suspect (e.g. `array(1)` for `add_ones('field_name', at='grid')`

I don't think we should as I don't think these methods are really applicable to fields located `at_grid`. What do you guys think?

Other than this, I suspect we are pretty close to merging. If there are more things that you think need to be done, @mcflugen, let me know and I can stop by tomorrow. And the next time I'm at SEEC I will likely ask for a basic primer on how comprehensive unit tests should be.

SiccarPoint:
@kbarnhart , think you have it right for `add_ones`, `add_zeros`, `add_empty` with the value error. If we can get `add_field` behaving itself if passed `at=grid` it might be nice to have, but whatever.

Toying with adding a sentence on this to the revised manuscript, if the timeline is looking good for ~end of the week/early next week. It's cool functionality, and we've wanted it FOREVER. What do people think?

kbarnhart:
@siccarpoint, right now `add_field` does behave itself when given `at='grid'`. @mcflugen fixed that yesterday!

kbarnhart:
I'm now working on some more unit tests, but unless someone wants to look at them, I won't push them until I've *finished* the first draft, so we don't overload all of the CI infrastructure.

kbarnhart:
I'm going to also add ValueErrors for `ones`, `zeros`, and `empty` because these reference the dictionary `_ARRAY_LENGTH_ATTRIBUTES` in `base.py`. If we want these to work with `at_grid`, we'd need to make a method for a`at_grid` like `number_of_nodes`. I tried this approach yesterday and then we removed it.

I support not supporting these functions as it is (a) cluncky and (b) not really necessary since we would want a call like `grid.ones('grid')` to yield something like `array(1)`.

kbarnhart:
I've now added all the unit tests that make sense to me. @mcflugen are there others that you think we should add?

If no more tests are necessary, and if everything integrates successfully, perhaps @SiccarPoint or @gregtucker could merge this request for us?

PR #408, opened 2016-11-18
edit depth-dep prod and wxing

gregtucker:
Some syntactical and PEP8 edits to RCG's Depth-Dependent Diffusion and Expo Weathering components.

@kbarnhart when you review this could you get Rachel to look over your shoulder? Want to make sure she's on board. Edits are mostly to do with var naming. Parameter names should mirror those used in input files, and longer descriptive names are generally better than names like "k". I changed "weathering" to "soil production" because geochem folks have other meanings for "weathering" (of course soil scientists might not like "soil production" either, but "mobile regolith production" seems too long)

### DIFF
--- a/landlab/components/__init__.py
+++ b/landlab/components/__init__.py
@@ -21,7 +21,7 @@ from .plant_competition_ca import VegCA
 from .gflex import gFlex
 from .drainage_density import DrainageDensity
 from .fire_generator import FireGenerator
-from .weathering import ExponentialWeathering
+from .weathering import ExponentialWeatherer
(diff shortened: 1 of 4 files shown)


kbarnhart:
Yup, I'll go over it with Rachel later today.

kbarnhart:
@gregtucker , went over this with Rachel. She thought all of the edits were fine. Looks like one Travis build failed because of a time out during the matplotlib import.  Is it worth having this build re-run? I can ask @mcflugen to re-start it.

gregtucker:
Since it was just a time-out and the other tests passed, I think it's safe to merge.

PR #411, opened 2016-11-18
Modifications to @use_field_name_or_array and calc_flux_div_at_node

kbarnhart:


kbarnhart:
I've now made the changes we discussed on Monday and I summarized in Issue #411.

If someone wants to take a look and potentially commit that would be great!

PR #420, opened 2016-12-15
SiccarPoint/stabilize sedfluxdep

SiccarPoint:
This pull finishes the job (mostly!) of stabilising the SedFluxDepEroder.

Major issues were resolved allowing it to better stabilise its own internal time steps, and also to allow it to take a sediment supply imposed externally to the component. The component now works with a field called `channel_sediment__depth`, which stores the sediment load at each node in the channel network a thickness.


SiccarPoint:
Closing while final updates performed. Will re-open soon.

Issue #421, opened 2016-12-23
Should D8 or D4 director/accumulator be implemented for irregular grids?


kbarnhart:
In a prior conference call we discussed which between D8 or D4 flow director/flow accumulator makes sense for Voroni grids. The general consensus was that it should be D4---though in the current flow_router it is D8. 

The code from flow_routing.py is at line 352 and is pasted below. 
 
```
# Calculate flow directions
        if self.method == 'D4':
            num_d4_active = self._grid.number_of_active_links  # only d4
            receiver, steepest_slope, sink, recvr_link = \
                flow_direction_DN.flow_directions(elevs, self._active_links,
                                         self._activelink_tail[:num_d4_active],
                                         self._activelink_head[:num_d4_active],
                                         link_slope,
                                         grid=self._grid,
                                         baselevel_nodes=baselevel_nodes)
        else:  # Voronoi or D8
            receiver, steepest_slope, sink, recvr_link = \
                flow_direction_DN.flow_directions(elevs, self._active_links,
                                     self._activelink_tail,
                                     self._activelink_head, link_slope,
                                     grid=self._grid,
                                     baselevel_nodes=baselevel_nodes)
```

The way this is currently coded, I think it does make sense to keep Voroni with D8 instead of D4. What do you guys think?

mcflugen:
I wasn't on that call so I'm not sure what you all discussed, and so I'm probably missing something. With that said, I don't think either of those methods makes sense for anything other than a rectilinear grid. I would say both D4 and D8 imply rectilinear grids.

Now, if by "D4" you mean all flow goes to only one other cell, then I guess that could work. In that case, it's just that "D4" is a misnomer since cells don't have 4 sides.

In the case of D8, I don't see how it would work with a Voronoi grid. Apart from not having 8 neighbors, flow across a "diagonal" wouldn't make sense. Or maybe you just mean flow goes to multiple cells, but not across diagonals?

kbarnhart:
@mcflugen you make a good point. D8 and D4 do imply a raster grid and so it may make more sense to be explicit about regular vs irregular grids in how the new director/accumulator components are implemented. Under this idea we might imagine the following director/accumulator options:

**Single flow direction:**

- D4 (Raster) 
- D8 (Raster)
- SteepestDecent (Voroni)

**Multiple Flow Directions:**

- D-infinity (Raster) Flow partitioned to two nodes using the 8 sub-triangles, flow along diagonals permitted.
[...]

PR #422, opened 2016-12-29
Commit for discussion: initial draft of new flow directors and accumulators

kbarnhart:
New components for review listed below(_italic indicates planned but not yet implemented)_. The goal of this pull request is to get feedback regarding the structure on the first drafts these new components which split apart the FlowRouter component into FlowDirectors and FlowAccumulators. Once I get this feedback, I will finish the planned but not implemented components so all new components can join the master branch at the same time. 

New components presently documented and tested in the docstrings ONLY. I will make **tests** folder and associated files once the structure of the components is quasi-finalized. 

One added bit of functionality is that FlowAccumulators can now call a provided DepressionFinder instance from within the `run_one_step()` call. At the moment this does not provide any computational gains - however when combined with planned improvements to the DepressionFinder, this should result in computational gains.  

**Directors**

- **Base classes**

  - FlowDirector (base class)
  - FlowDirectorToOne (intermediate base class)
  - _FlowDirectorToMany_ (intermediate base class)

- **Flow-To-One**

  - FlowDirectorD4 (raster)
  - FlowDirectorD8 (raster)
  - FlowDirectorSteepestDescent (voroni)
  
- **Flow-To-Multiple**

   - FlowDirectorDinfinity (raster)
   - _FlowDirectorDtrig_ (raster)
   - _Various MFD flow directors_ (see #421 for notes)
 

**Accumulators**

- **Base Class**
  - FlowAccumulator (base class)

- **Flow-To-One**

  - FlowAccumulatorD4 (raster)
  - FlowAccumulatorD8 (raster)
  - FlowAccumulatorSteepestDescent (voroni)

- **Flow-To-Multiple**

  - _FlowAccumulatorDinfinity_(raster)
  - _FlowAccumulatorDtrig_
  - _Various MFD flow accumulators_


kbarnhart:
Looks like I'm failing via Travis and AppVeyor for Python 2.7 (I'm running python 3.5). I'll work on fixing those issues.

gregtucker:
Looks like py 2 and 3 define dict.keys() differently

kbarnhart:
Yes. I **think** I fixed it...

SiccarPoint:
I could certainly live with what Katy posted on Slack as the standard way of doing this:

```
from landlab.components import FlowDirectorFavoriteMethod
from landlab.components import FlowAccumulator

rmg = RasterModelGrid('whatever')
rfd = FlowDirectorFavoriteMethod(rmg)
rfd.run_one_step()
rfa = FlowAccumulator(rmg)
rfa.run_one_step()
```

We would then (in part for back compatibility, and in part because I think it would also be pretty useful) also be maintaining:

```
class FlowRouter(Component):
    def __init__(self, mg, flow_direction={'D4', D8', 'DN', 'SteepestDescent',...}, **kwds)
        '''
        Where flow_direction (or equivalent name) sets which FlowDirection
        component gets imported, then FlowAccumulator gets called after.
        D8 I guess would be the default, to match current behaviour.
        Note that the various flow_direction options wouldn't be mutually
        exclusive, so you could equally specify 'DN', 'D8', or 'SteepestDescent'
        if you had a Voronoi. This would be important for back-compatibility.
        '''
        self._accumulator = FlowAccumulator(mg)
        if flow_direction == 'DN' and type(mg) is Raster:
            self._director = FlowDirectorD4(mg)
        elif ...  # and so on

    def run_one_step(self):
        self._accumulator.run_one_step()
        self._director.run_one_step()
```

SiccarPoint:
I'd be happy to modify FlowRouter myself, once we've settled on a style for doing it "the proper way".

kbarnhart:
Had a good conversation just now with @gregtucker and we (perhaps) have come up with a good solution. Let me know what you think @SiccarPoint and @mcflugen. Also @mcflugen, I'd love to discuss this with you in person sometime soonish. I'll be down at SEEC starting around 12 noon. I suspect delta force is happening, so if meeting after then would work for you, let me know. 

the plan:
1) Each director method gets a unique component. This way if someone only wants direction calculated, it is possible. They all have names that start with `FlowDirector` and either inherit from either `FlowDirectorToOne` or `FlowDirectorToMany`. 

And if someone wants a new fun director method, they can make it!

We discuss more about how exactly to handle steepest descent for irregular grids.

2) There is ONE flow accumulator, called FlowAccumulator.
under the hood it figures out if it needs to to be either *AccumulateFromOne* or *AccumulateFromMany* (I don't think these are different classes at the moment, I might realized it is more effective to make the different classes while coding up this approach, however). 

The flow accumulator is decorated with a new decorator called
`@use_component_name_or_class` so that when the accumulator is instantiated you can pass either an uninitialized flow director component or the method name that lives after 'FlowDirector' in the name of the component. Thus the following two options are equivalent. 

```from landlab import RasterModelGrid
from landlab.components import FlowDirectorFavoriteMethod
from landlab.components import FlowAccumulator

rmg = RasterModelGrid('whatever')
rfd = FlowDirectorFavoriteMethod(rmg)
rfa = FlowAccumulator(rmg, director=rfd)
rfa.run_one_step()
```

AND

```from landlab import RasterModelGrid
from landlab.components import FlowAccumulator

rmg = RasterModelGrid('whatever')
rfa = FlowAccumulator(rmg, director='FavoriteMethod')
rfa.run_one_step()
```

If one wants to include a lake mapper in the accumulator, the following two options are equivalent:

```from landlab import RasterModelGrid
from landlab.components import FlowDirectorFavoriteMethod
from landlab.components import DirectionFinderAndRouter
from landlab.components import FlowAccumulator

rmg = RasterModelGrid('whatever')
rfd = FlowDirectorFavoriteMethod(rmg)
rfdf = DirectionFinderAndRouter(rmg)
rfa = FlowAccumulator(rmg, director = rfd, flow_director = rfdf)
rfa.run_one_step()
```

AND

```from landlab import RasterModelGrid
from landlab.components import FlowAccumulator

rmg = RasterModelGrid('whatever')
rfa = FlowAccumulator(rmg, director='FavoriteMethod', flow_director = 'DirectionFinderAndRouter')
rfa.run_one_step()
```

kbarnhart:
regarding backwards compatibility, @SiccarPoint what you outlined is exactly what we were thinking of doing once everyone was happy with the new Accumulator/Director components.

SiccarPoint:
I am very into this as a proposed approach. Endorse. Minor quibble - the naming and especially distinction between the `director` and `flow_director` arguments isn't super clear. The name of the argument you pass the sink router to may need a rethink.

kbarnhart:
Having conversation with @mcflugen at the moment, things we've come to agree upon:

Flow Director should have two major methods:
```python
fd.run_one_step()
```
and 

```python
r = fd.direct_flow() # for route to one methods
r,p = fd.direct_flow() # for route to multiple methods
```

So that users can get the receiver array directly (without accessing the grid). Under our planned way of coding this the grid fields would still get created, however. 

We also want to make it possible for the flow accumulator to be initialized using either a string, a class, or an instance of a class to identify the director and the depression_finder. For example the following three would be equivalent:

```python
from landlab.components import FlowDirectorFavoriteMethod
from landlab.components import FlowAccumulator

rmg = RasterModelGrid('whatever')
fd = FlowDirectorFavoriteMethod(mg)
rfa = FlowAccumulator(rmg, director=fd)
rfa.run_one_step()
```
AND

```python
from landlab.components import FlowDirectorFavoriteMethod
from landlab.components import FlowAccumulator

rmg = RasterModelGrid('whatever')
rfa = FlowAccumulator(rmg, director=FlowDirectorFavoriteMethod)
rfa.run_one_step()
```

AND

```python
from landlab.components import FlowAccumulator

rmg = RasterModelGrid('whatever')
rfa = FlowAccumulator(rmg, director='FavoriteMethod')
rfa.run_one_step()
```

We also want to be able to accumulate without recalculating directions. 

e.g. 

```python
from landlab.components import FlowAccumulator

rmg = RasterModelGrid('whatever')
rfa = FlowAccumulator(rmg, director='FavoriteMethod', calculate_directions=False)
rfa.run_one_step()
```

except we are trying to come up with a way to do this that doesn;t use keyword flags...


FINALLY. We discussed if there should be methods that would permit something like 

```python
r = direct_flow_d4(mg, z)
```

in which no grid fields are created and no class needs to be instantiated.

We realized that this ultimately is an issue regarding wanting to use Landlab for _analysis_ without _modeling_. I argued that we probably want to make this accessible, but that the way we do this should be designed based not only on the flow directors. Perhaps a new issue will be born shortly...

kbarnhart:
Also, I'll work on implementing some of these changes for your review. Likely early next week as I have a proposal due on Tuesday.

SiccarPoint:
Re accumulating without redirecting - this is of course a natural outcome of providing a wrapping method alongside an unwrapped part of that (i.e., your director and accumulator methods). It's not trivial to find a way of doing this robustly; I've tried to think this through before in other contexts (e.g., gradients) and it's very hard to find an efficient way of testing if a variable has already been updated this loop.

Hopefully though, it will never be an issue, as the user I think under your current proposals will never be able to do:

```
>>> direct_flow()
>>> accumulate_flow_only()
```

(though I still think this is more in our current style than the clever structures you are now proposing). I also might be misunderstanding what you mean by "accumulating without redirecting". If you simply mean enabling the style of the little code snippet above, I'd recommend simply allowing `None` to be passed to the `director` argument. For the record, I quite like keyword flag control structures, and don't feel like you need to work to avoid them. They're very explicit, and don't require the user (always remember, often an undergrad level geoscientist) to have an in-depth knowledge of Python to understand what's going on.

SiccarPoint:
Re analysis without modelling, as things stand, this is not what Landlab is - and also many platforms already exist to do analysis alone, including open source and surface process focussed (e.g., LSDTopoToolbox out of Edinburgh). Right now, for me this is not worth a large time investment. It certainly would be "nice to have", but doing this will require eventual modification of every component. That's not going to be easy to sort out... And it's not like we're particularly "done" with the actual modelling side if things.

kbarnhart:
Just about to push a new revised commit for comment. 

A comment in response to one of Dan's questions: **Why is b required? Should be optional surely? **

I assume this is in flow_accum_to_n.py which has the following method:
```python
a, q, s = flow_accumulation_to_n(r, p, b)
```
You are right, b is implied by the places where the receiver id is the same as the node id. I have it written the way it is now because this is how the equivalent function in `flow_accum_bw.py` is written. 

Would be easy to fix this and make b implicit by where node_id == receiver. What do you guys think? Should I do this?

Another question: I have a draft tutorial (Jupyter notebook form) for using the flow accumulator and flow director classes in the works. 

How/where do I add this to the landlab folders?

kbarnhart:
Just pushed a commit so I can use Landscape on the most recent version of the branch.

SiccarPoint:
Re b, I think definitely do it (or at least, set it so that default behaviour is to use it where possible, but with an optional override). Stuff like this is nice for the users. And in this immediate instance, for me ;-)

jennyknuth:
Hi Katy, tutorials live in the landlab/tutorials repo. You can see the standard readme and landlab header for your notebook in other tutorials there. 

Jenny

kbarnhart:
Thanks Jenny and Dan! 

b is gone now!

and could someone add me as a contributor to the landlab/tutorials repo?

kbarnhart:
A new update (that looks like it **almost** works...

FlowAccumulator, and four different FlowDirectors. See also three new tutorials on how to use them. 

I've also gutted FlowRouter, and replaced it with code that points into FlowAccumulator.

kbarnhart:
The windows bug has been fixed!

I've made a few additional changes (mostly pep8 on the newer files and few variable name changes). 

Would be great if someone could take a look at this and potentially merge.

gregtucker:
The two Travis fails look like website errors on their end, not actual Landlab fails.

This PR looks good to me from a quick scan.

kbarnhart:
@gregtucker was going to merge today, but there was a conflict with a recent update to the stream power smooth threshold eroder. I've fixed that. 

If tests pass @SiccarPoint is it ok if we merge this?

PR #424, opened 2017-01-05
enable spatially var threshold in smooth-thresh eroder

gregtucker:
This allows the smooth-threshold stream power component to use a field or array of threshold values.

### DIFF
--- a/landlab/components/stream_power/stream_power_smooth_threshold.py
+++ b/landlab/components/stream_power/stream_power_smooth_threshold.py
@@ -105,7 +105,7 @@ class StreamPowerSmoothThresholdEroder(FastscapeEroder):
     n_sp : float, optional, ~ 0.5<n_sp<4.
         n in the stream power equation (power on slope). NOTE: NOT PRESENTLY
         HONORED BY StreamPowerSmoothThresholdEroder (TODO)
-    threshold_sp : float (TODO: array, or field name)
+    threshold_sp : float, array, or field name
         The threshold stream power.
     rainfall_intensity : float; optional
         NOT PRESENTLY HONORED (TODO)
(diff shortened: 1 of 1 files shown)


kbarnhart:
Three of the Travis builds all have the same error with netcdf (which I don't think should be related). is that something to worry about @mcflugen and @gregtucker?

Other than that it looks good to me.

gregtucker:
I'm not sure ... @mcflugen ?

mcflugen:
Yes, this is something to worry about (but not too much). It appears netcdf4 v1.2.6 that's distributed through the default Anaconda channel doesn't work on Python 3.5 on Mac. I see two possible solutions:
1. Require netcdf4 v1.2.4 (or maybe < 1.2.6)
2. Install dependencies from the conda-forge channel

Neither option is great. I usually install from conda-forge anyway but I'm sure others don't. Attaching ourselves to a specific version of a dependency isn't great either as it could cause dependency conflicts for users.

kbarnhart:
@mcflugen does this error make sense to you in the context of this specific pull request?

mcflugen:
Yes. Kind of. Anaconda upgraded netcdf4 to 1.2.6 since our last build of master. We should probably create a fix on master (and on our release branch) that deals with this. If we were to rerun our builds on master or of our release, I'm sure they would fail in the same way.

kbarnhart:
OK. So it sounds like if @gregtucker really needs this branch in master soon, we can pull it in since the error isn't coming from a change in this branch. But if he doesn't need it asap I'll wait until there is a fix for this issue. 

Good plan?

mcflugen:
That sounds right. I'll try to come up with a fix for the netcdf issue
today.

Issue #425, opened 2017-01-11
set_watershed_boundary_condition breaks if the grid has two nodes with the minimum value 


kbarnhart:
the raster function `set_watershed_boundary_condition()` doesn't handle the case in which the grid has two nodes with the minimum value. It throws the following error if this is the case.

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-7-9f7194105aab> in <module>()
      7 # set_watershed_boundary_condition() and it would find it.
      8 
----> 9 outlet=grid.set_watershed_boundary_condition(z, nodata_value=-9999.,)
     10 grid.set_closed_boundaries_at_grid_edges(True, True, True, True)

/Users/katybarnhart/git/landlab/landlab/grid/raster.py in set_watershed_boundary_condition(self, node_data, nodata_value, return_outlet_id, remove_disconnected, method)
   5272             i = 0
   5273             while (i < len(min_locs) and local_not_found):
-> 5274                 if self.has_boundary_neighbor(min_locs[i]):
   5275                     local_not_found = False
   5276                     # outlet_loc contains the index of the outlet location

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()```


This came up in a test dem that @saisiddu shared with me.

jadams15:
So in this case there are two core nodes adjacent to boundary nodes, with the same elevation values? Is this one delineated watershed?

saisiddu:
Yes, this is a delineated watershed.
________________________________

jadams15:
thanks, just making sure I understand the issue! I can see this being problematic. Do you know if the two "outlet"/min value elevations are next to each other?

saisiddu:
Yes, they are connected. The watershed itself was a little problematic at first. But, after using the DepressionFinder correctly, the router behaved well.
________________________________

kbarnhart:
In Sai's watershed they are adjacent (node ids: 1389 and 1390).

kbarnhart:
See PR #426 
That fixes this issue

There are, however some outstanding issues in _how to handle_ this sort of issue. 

I'll tag you in that PR @jadams15 and suggest we discuss how to handle it in that PR instead of here.

kbarnhart:
[hugows10mpitf.txt](https://github.com/landlab/landlab/files/700113/hugows10mpitf.txt)
So we have @saisiddu 's raster in this issue thread I'm attaching it now.

kbarnhart:
[...]

PR #426, opened 2017-01-11
Fixing the set_watershed_boundary() issue identified in #425 

kbarnhart:
Minor changes to fix `set_watershed_boundary()` (see issue #425).

An outstanding question:

I have not grappled with _which_ of multiple minimums may be selected.

If a grid has two minimums, and only one is adjacent to a boundary, this one will be chosen.

However, if both minimums are on a boundary, then the FIRST one returned by np.where() is the one that `set_watershed_boundary()` will set as the outlet_node and with status `FIXED_VALUE_BOUNDARY`

I suggest we discuss if this should be tested for and how it should be handled.

### DIFF
--- a/landlab/grid/raster.py
+++ b/landlab/grid/raster.py
@@ -5254,7 +5254,7 @@ class RasterModelGrid(ModelGrid, RasterModelGridPlotter):
         min_val = np.min(node_data[locs])
 
         # now find where minimum values are
-        min_locs = list(np.where(node_data == min_val)[0])
+        min_locs = np.where(node_data == min_val)[0]
 
         
         # check all the locations with the minimum value to see if one
@@ -5283,7 +5283,13 @@ class RasterModelGrid(ModelGrid, RasterModelGridPlotter):
             if any(next_to_boundary):
                 local_not_found = False
                 if sum(next_to_boundary)>1:
-                    raise ValueError('Grid has two potential outlet nodes')
+                    potential_locs = min_locs[np.where(np.asarray(next_to_boundary))[0]]
+                    raise ValueError(('Grid has two potential outlet nodes.'
+                                      'They have the following node IDs: \n'+str(potential_locs)+
+                                     '\nUse the method set_watershed_boundary_condition_outlet_id '
+                                     'to explicitly select one of these '
+                                     'IDs as the outlet node.'
+                                     ))
                 else:
                     outlet_loc = min_locs[np.where(next_to_boundary)[0][0]]
 
@@ -5298,7 +5304,7 @@ class RasterModelGrid(ModelGrid, RasterModelGridPlotter):
                                 (node_data != nodata_value))
                 # now find new minimum of these values
                 min_val = np.min(node_data[locs])
-                min_locs = list(np.where(node_data == min_val)[0])
+                min_locs = np.where(node_data == min_val)[0]
             else:
                 # if locally found, it is also globally found
                 # so done with outer while
--- a/landlab/grid/raster.py
+++ b/landlab/grid/raster.py
@@ -5256,6 +5256,7 @@ class RasterModelGrid(ModelGrid, RasterModelGridPlotter):
         # now find where minimum values are
         min_locs = list(np.where(node_data == min_val)[0])
 
+        
         # check all the locations with the minimum value to see if one
         # is adjacent to a boundary location.  If so, that will be the
         # watershed outlet.  If none of these points qualify, then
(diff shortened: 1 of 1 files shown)


jadams15:
I think arbitrarily selecting the first ID in np.where() *could* be an issue when using D4 instead of D8 flow routing.. I'm making a sketch in illustrator right now to demonstrate... give me a few minutes and I'll explain!

kbarnhart:
I agree its an issue and will influence flow direction and routing. (I can envision your sketch @jadams15 and I'm already convinced its awesome!)

Since, `set_watershed_boundary_condition()` is a boundary condition method. I think it shouldn't presume anything about what the user will want to do with it next (e.g. specify desire for D8 or D4 directions). 

I think my preferred solution is that ONLY one node can be lowest AND next to a boundary (Raise an ValueError otherwise).

Everything else I can come up with has lots of  nonuniqueness. For example selecting among nodes that have the minimum elevation value based on having the most closed node neighbors.

jadams15:
While I completely agree that there can be only one node for the outlet, this does need to include some thought towards flow routing, as the user can select D4 or D8 when calling set_watershed_boundary_conditions, using the keyword "method". 

I agree that figuring out which has most NODATA neighbors won't work...

nicgaspar:
I don’t see anything other than an arbitrary solution for this issue. That isn’t awesome. But a watershed outlet is defined as the lowest point on the boundary. That’s just the definition. Real life doesn’t hold with definitions. Maybe in this case the user has to be a bit more knowledgeable? Maybe throwing an error is the right thing to do?

jadams15:
also my sketch has actually had the opposite outcome I predicted. if anything, it merely proved to me that it shouldn't matter *too* much! So for now, seems a reasonable patch. I think Nicole is in a meeting at the moment, so we should definitely wait until she's had a time to chime in. Good find @kbarnhart and @saisiddu

nicgaspar:
I agree Jordan. I can’t envision how an arbitrary decision would matter for the case of two nodes next to each other. The real issue is when there are two nodes of the same value that are not next to each other. That is why I think maybe we should just throw an error and force the user to think about it.

kbarnhart:
Just made a commit that throws such an error.

SiccarPoint:
Couple of potentially simple-minded questions on this:

1. Why is having multiple outlets forbidden? We do this all the time in fully synthetic landscapes, e.g., opening a whole side of a grid.

2. Assuming the answer to (1) is obvious, if we do raise an error, we should probably add more functionality to help deal with this problem, as manual editing of nodes in a raster is - as we all know - a right pain. I’m thinking, e.g., an additional method that the error would refer to that returns the IDs of all the potential outlets. Or this info gets printed as part of the error. Then (e.g.) there could be an input argument on `set_watershed_boundary` that lets you specify which outlet option should be used if multiple are possible.

kbarnhart:
Looks like the build errors are on the netcdf that @mcflugen @gregtucker and I were discussing with regard to PR #424. I know that @mcflugen is currently working on this. 

As to Dan's comments:

1. I think that adding the functionality of `running set_watershed_boundary` that would work with multiple watersheds is a potentially very interesting one. In theory I like it.

Regarding the present function: I suspect that the reason it is not implemented in the current function is that the current function finds the outlet by finding the lowest elevation node that is adjacent to a boundary node. Finding multiple outlets by this method wouldn't work because multiple outlets shouldn't need to be a the same elevations. The method was designed for one watershed (and its documentation says just as much).

Setting all edge nodes to outlets also won't work since this method should work for irregular watersheds and for watersheds with outlets that do not lie on a convex hull that encircles the watershed.  

I _think_ that re-writing the `set_watershed_boundary_condition` to be able to handle multiple watersheds just requires running flow directions on the watershed. The **FlowDirectors** identify **sinks**, or nodes that drain into themselves. An outlet would then be identified as  any sink node that is adjacent to a closed node. 

A flow direction method would need to be specified. I think this is fine but it brings up  something @jadams15 mentioned yesterday. I should change the name of the *method* keyword that is presently in `set_watershed_boundary_condition` because it is potentially confusing. It should be something like *adjacency_method* because it is related to a method called `set_open_nodes_disconnected_from_watershed_to_closed` that I wrote to deal with stupid open nodes that sometimes exist if an arc clip was designed for a raster that is a different resolution that the raster being clipped and thus leaves some straggler nodes.

Does anyone see a problem with permitting this functionality in this way? I'm trying to think about how it would handle the case that @saisiddu 's watershed has, in which two nodes are next to one another. It would set both nodes to outlet nodes because they are both the same elevation AND adjacent to a boundary node. In D8 and D4, flow would be routed into one or the other based on the link slopes rather than by node elevations. I think this is how we would hope this to behave, but if someone thinks otherwise, let me know. 

2. Re error messages: If we don't do what is outlined in (1) I agree. A much better error message would be a great idea.

SiccarPoint:
I don’t think I’ve been clear with (1). I meant: it should be possible to have multiple (contiguous) nodes be valid outlets _for a single catchment_. Think of e.g. a river leaving a catchment as a wide braid plane.

kbarnhart:
ahhh, yes, I see. 
Though, from a topologic network sense (in a D8 or D4 routing framework) wouldn't this functionally be two watersheds?

SiccarPoint:
Not always, right? If you have divergent flow…

kbarnhart:
agreed, but that isn't permitted under D8 and D4.

kbarnhart:
Which I think is an argument to use a specific **FlowDirector** to find the outlets, yes?

SiccarPoint:
Yeah, seems like the current method is very much flow director specific. We may wish to generalise it as part of the work you’re doing…?

Perhaps the answer then is to apply your current, arbitrary fix (outlet is the first of the lowest nodes) just so this doesn’t crash sometimes, then work on a separate, alternative, more full functioned method as part of your ongoing flow router improvements??

kbarnhart:
Sounds good to me. I'll do that later today!

nicgaspar:
Thanks for adding the conversation Dan. I didn’t appreciate what you are getting at. I think the issue is that set_watershed_boundary() does what it is designed to do but people want it to do other things as well. I’m fine with that, but I’m not sure the best way to proceed - more functions, more options in this function? The temporary fix is OK, but I do not really like the idea of just choosing one outlet arbitrarily if there are more than one node with the same lowest value. That could also cause big problems for some people who don’t realize what it is doing.

saisiddu:
My thoughts about the way my watershed is being dealt right now: (nothing to do specifically with the ongoing conversation)

- I like it
- I have two nodes with same elevation, but the results are the same when either of them are selected
- This was probably OK probably because these are adjacent nodes
- I don't know if a watershed can have two same lowest elevation nodes that are not adjacent? I am guessing NOT!

Thanks for the help group.

kbarnhart:
Would it be OK if I merged the temp fix?

gregtucker:
Ok by me!

SiccarPoint:
Sorry, think I'm assigned, aren't I? Go for it

PR #427, opened 2017-01-15
fastscape assert now passes float64; also tiny doc tweak

gregtucker:
FastscapeEroder has an assert to test assumption that if rainfall_intensity is not a field (name, array, etc.), then it is either float or int. But it's possible for a driver to have it be a numpy.float64. That's now added.
Also changed an outdated doc comment saying (incorrectly) that StreamPowerEroder doesn't use fastscape algo.
By the way, why do we have 2 stream power components that use the same algo?

### DIFF
--- a/docs/landlab.components.stream_power.rst
+++ b/docs/landlab.components.stream_power.rst
@@ -6,7 +6,7 @@ FastscapeEroder: Compute fluvial erosion using stream power theory ("fastscape" 
     :undoc-members:
     :show-inheritance:
 
-StreamPower: Compute fluvial erosion using stream power theory (explicit forward-difference solution)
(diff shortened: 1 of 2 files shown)


SiccarPoint:
We have two for two reasons:

1. Backwards compatibility from a time when they didn’t use the same algo, and
2. The syntax is not identical (though that could be fixed). IIRC, only one permits spatially variable rainfall right now, and I think only one permits working directly with discharge.

Re (1), I believe the fastscape algorithm is always preferable to an explicit solution, so we always use it now.

gregtucker:
I need this now, so am going to merge it despite the netCDF fails on mac numpy 1.11, which I think must also impact master already.

Issue #429, opened 2017-01-20
syntax warning in sed_flux_dep_incision


gregtucker:
On import (of another component) the following warning appears:

/Users/gtucker/Dev/Landlab/landlab/landlab/components/stream_power/sed_flux_dep_incision.py:363: SyntaxWarning: assertion is always true, perhaps remove parentheses?
  assert(self.Qc == 'MPM', "Qc must be 'MPM' to return stream " +

I suspect there needs to be an extra paren after 'MPM' and before "Qc

PR #430, opened 2017-01-20
move paren

SiccarPoint:
Should resolve GT’s issue #429


SiccarPoint:
Bump- @kbarnhart ?

kbarnhart:
Looks like the only issues were with the known netcdf-travis issue that is now fixed. Merging.

PR #434, opened 2017-01-24
Make nd fields

kbarnhart:
Changes data field class to permit n-d arrays at grid locations. 

Adds a new unit test.

### DIFF
--- a/landlab/field/scalar_data_fields.py
+++ b/landlab/field/scalar_data_fields.py
@@ -403,18 +403,9 @@ class ScalarDataFields(dict):
             raise FieldError('{name}: already exists'. format(name=name))
 
         value_array = np.asarray(value_array)
-        if value_array.ndim > 0:
-            if value_array.size == self.size:
-                value_array.shape = (-1, )
-                
-        elif np.mod(value_array.size,self.size) != 0:
-            raise ValueError(
-                'size of new array must be divisible by the number of elements'
-                ' in of the field')   
 
         if copy:
             value_array = value_array.copy()
-            value_array.shape = (value_array.size, )
 
         self[name] = value_array
 
@@ -446,11 +437,10 @@ class ScalarDataFields(dict):
 
         if self.size is None:
             self.size = value_array.size
-        if value_array.ndim > 0:
-            if value_array.shape[0] != self.size:
-                raise ValueError(
-                    'size of 0th dimention of the new array'
-                    'must be the same as the field')
+        if self.size == 1 and value_array.ndim != 0:
+            raise ValueError('size mismatch for scalar field')
+        if value_array.ndim > 0 and value_array.shape != (self.size, ):
+            value_array = value_array.reshape((self.size, -1)).squeeze()
 
         if name not in self:
             self.set_units(name, None)
--- a/landlab/field/scalar_data_fields.py
+++ b/landlab/field/scalar_data_fields.py
@@ -64,6 +64,19 @@ class ScalarDataFields(dict):
     Traceback (most recent call last):
     ValueError: total size of the new array must be the same as the field
 
+    Fields can also be multidimensional arrays so long as they can be
+    resized such that the first dimension is the size of the field.
+    The stored field will be resized view of the input array such that
+    the size of the first dimension is the size of the field.
+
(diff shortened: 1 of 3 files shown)


kbarnhart:
Looks like something funny happened with the Travis builds. @mcflugen, can you re-start them. If they pass (except for the netcdf issue), I can merge. 

Thanks for your modifications!

mcflugen:
@kbarnhart: Travis is having some pretty big trouble with their Mac builds. To help out I went ahead and killed all our builds. I waited for a few Linux builds to complete but then canceled the rest. Since things are passing on Linux (with old numpy) and Windows, things are probably OK. I could restart them but I think until things are fixed with Travis, the Mac builds will take a long time to complete.

I think before merging we may want to add a few more tests for the new fields and also run the change by some of the others (maybe @gregtucker or @SiccarPoint?). I assume you're OK with it, is that right?

@gregtucker, @SiccarPoint: This pull request does a number of things, but one thing it does is (slightly) change how arrays are stored when they are added as fields. With this PR, when an array is added to a grid as a field, the array is stored as a reshaped version of itself such that the first dimension is the number of elements (the input array is unchanged, a view of it with the new shape is stored). As an example, if you were to add an array that is of size `2 * n_node ` as an `at_node` node field you will not get an error but instead store the array as shape `(n_nodes, 2)`. That is, at every node the value will be an array (of length 2) rather than a scalar. If you add an array of length `n_nodes`, nothing changes from before. Do you see any problems with that?

gregtucker:
I encouraged this modification, because it's clear we'll have cases in which we need 2D arrays as fields, wherein one dimension contains multiple values associated with a particular element. Can't think of any problems; seems unlikely that a user would accidentally pass a mis-sized array that just happens to have N x M elements.

mcflugen:
Thanks, @gregtucker.

@kbarnhart, @gregtucker: What do you think about "global" values? To make all the tests pass, I require that global values (that is, a single value for an entire array) can only be scalars. Should we relax that requirement as well? Maybe someone would want a global value to be an array?

kbarnhart:
When you say "global" values, do you mean those stored at_grid? 

I'm slightly confused by the change you made regarding the scalar fields... I *think* that what the change you made does is to say that if the array passed has a size of one, but has more than one dimension, a value error is raised. 

I think the better thing to do is to test if the size of the value array size is also 1. if it is, then we are OK no matter how many empty dimensions there are (we can just squeeze it as you have done in the next case). 

I should be able to do that later today (won't take long but have to prep for a meeting happening soon).

SiccarPoint:
Concur with Greg. Sounds good.

mcflugen:
Yes, by "global" I meant `at_grid`.

I've now changed it so that even `at_grid` values can be arrays (of any size). I think that makes more sense and is consistent.

If a numpy array is provided as a field and is of the correct shape, a reference to that array is used. Otherwise, the array is reshaped and that array stored, which will be a view of the original array. This is true for either `at_grid` fields or `at_node` etc.

SiccarPoint:
This all sounds great. This is mainly just a note that we also need to update the docs as part of this merge, and also to note that my PR #437 probably needs to go in before that happens. It moves around a lot of the text that describes this functionality on the grid. I can probably do this once someone checks that PR over.

kbarnhart:
Maybe I'm confused. 

@SiccarPoint do you mean that we need to change the documentation about the fields (like here https://github.com/landlab/landlab/wiki/Grid#adding-data-to-a-landlab-grid-element-using-fields) before merging this? And that we should wait for you to finish the auto-build-docs work first before doing this. 

Or do you mean that we need to improve the documentation, but not necessarily before the merge?

Or do you mean something entirely different?

And @mcflugen, thanks for making that change!

SiccarPoint:
I mean the second thing - just remember this adds functionality that needs manual listing in the docs, basically. Doesn’t necessarily need to be done before this merge, but it would make sense as part of it so we don’t forget.

The auto-doc pull is now ready to go, if someone could take a look.

kbarnhart:
@SiccarPoint thanks for clarifying. I can put in the changes once the new docs are up. 

OK with everyone if we merge this?

gregtucker:
ok by me

PR #436, opened 2017-01-29
Jk docs deep dive

jennyknuth:


SiccarPoint:
I'm not sure about what's been added to `landlab.grid.base.rst`, i.e., the stuff from `grid_funcs`. I believe the contents of `grid_funcs` that we'd want is already replicated as a grid method in the list above, and the other stuff is both in a very antiquated naming style and I think superseded by other grid methods (i.e., `calculate...`).

I also say this pragmatically, as when this file gets rendered as html, the new grid funcs bit just gets appended at the bottom of the thousands of lines of base grid methods.

If we want these kinds of functions that take grids (as opposed to grid methods sensu stricto) documented, I think we should be listing them somewhere separately in the docs, not in amongst/alongside the grid methods.

jennyknuth:
Good feedback. As it was before, grid_funcs was being referenced in the toc
without a corresponding .rst file. It is a little cryptic to me what you
are saying. Are you saying that grid_funcs.py is out of date, doesn't need
to be in the docs at all, or should go somewhere else?

We all share the ultimate goal of a sustainable balance between
autogeneration and curation and it's not there yet: as it is, the
autogeneration is unreadable and the curation is out of date…

Jenny

SiccarPoint:
I’m trying to improve the automation right now! Things should improve notably…

gregtucker:
I'm going to merge this so @siccarpoint can proceed from the most up to date version of docs

PR #437, opened 2017-01-31
SiccarPoint/auto build docs

SiccarPoint:
This pull adds functionality to auto-generate our documentation that5 lists grid functions. The script `docs/build_grid_func_docs.py` gathers the necessary information from the LL code base, then overprints the rest files for the grid types with the new text including the most current methods. We still use the LLCATS system, though it would be relatively straightforward to also parse the _names_ of the methods to extend the system beyond LLCATS.

All this means we no longer use the module level docstrings to build our grid docs - the rest files are directly constructed by the new script. This means the text in the modules is now duplicative. I propose we remove it entirely, as users would not see it anyway (it’s module level, not class level) - but would appreciate a 2nd opinion.


SiccarPoint:
@eric: What I haven't done is attempt to weave `build_grid_func_docs.py` into the set of commands that gets called every time we have a new build/commit. Mainly because I don't know how/where this code is. Could you remind me? I believe just running this script should be totally adequate now.

mcflugen:
@SiccarPoint: To have build_grid_func_docs.py called whenever the docs are built, you'll have to add the command into the Makefile. I think under the `pre-hook` target should work.

We definitely should remove any documentation that is ignored. If all module-level documentation for the grid stuff is really ignored, it should be replaced with a string saying as much. If not, in the future, folks may want to add some documentation and wonder why it's being replaced. That raises a question though: where should a user put that kind of documentation? I guess that would be in `text_for_*.py.txt`?

SiccarPoint:
OK, I _think_ I did this right? Please check I put it where you thought @mcflugen ! Ready to merge if the makefile change looks OK...

mcflugen:
@SiccarPoint: Yes, that's the right place. Looks good!

A couple things:
* You *may* want to move your doc utilities (`get_categories_from_grid_methods`, etc.) out of the landlab package and into the `docs` folder. It's probably not necessary but it keeps things separate and allows building the docs without installing landlab.
* You should remove all of the auto-generated files from version control. No need to have files in version control that are overwritten.
* Under **Grid types** in `index.rst`, it would be nice if there were links to your grid pages.

SiccarPoint:
How do I remove things for version control? Not done that before...

I'll add those links. For now, I'll leave the functions within Landlab, as they do also have a role there. But we could reconsider later.

mcflugen:
@SiccarPoint: `git rm <filename>` should do the trick. I think it's totally fine leaving your documentation functions in landlab.

SiccarPoint:
Did it, hopefully that worked. Should be ready to merge once tests complete. Could also create a new release?

Issue #432, opened 2017-01-22
netCDF4 will not import


gregtucker:
This is not a Landlab issue per se, but it is causing some of our netcdf tests to fail. 

(1) verify that I have netcdf4 installed:

Greg-Tuckers-MacBook-Pro:landlab gtucker$ conda install netcdf4
Fetching package metadata .........
Solving package specifications: .

# All requested packages already installed.
# packages in environment at //anaconda:
#
netcdf4                   1.2.4               np111py27_1  

(2) now try to import it:

Greg-Tuckers-MacBook-Pro:landlab gtucker$ python
Python 2.7.13 |Anaconda custom (x86_64)| (default, Dec 20 2016, 23:05:08) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> import netcdf4
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named netcdf4

(3) try it with capitalized "CDF":

>>> import netCDF4
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "//anaconda/lib/python2.7/site-packages/netCDF4/__init__.py", line 3, in <module>
    from ._netCDF4 import *
ImportError: dlopen(//anaconda/lib/python2.7/site-packages/netCDF4/_netCDF4.so, 2): Library not loaded: @rpath/libmfhdf.0.dylib
  Referenced from: /anaconda/lib/libnetcdf.11.dylib
  Reason: image not found
>>> 

@mcflugen do you have time to look into this?

kbarnhart:
Unless I'm mistaken, this also prevents running

```
python test-installed-landlab.py
```

from finishing.

Mine ends with `FAILED (SKIP=7, errors=2, failures=1)` and ends up not catching some errors I know I've (temporarily) induced.

PR #439, opened 2017-02-01
Fix issue 432

mcflugen:
This PR kind of fixes issue #432. The problem is that installing from the default Anaconda channel (installing from the conda-forge channel works) creates a version conflict (between numpy-1.11, jpeg-9b, and hdf4-4.2.12) when using netcdf4-1.2.4. This seems to only be a problem on Mac and Linux and numpy-1.11.

I've added a requirement in `meta.yaml` for `hdf4 >= 4.2.12` for the problem build cases. I hope this will resolve itself and, in the future, we can remove this requirement. Anyway, this should get our builds on `master` passing again.

### DIFF
--- a/.conda-recipe/meta.yaml
+++ b/.conda-recipe/meta.yaml
@@ -15,7 +15,7 @@ requirements:
     - matplotlib
     - numpydoc
     - netcdf4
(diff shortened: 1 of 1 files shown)


SiccarPoint:
Looking good. Merging.

PR #443, opened 2017-02-19
Gt add q stream power smooth thresh

gregtucker:
o Fixes a mistake in smooth-threshold stream power solver ("epsilon" includes elevation of downstream receiver node, and so needs to be calculated within the node loop rather than ahead of time)
o Enables possibility of passing an array or field name to use in place of drainage area (i.e., a field to use as discharge)


kbarnhart:
looks good to me. you have one travis build that quit and won't load the log file... That usually is a bug on the travis side, rather than landlab, so I'll merge.

PR #445, opened 2017-02-20
Jadams15/depth slope product

jadams15:
new shear stress component, fixed bugs in green ampt, updated OF to work with spatially variable manning's n.


nicgaspar:
Ready to go as soon as tests pass.

PR #450, opened 2017-03-02
Kbarnhart/stream power error

kbarnhart:
A few errors in stream power discovered through use. 

1. accidental conversion of flooded_nodes to boolean if it is provided as at_node array

2. making the critera uniform between the two functions in the cfuncs.pyx in stream_power.


kbarnhart:
Looks like with the exception of two travis builds that stopped after <20seconds, this works.

PR #452, opened 2017-03-04
Gt create depth dep cubic diffuser

gregtucker:
This component combines features of Rachel's depth-dependent LINEAR diffuser and (not depth dependent) cubic diffuser. It includes a unit test against an analytical solution, which is worked out in a Jupyter notebook in the tests folder.


kbarnhart:
Except for a few Travis builds that failed to start, this looks good. Merging.

PR #456, opened 2017-03-10
Gt fix fastscape variable k

gregtucker:
Minor change to FastscapeEroder: previously, when user passes an array as K_sp, the component makes a copy. Better if the user can change the K_sp field dynamically, however. Now changed to use a reference to the passed K_sp array, so if user changes the array, FastscapeEroder will "see" the changes.

@kbarnhart or @SiccarPoint could you take a quick look and merge please?


kbarnhart:
I can look once it's done checking

PR #459, opened 2017-03-12
Kbarnhart/consistent parameter names

kbarnhart:
Changes parameter names and input parameters for three hillslope components

CubicNonLinearDiffuser
DepthDependentCubicDiffuser
DepthDependentDiffuser

that took as input parameters a variety of "efficiencies" that were derivatives of the hillslope diffusivity and other given input parameters. 

In order to move towards standardization, all have been replaced with linear_diffusivity and small bits of the component code has changed to account for conversion.


kbarnhart:
I've fixed the error you caught @gregtucker.

PR #460, opened 2017-03-20
Cmshobe hybrid alluvium model

cmshobe:
SPACE model


gregtucker:
Charlie, this looks good. Nice work. 'm going to go ahead and merge so we can start crafting models with it. One suggested fix: at the end of run_one_step, you assign things like *self.q* to their respective fields. I think it would be more economical to assign these vars as references TO the fields in __init__. For example, if you do

self.elev = grid.at_node['topographic__elevation']

then (as long as you're careful) your self.elev will thereafter refer to the correct elevation array, and any changes you make to self.elev will also be made to topographic__elevation (because they are one and the same). The only thing to watch out for is operations like

self.elev = (some array calculation)

which can break the reference by creating a brand new array and assigning the self.elev reference to the new one. The way to avoid this is:

self.elev[:] = (some array calculation)

PR #462, opened 2017-03-24
Kbarnhart/make raster netcdf

kbarnhart:
Changes to add write_raster_netcdf a netcdf writer that puts node x and node y information only along ni and nj dimensions instead of as (ni, nj) arrays.

Modifications to read_netcdf to make it able to read from netcdfs written this way.

### DIFF
--- a/landlab/io/netcdf/read.py
+++ b/landlab/io/netcdf/read.py
@@ -136,6 +136,34 @@ def _read_netcdf_structured_grid(root):
 
     return coordinates
 
+def _read_netcdf_raster_structured_grid(root):
+    """Get node coordinates for a structured grid written as a raster.
+
+    Parameters
+    ----------
+    root : netcdf_file
+        A NetCDF file.
+
+    Returns
+    -------
+    tuple of ndarray
+        Node coordinates for each dimension reshaped to match the shape
+        of the grid.
+    """
+    shape = _read_netcdf_grid_shape(root)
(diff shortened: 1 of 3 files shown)


kbarnhart:
@mcflugen do you think this looks OK as a short term netcdf raster writer/reader before your graph functions debut?

landlab/io/netcdf/read.py
```diff
@@ -247,8 +275,11 @@ def read_netcdf(nc_file, just_grid=False):
     except TypeError:
         root = nc4.Dataset(nc_file, 'r', format='NETCDF4')
 
-    node_coords = _read_netcdf_structured_grid(root)
-
+    try:
+        node_coords = _read_netcdf_structured_grid(root)
+    except:
```
mcflugen: No bare `except` statements. Please add a specific exception or, an `if` statement that checks for the grid type.

Maybe something along the lines of,
```python
if is_raster_grid(root):
    node_coords = _read_netcdf_raster_structured_grid(root)
else:
    node_coords = _read_netcdf_structured_grid(root)
```

kbarnhart:
What do you think of this? - assuming it passes tests. 

I added a specific exception and then included an if statement test to make sure that the criteria the function `_read_netcdf_raster_structured_grid` expects are present in the netcdf dataset. 
 
Thanks for catching that. The reviewing functionality is neat!

PR #463, opened 2017-03-30
no deposit option for linear diffusion

nicgaspar:
deposit flag added to linear diffusion for new option to make diffusion only erode. Material that would be deposited 'disappears' - presumably because it is taken up by fluvial channels.


jadams15:
I'm going to make one minor change, which is to change the flag to 'True' or 'False'. This will remove the 'if' statement where it sets self._deposit. VERY nitpicky, but streamlined a bit. Will do this now.

PR #464, opened 2017-04-01
add basic version of implicit kinwave model

gregtucker:
This is a new component that implements an order-N "node by node" implicit solution to the kinematic wave approximation of the shallow-water equations. The basic idea is that if you can process the nodes in upstream-to-downstream order, you can always know the inflow discharge at a particular node. With that, and assuming that the depth of water at the node sets the outflow to the adjacent downhill nodes, it's possible to do an implicit solution for water depth at the new time step. Here, more generally, it's weighted between H at the new and old time steps, with a user-specified weight factor. The algorithm uses the MFD FlowAccumulator, which is order-N. Moreover, the algorithm itself should be order-N, since it sweeps through nodes once.

### DIFF
--- a/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
+++ b/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
@@ -17,7 +17,9 @@ import numpy as np
 
 def water_fn(x, a, b, c, d, e):
     """Evaluates the solution to the water-depth equation.
-    
+
+    Called by scipy.newton() to find solution for $x$ using Newton's method.
+
     Parameters
     ----------
     x : float
--- a/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
+++ b/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
@@ -228,7 +228,8 @@ class KinwaveImplicitOverlandFlow(Component):
 
         # Instantiate flow router
         self.flow_accum = FlowAccumulator(grid, 'topographic__elevation',
-                                          flow_director='MFD')
+                                    flow_director='MFD',
+                                    partition_method='square_root_of_slope')
 
         # Flag to let us know whether this is our first iteration
         self.first_iteration = True
--- a/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
+++ b/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
@@ -16,6 +16,41 @@ import numpy as np
 
 
 def water_fn(x, a, b, c, d, e):
+    """Evaluates the solution to the water-depth equation.
+    
+    Parameters
+    ----------
+    x : float
+        Water depth at new time step.
+    a : float
+        "alpha" parameter (see below)
+    b : float
+        Weighting factor on new versus old time step. $b=1$ means purely
+        implicit solution with all weight on $H$ at new time step. $b=0$ (not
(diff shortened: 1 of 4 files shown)


landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -0,0 +1,320 @@
...
+    ----------
+    grid : ModelGrid
+        A Landlab grid object.
+    precip_rate : float, optional (defaults to 1 mm/hr)
+        Precipitation rate, mm/hr
+    precip_duration : float, optional (defaults to 1 hour)
+        Duration of precipitation, hours
+    infilt_rate : float, optional (defaults to 0)
+        Maximum rate of infiltration, mm/hr
+    roughnes : float, defaults to 0.01
```
kbarnhart: Do we want to support spatially values for quantities like precipitation rate, infiltration rate, and roughness? 

I think this would just require checking that they are either single valued quantities or at node quantities. 

Other than that this looks good to me?

landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -0,0 +1,320 @@
...
+"""
+
+
+from landlab import Component
+from landlab.components import FlowAccumulator
+from scipy.optimize import newton
+import numpy as np
+
+
+def water_fn(x, a, b, c, d, e):
```
kbarnhart: Needs docstring, and perhaps more intuitive function and variable names. Looks like this is the function you use Newtons method on in order to calculate water depth.

landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -0,0 +1,320 @@
...
+                # each node, the proportion of flow that heads out toward each
+                # of its N neighbors. The proportion is zero if the neighbor is
+                # uphill; otherwise, it is S^1/2 / sum(S^1/2). If for example
+                # we have a raster grid, there will be four neighbors and four
+                # proportions, some of which may be zero and some between 0 and
+                # 1.
+                self.disch_in[self.grid.neighbors_at_node[n]] += (outflow
+                    * self.flow_accum.flow_director.proportions[n])
+
+                # TODO: the above is enough to implement the solution for flow
```
kbarnhart: I agree that we want to eventually included flow velocity and discharge on links.

landlab/components/overland_flow/tests/test_kinwave_implicit.py
```diff
@@ -0,0 +1,146 @@
...
+
+    # Try with default runoff rate of 1 mm/hr = 2.78e-7 m/s
+    kw = KinwaveImplicitOverlandFlow(rg)
+    assert_equal(round(kw.runoff_rate * 1.0e7, 2), 2.78)
+    kw.depth[:] = 0.0
+    for i in range(18):
+        kw.run_one_step(10.0)
+
+    # Look at a column of nodes down the middle. The inflow from uphill should
+    # be, from top to bottom: 0, 0.004, 0.008, 0.012, 0.016, 0.02, 0.024, 0.028
```
kbarnhart: I think this comment is incorrect (pasted from line 87 example).

kbarnhart:
I think this looks good enough to merge @gregtucker, unless you had other things you were planning on adding. They remaining issues to address would just add general functionality.

PR #467, opened 2017-04-07
inits for Hybrid Alluvium

cmshobe:
made/edited __init__ files for Hybrid Alluvium component @kbarnhart

### DIFF
--- a/landlab/components/__init__.py
+++ b/landlab/components/__init__.py
@@ -29,6 +29,7 @@ from .flow_director import FlowDirectorDINF
 from .flow_accum import FlowAccumulator
 from .cubic_nonlinear_hillslope_flux import CubicNonLinearDiffuser
 from .depth_dependent_cubic_soil_creep import DepthDependentCubicDiffuser
+from .hybrid_alluvium import HybridAlluvium
 
 COMPONENTS = [ChiFinder, LinearDiffuser,
               Flexure, FlowRouter, DepressionFinderAndRouter,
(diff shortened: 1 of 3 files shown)


kbarnhart:
@cmshobe it looks like you have some python 2-python 3 compatibility issues (only 3.5 and 3.6 fail). You can either create a python3 environment on your computer for testing or we can use my computer. 

In the builds I look at I see the following error:
```
DeprecationWarning: elementwise == comparison failed; this will raise an error in the future.
  flooded = self._grid.nodes.flatten() == flooded_nodes
```
and the log is truncated so I can't see the full reason the component failed. 

This error often occurs if broadcasting must be done for the comparison. 

Looking at your unit tests it looks like flooded nodes is an array of node IDs that are flooded, constructed by np.where(). In the above lines it is being used as a boolean array of length number_of_node.

kbarnhart:
ahh. I see you can download the full log file.

mcflugen:
It looks to me like you are trying to do a relative import in Python 3 but imports in Python 3 default to absolute. In your doctest for `HybridAlluvium` you have
```python
from hybrid_alluvium import HybridAlluvium
```
This should be
```python
from landlab.components import HybridAlluvium
```
The reason this works in Python 2 is that the doctest is run in the same folder as `hydrid_alluvium.py`. Doctest should always be written as examples to show how an end user would run your component.

cmshobe:
Thanks @kbarnhart and @mcflugen ! I've fixed the deprecated == and also the doctest import statement; we'll see if there's anything else

mcflugen:
Your tests are failing for your deprecation fix. I'm not totally sure what you are wanting to do so this might not be the most efficient way of doing it but instead of,
```python
flooded = np.equal(self._grid.nodes.flatten(), flooded_nodes)
```
something like the following should work,
```python
flooded = np.full(grid.number_of_nodes, False, dtype=bool)
flooded[flooded_nodes] = True
```

PR #487, opened 2017-05-31
Removing unnecessary lines in HybridAlluvium component

kbarnhart:
This subset of float conversions are superseded by code just below this block that tests for different possible input types (e.g. field name as a string).

@cmshobe could you add some tests that include providing K_sed, K_br, sp_crit_sed, sp_crit_br as float, field name, or array so that tests test this functionality. Also a test that provides something you expect to break this functionality.


cmshobe:
Whoops! Good catch @kbarnhart -- these changes look good. In the near future I'll improve the doctests for different types of inputs. @gregtucker this shouldn't cause any problems if you pull it into the master branch.

PR #488, opened 2017-06-06
fixed variable assignment, q=a^m works now

cmshobe:
@kbarnhart sorry for being a dum dum but what I had done is accidentally converted the discharge method to a string before the component checked whether it was None or not. Fixed now.


kbarnhart:
@cmshobe looks like these fail on two docstring tests. Is this because the value for A**m is now correct and you didn't update the expected value (since it is pretty hard to determine an a-priori analytical solution for this component)... or was this unexpected. 

looking at the commit, it doesn't look like you updated the docstring. 

once it passes, I take a look again and merge.

cmshobe:
Doctest should be fixed now.

kbarnhart:
Since these are just the know NetCDF errors, I'll merge.

Issue #489, opened 2017-06-06
FlowDirectorMDF and FlowDirectorDINF produce slopes instead of gradients


kbarnhart:
FlowDirectorSteepest and FlowDirectorD8 produce gradients, but the new two don't.

PR #490, opened 2017-06-06
Switching  MFD and dINF methods from returning slope to returning gra…

kbarnhart:
…dient.

# this PR addresses Issue #489


kbarnhart:
@mcflugen the errors that have occurred thus far are related to NetCDF4. I just dealt with this on my computer and the way I fixed it was:

```python
conda uninstall netCDF4
pip install netCDF4
python setup.py develop -u
python setup.py develop 
```

if this is the only type of error, OK to have @gregtucker merge?

Issue #491, opened 2017-06-06
Fastscape and StreamPower Eroder have solver convergence issues with n<1 (mostly n<0.67)


kbarnhart:
This is an issue for using the shear stress formulation. 

I've found that it can be improved with smaller time steps. 

Perhaps using a different solver would help...?

Issue #475, opened 2017-04-21
problem with stream power when n < 1


nicgaspar:
I know this just came up with @awickert. I'm sure I have the most recent version of the code and looked at the issues (#431, #444, #433) Dan cited for Andy. @SiccarPoint point told Andy that it was solved with issue #433. I checked what happened with that, and it seems @gregtucker added some code to fix an infinite look in the cfuncs.  But the fix is kicking me out because of that solution.  To be fair, if that solution wasn't there, then I'd run into an infinite loop.

More details below. Anyway, just asking in case it spurs something for @gregtucker or @SiccarPoint because they know this code well. Is it that I need to run a sink filler for this to work? The problem is definitely a flat slope.

Important code lines:

fr = FlowRouter(mg, method='D4')
sp = FastscapeEroder(mg, K_sp=0.00005, m_sp=0.4, n_sp=0.8, threshold_sp=0.,
                    rainfall_intensity=1.)
fr.run_one_step()
sp.run_one_step(dt)

What happens:

  warnings.warn(_use_error_msg)
===============
4.50000007032
4.50000007032
nan
nan
nan
0.0776460875399
===============
4.50000007032
4.50000007032
nan
nan
nan
0.0776460875399
===============
4.50000007032
4.50000007032
nan
nan
nan
0.0776460875399
===============
4.50000007032
4.50000007032
nan
nan
nan
0.0776460875399
===============
4.50000007032
4.50000007032
nan
nan
nan
0.0776460875399
===============
4.50000007032
4.50000007032
nan
nan
nan
0.0776460875399
===============
4.50000007032
4.50000007032
nan
nan
nan
0.0776460875399
===============
4.50000007032
4.50000007032
nan
nan
nan
0.0776460875399
===============
4.50000007032
4.50000007032
nan
nan
nan
0.0776460875399
===============
4.50000007032
4.50000007032
nan
nan
nan
0.0776460875399
Traceback (most recent call last):

  File "<ipython-input-1-4ead74e3e67e>", line 1, in <module>
    runfile('/Users/nicgaspar/LandlabDiffusionFluvialExamples/diffusion_fluvial_test_one_outlet.py', wdir='/Users/nicgaspar/LandlabDiffusionFluvialExamples')

  File "/Users/nicgaspar/anaconda/lib/python2.7/site-packages/spyder/utils/site/sitecustomize.py", line 866, in runfile
    execfile(filename, namespace)

[...]

PR #492, opened 2017-06-06
New solver for Fastscape?

kbarnhart:
A revised solver for Fastscape and StreamPower erodes which I **think** is more stable than the present solver. 

Uses Brent's method. 

Potentially addresses Issue #491


kbarnhart:
A few more notes. I haven't added a short example of the solver not converging. Will do so soon.

gregtucker:
Looks like some netCDF4 errors under Numpy 1.11 on MacOS

kbarnhart:
@gregtucker that seems to be coming up other places too. 

I've also found in use, that this implementation is slower than @SiccarPoint 's earlier implementation.
Good news, however is that with n = 0.67 it is stable where it previously gave SP solver errors (sorry for not having posted simple code yet...)

May make some progress on speed tonight. 

I'll let you all know when its worth talking a more careful look at.

kbarnhart:
@mcflugen how do I push without triggering CI builds?

SiccarPoint:
Bump me when it's time to take a look

kbarnhart:
@SiccarPoint I've now finished cythonizing what I did. Feel free to take a look as you wish. 

If you want to Skype so I can more quickly explain what I've done, let me know. In brief, I reformulated Braun and Willet's appendix equations to include a term for a threshold. Assuming enough erosion occurs to exceed that threshold, then the function to find a root will have a root between x = 0 and 1, where x is the normalized amount of erosion. I check that erosion will occur by calculating f(x=1) before doing any root finding. I've played around with this function in the Desmos Graphing Calculator (https://www.desmos.com/calculator/yjltmbsaew) and for reasonable values (a>0, n>0, b>0) so long as some erosion occurs (e.g. f(x=1)>0) there is a root between zero and one. 

There is one place where I'm not 100% sure if it can be further cythonized... I'll make a note in the code. Perhaps you, @mcflugen  or @gregtucker will have recommendations. 

I've timed this on a larger model grid that we're using and I get similar time-per-timestep as @SiccarPoint 's original code.

landlab/components/stream_power/cfuncs.pyx
```diff
@@ -4,6 +4,9 @@ cimport cython
 
 #from libc.math cimport fabs
 
+# as best as KRB can tell, the function _brentq is in c.
+# not sure what, if anything I need to do to 
+from scipy.optimize._zeros import _brentq as brentq
```
kbarnhart: The part of the code that I would point out as potentially needing the most improvement is making sure that the code used for the brentq root finding is fully cythonized. 

As best as I can tell, this function, spicy.optimize._zeros._brentq is in C and is what is actually run when one uses
scipy.optimize.brentq.

However there may be a better way to import it into this cfuncs file. 

Most of what I know about cython I learned this morning... so if @SiccarPoint @mcflugen or @gregtucker have recommendations, please commit.

kbarnhart:
@nathanlyons 
I know you've been experimenting with different values for n. Perhaps you want to test this for stability in your use case.

kbarnhart:
Looks like it passes except for the NetCDF issues.

nathanlyons:
@kbarnhart 
Yes, this fixed our problem (Issue #475). Before, n < 0.9 in StreamPowerEroder would give the error, "failure to converge in SP solver". With this branch, I went as low as n=0.6 without getting this error.

kbarnhart:
@nathanlyons thanks for checking! Glad that it worked. 

We discussed this in the landlab call today and @SiccarPoint will look through code tomorrow. 

I merged the new master with the netcdf updates @mcflugen made in PR #495 and so far tests seem good (one of the travis builds seems to have been canceled).

landlab/components/stream_power/cfuncs.pyx
SiccarPoint: Python docs says "Calling NumPy/SciPy functions currently has a Python call overhead; it would be possible to take a short-cut from Cython directly to C. (This does however require some isolated and incremental changes to those libraries; mail the Cython mailing list for details)."

Not sure what this is getting at, process wise. The easiest way (though kind of a crappy option) might be to find the function in scipy, get hold of its Cython code, and just paste the function part into our cfuncs...?

SiccarPoint: IIRC I wrote out the Newton method longhand in my original version, for this exact reason. (Could be remembering this wrong, but I definitely also hit this issue and concluded import was not ideal, for this reason).

SiccarPoint: The cython can be seen at https://github.com/scipy/scipy/tree/v0.14.0/scipy/optimize/Zeros; note that when this is implemented in scipy there are a few checks in pure Python that wrap the C code: https://github.com/scipy/scipy/blob/v0.14.0/scipy/optimize/zeros.py#L307

@mcflugen probably has the best handle on how this should work. An import of the _zeros._brentq cython function direct would seem to be the way to go, but beyond my technical skill.

SiccarPoint: ETA: I'm clearly behind the times on this. I think the direct import of _brentq that you already do is, as you note in the code, the optimum you can achieve. So we're looking good I think.

kbarnhart:
@SiccarPoint Thanks for the comments and the PEP8. You are right, both of these branches added a bunch of material to the streampower/cfunc.pyx

I'll fix the conflicts and if everything passes, I'll merge.

kbarnhart:
Also, regarding speed. If @mcflugen or others have a better idea for a better way to import the c version of brentq, I'd love to know. I _think_ what I'm importing is in c... but not 100% sure. At minimum my testing thus far indicates that using the _brentq as it is imported was at least as fast as the your algorithm.

landlab/components/stream_power/cfuncs.pyx
mcflugen: I suspect the way you've imported `_brentq` isn't quite right and is slowing things down. Perhaps try:
```python
from scipy.optimize._zeros cimport _brentq as brentq
```
and see if that speeds things up at all.

kbarnhart:
@mcflugen it looks like when I try to import the way you recommended
`from scipy.optimize._zeros cimport _brentq as brentq`
I get the following error when I run `python setup.py develop`
`landlab/components/stream_power/cfuncs.pyx:9:0: 'scipy/optimize/_zeros/_brentq.pxd' not found`

kbarnhart:
Is there something that I I need to do to make the pdx file?

Issue #493, opened 2017-06-08
Hybrid Alluvium component has a part that could be cythonized


kbarnhart:
The component is relatively slow at the moment and this is primarily related to a portion of the code that cannot be vectorized. It can, however, be cythonized.

PR #494, opened 2017-06-08
Cythonize the qs_in routine

kbarnhart:
This makes the hybrid alluvium component faster and addresses issue #493. 

This commit also changes an error in the hydrology routines:
 discharge_method should have been self.discharge_method and area_method should have been self.area_method

@cmshobe please add tests that use each of these different hydrology routines and provide area and discharge methods in all of the possible forms (str, none, etc).

### DIFF
--- a/landlab/components/hybrid_alluvium/cfuncs.pyx
+++ b/landlab/components/hybrid_alluvium/cfuncs.pyx
@@ -1,7 +1,9 @@
 import numpy as np
 cimport numpy as np
 cimport cython
-from math import exp
+
+cdef extern from "math.h":
+    double exp(double x) nogil
 
 DTYPE_FLOAT = np.double
 ctypedef np.double_t DTYPE_FLOAT_t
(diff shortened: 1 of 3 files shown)


kbarnhart:
@mcflugen This is the first time that I've added new cython code. 

I added a new extension to the setup.py file. Let me know if I need to add other code to make the new cython work for all install methods.

mcflugen:
@kbarnhart: I can have a closer look at this later but a couple things after a first look.

1. Have you done any speed tests on this? I suspect one thing that may slow it down is your use of `exp` that you've imported from `math`. You probably import `exp` in the same way that you did with `pow` in your stream power cython code.
2. It seems like this could be done in pure Python without loops. Did you try that?
3. Could you please fix the indentation? The `else` block is a little difficult to read (I suggest to not use the `\`s but, instead, use parenthesis).

kbarnhart:
@mcflugen This algorithm  requires iteration from upstream to downstream because calculating qs at one node, requires knowing qs_in at that node, which requires having calculated qs at all the upstream nodes. So it is similar to the fastscape erosion routine, except that it must go from upstream to downstream instead of downstream to upstream. 

 I've just pushed up a commit that fixed the else block, added some more comment that describe what the function is doing and explain why it can't be taken out of the loop, and changed the exp import thanks for the recommendations. 

I haven't done any formal timing, but informally putting this portion of @cmshobe's hybrid model in cython speeds up runs that I've done using it.

mcflugen:
Ah. I see. So `stack_flip_ud` can have repeated values?

Some ideas (that may or may not be good ones) for optimizing:
* Line 54 should be within the `else` block, I think.
* Floating point equalities are always a little dangerous. Could you use `q[node_id] > 0` instead? Or, maybe even better, precompute non-zero `q`s using numpy?
* It may also be quicker to sort `stack_flip_ud ` first and then iterate through it that way. In doing so, you would be looping over values that are held more closely to one another in memory.

Again, just some ideas. There's not need to spend time optimizing something if it proves it doesn't need optimizing.

@gregtucker has some useful tools for checking cython performance but I forget what they are. @gregtucker?

kbarnhart:
thanks @mcflugen , I've made most of those fixes. 

`stack_flip_ud` is the drainage stack, flipped upside-down, so iteration goes from upstream to downstream. If flipped it in the main python code rather than in this routine. It doesn't have repeated values. Sorting it doesn't make sense since it has the order we need to move through the drainage network correctly.

kbarnhart:
Looks like with the updates @mcflugen made about netCDF in PR #495 this passes. @gregtucker , merge?

PR #502, opened 2017-06-21
Cmshobe/fix hybrid q mechanics

cmshobe:
@kbarnhart:

There are a lot of changes here:

First and most importantly, the cfuncs loop for calculating Qs has changed, and now results in the model matching analytical predictions.

Second, all hydrology options support a big Q (defaults to surface_water__discharge) and a lil Q (generally either A^m or Q^m, depending on the formulation of choice).

Assorted other small bug fixes that I found along the way.


kbarnhart:
@cmshobe great! thanks! 

@gregtucker , the three of us should discuss having both a big Q and a little q represented. Perhaps tomorrow at group meeting?

PR #504, opened 2017-06-27
Add logging functions and utilities

mcflugen:
This pull requests adds a few new functions to `landlab.core.utils` that aid in printing landlab-style messages.

The main new functions are:
* `format_message`, (and the shortcut functions `error_message` and `warning_message`) which formats a text block by dedenting paragraphs and wrapping lines to 70 characters.
* `assert_or_print`, which makes an assertion and, based on the `onerror` keyword, if the assertion fails will either: ignore the failure (`pass`), print a warning message but continue (`warn`), or print and error message and raise an exception (`raise`).


kbarnhart:
@mcflugen The only comment that I have is that some documentation (in docstring? somewhere else?) that describes recommended use of these logging and printing functions would be helpful. 

Otherwise looks good to me.

mcflugen:
@kbarnhart, I added some documentation to the top of the `messages` module that describes how to use these functions. Have a look. Is that what you had in mind?

kbarnhart:
@mcflugen The documentation helps a lot! I think there is just one question regarding use of `error_message` that the document doesn't answer. As best as I can tell, the `error_message` doesn't actually result in an error being raised. Am I mistaken about this?

I think that this means that I should use `error_message` to format things nicely within my own "if bad condition, then assert error" blocks to throw errors and use the message functions you've written within them. Alternatively, if my test fits within the purview of `assert_or_print`, I should use it.  I might expand a section you wrote in the following way. Is this correct?

Other than this clarification, I think that this looks awesome!

```python
landlab also provides functions for printing warning and error
messages. 

>>> from landlab.core.messages import warning_message
>>> message = ('Lorem ipsum dolor sit amet, consectetur\\n'
...            'adipiscing elit, sed do eiusmod tempor\\n'
...            'incididunt ut labore et dolore magna aliqua.')
>>> print(warning_message(message))
WARNING
=======
<BLANKLINE>
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
eiusmod tempor incididunt ut labore et dolore magna aliqua.

If a developer wants an error message to accompany a particular
type of error, the error_message method can be used along with 
raise in the following way.  

>>> from landlab.core.messages import error_message
>>> if error_condition == True:
>>>    raise RuntimeError(error_message(message))
ERROR
=====
<BLANKLINE>
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
eiusmod tempor incididunt ut labore et dolore magna aliqua.

Alternatively, if a developer does not desire a formal error to be 
raised, the error_message function error can be used just to
print the message. 

>>> from landlab.core.messages import error_message
>>>print(error_message(message))

Another common design pattern used in landlab is to allow the
user, usually through a keyword, to control what happens
if a particular assertion fails. For instance, the user
may want the code to raise an error, or print a warning
message, or do nothing at all. The `assert_or_print` function
should be used in these cases.

>>> from landlab.core.messages import assert_or_print

>>> dt = 1e6
>>> assert_or_print(dt < 1, 'Unstable time step!', onerror='pass')
>>> assert_or_print(dt < 1, 'Unstable time step!', onerror='warn')


```

mcflugen:
Thanks, @kbarnhart. You are correct, `error_message` just generates an error message but doesn't raise an error.

Please feel free to add change any documentation! (In your example though, I would change `if error_condition == True:` to `if error_condition:`)

gregtucker:
This looks good to me, @mcflugen -- very useful functionality. Shall I merge?

PR #505, opened 2017-06-28
imshow grid query on mouse button press

nathanlyons:
This pull request includes a new file, `landlab.plot.event_handler` for matplotlib events. The sole function (so far) prints node coordinates when users mouse press imshow figures. This function is attached to a figure in `landlab.plot.imshow.imshow_grid_at_node`.

### DIFF
--- a/landlab/plot/event_handler.py
+++ b/landlab/plot/event_handler.py
@@ -8,12 +8,12 @@ from pprint import pprint
 
 
 def query_grid_on_button_press(event, grid):
-    """Print node information using an imshow plot.
+    """Print and returns node information using an imshow plot.
 
     This function is triggered when a mouse button is pressed on the matplotlib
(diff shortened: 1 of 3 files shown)


mcflugen:
Thanks, @nathanlyons.

Before merging, I would like folks to manually test this out on several platforms. My experience with event handling in matplotlib is that it can be finicky on some platforms. What are you working on, @nathanlyons?

nathanlyons:
@mcflugen, I am working on a mac. Thanks for pointing this out. It worked in all the environments I tried, which were:
- macOS Sierra, python 3.6
- Linux Mint 64-bit, python 2.7
- Windows 7 64-bit, python 2.6
- VNC from the mac to the linux described above

nicgaspar:
It looks good to me. I'm merging.

PR #509, opened 2017-07-05
Cmshobe/minor fixes to space

cmshobe:
@kbarnhart would you check this out for me?

In response to helpful comments by Caroline Le Bouteiller, I have:
-corrected the use of porosity in the H(t) calculation
-changed where Q_to_the_m is instantiated so it works for all discharge options
-fixed initial calculation of bedrock__elevation so that it is topo__elev - soil__depth
-updated the doctests to reflect the changes


kbarnhart:
@cmshobe I'll look once tests are done.

kbarnhart:
@cmshobe can you add docstring descriptions to the three hydrology methods used by space?

Other than that this looks good to me.

cmshobe:
@kbarnhart Ok, there's a short description of each in there now.

kbarnhart:
@cmshobe I meant docstrings under the function definition for each of the option related functions.

I have found the PEP on docstrings super useful in the past. Just in case it is useful to you here is the link:
https://www.python.org/dev/peps/pep-0257/

cmshobe:
Ohhhh ok I see-- How about I do that tomorrow under a separate PR? It's not really related to the other changes and I know some folks are waiting on the other changes to get merged. Don't worry though, I'll do a new PR tomorrow if you'll merge this one!

cmshobe:
OK simple docstrings are now in the function definitions. Let's get this pulled in so the other important changes in this PR make it into master, then I think down the line there's a way to simplify the hydrology choices.

Issue #507, opened 2017-07-03
IndexError: Out of bounds on buffer access (axis 0) -- Using FastscapeEroder component


amanaster2:
Hi!

As I was trying to use the FastscapeEroder component earlier, the following error popped up:
![image](https://user-images.githubusercontent.com/26778806/27810489-da420274-600f-11e7-8ede-e787ab6dd7da.png)

I tried uninstalling and reinstalling the developer version of Landlab (per @saisiddu 's suggestion) to no avail. Is there a known issue with this .pyx file? Could my code be causing the error? Any suggestions are appreciated!!

Thank you,
Amanda

jadams15:
Hi Amanda! Can you try to recompile? From the terminal, change directories to the top level "landlab" folder and try to run the following line of code:

`python setup.py develop`

and see if that helps!

amanaster2:
Hi Jordan! I've already recompiled, and unfortunately I haven't had any luck.

kbarnhart:
@amanaster2 I most recently worked with the .pyx file you are having issues with. Could you create a minimally complete example that reproduces the error and post it (tagging me)? I'll then look into what is causing the error.

amanaster2:
@kbarnhart : Below is the basic idea of the script I'm using. It's very possible that I'm incorrectly using FastscapeEroder as this is the first time I'm implementing it!

```
from landlab.components import KinwaveImplicitOverlandFlow, FastscapeEroder
from landlab.io import read_esri_ascii
import numpy as np

#load DEM for grid -- this is just the DEM I used
(mg, z) = read_esri_ascii(r'C:\Users\Amanda\Desktop\Lidar_amanda\Lidar_amanda\lidar_new.txt', name='topographic__elevation')

#set boundary conditions
mg.set_nodata_nodes_to_closed(z, -9999)
outlet_id = mg.core_nodes[np.argmin(mg.at_node['topographic__elevation'][mg.core_nodes])]                     
mg.set_watershed_boundary_condition_outlet_id(outlet_id, z) 

#initialize components
knwv = KinwaveImplicitOverlandFlow(mg, runoff_rate = 0.0, roughness = 0.03, depth_exp = 1.6666667)
sp = FastscapeEroder(mg, K_sp = 1.0)

#set initial conditions/constants
elapsed_time = 0
model_run_time = 7200 #seconds
storm_duration = 3600 #seconds
dt = 10. #seconds
rr = 10*2.77778E-07 #runoff rate in m/s
[...]

PR #510, opened 2017-07-05
Changing the default behavior of FlowDirectors to many (DINF and MFD)

kbarnhart:
Previously FlowDirectorDINF and FlowDirectorMFD created
“flow__link_to_receiver_node” and if the node flowed to itself, set the
value to -1. This is different than how the flow-director to one
methods work (value was set to node-id value).

This discrepancy was discovered by @amanaster2 while trying to use the
implicit kinematic wave component (which relies on flow director MFD)
along with the fastscape stream power component (which expects
flow-to-self nodes to have node id as the receiver not -1) (PR #507)

This commit should fix this issue. @amanaster2 - want to check out this branch and test that it addresses your problem.

### DIFF
--- a/landlab/components/flow_director/flow_direction_dinf.py
+++ b/landlab/components/flow_director/flow_direction_dinf.py
@@ -393,7 +393,7 @@ def flow_directions_dinf(grid,
     steepest_link[drains_to_self] = UNDEFINED_INDEX
 
     steepest_receiver = receivers[slope_sort]
-    steepest_receiver[drains_to_self] = UNDEFINED_INDEX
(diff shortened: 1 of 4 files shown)


kbarnhart:
@mcflugen would you take a look and comment/merge?

mcflugen:
@kbarnhart: Sure. If you have a look at #504.

kbarnhart:
@mcflugen yes! will do. sorry - I seem to have missed your request when I was working in Philly...

kbarnhart:
@mcflugen can I merge this?

mcflugen:
@kbarnhart Yes, looks good.

PR #511, opened 2017-07-10
Cmshobe/make erosion deposition component

cmshobe:


kbarnhart:
@cmshobe @gregtucker note about todo items before merge (other than fixing failing tests)

1) Add fraction of fines capabilities.

kbarnhart:
@mcflugen could you please re-run these tests since it looks like the last time they were tested was a time at which the major conda issue was still impacting testing.

PR #517, opened 2017-08-01
SiccarPoint/uniform precip

SiccarPoint:
This pull adds a new generator, yield_storms(), to the uniform_precip.PrecipitationDistribution class.

It is designed to mirror the necessary fields for a spatially resolved precip distribution (coming soon!) exactly, and thus makes use our novel scalar grid fields, “at_grid”. This follows a suggestion by @mcflugen on the call yesterday. Note doing this requires a grid be passed where it was not before; __init__ is modified to do this in a back-compatible way.

Method name (yield_storms), output field (‘rainfall__flux’) and generator return format (tuple of storm_t, interstorm_t) all mirror what I have in spatial_precip. Thoughts welcome on suitability of these names.

Added a new doctest for the key functionality.

### DIFF
--- a/landlab/components/uniform_precip/generate_uniform_precip.py
+++ b/landlab/components/uniform_precip/generate_uniform_precip.py
@@ -34,7 +34,7 @@ class PrecipitationDistribution(Component):
 
     Parameters
     ----------
-    grid : ModelGrid
+    grid : ModelGrid or None
         A Landlab grid (optional). If provided, storm intensities will be
         stored as a grid scalar field as the component simulates storms.
     mean_storm_duration : float
@@ -61,9 +61,9 @@ class PrecipitationDistribution(Component):
     model run time and delta t...  Say we use 1.5 for mean storm, 15 for mean
     interstorm, 0.5 for mean depth, 100 for model run time and 1 for delta t...
 
-    >>> precip = PrecipitationDistribution(mean_storm_duration = 1.5,
-    ...     mean_interstorm_duration = 15.0, mean_storm_depth = 0.5,
-    ...     total_t = 100.0, delta_t = 1.)
+    >>> precip = PrecipitationDistribution(mean_storm_duration=1.5,
+    ...     mean_interstorm_duration=15.0, mean_storm_depth=0.5,
+    ...     total_t=100.0, delta_t=1.)
     >>> for (dt, rate) in precip.yield_storm_interstorm_duration_intensity():
     ...     pass  # and so on
     """
@@ -493,9 +493,9 @@ class PrecipitationDistribution(Component):
 
         Examples
         --------
-        >>> precip = PrecipitationDistribution(mean_storm_duration = 1.5,
-        ...     mean_interstorm_duration = 15.0, mean_storm_depth = 0.5,
-        ...     total_t = 100.0, delta_t = 1.)
+        >>> precip = PrecipitationDistribution(mean_storm_duration=1.5,
+        ...     mean_interstorm_duration=15.0, mean_storm_depth=0.5,
+        ...     total_t=100.0, delta_t=1.)
         >>> round(precip.storm_duration, 2)
         2.79
         >>> round(precip.interstorm_duration, 2)
@@ -513,9 +513,9 @@ class PrecipitationDistribution(Component):
         2.45
         >>> round(precip.get_storm_intensity(), 2)
         0.88
-        >>> precip = PrecipitationDistribution(mean_storm_duration = 1.5,
-        ...     mean_interstorm_duration = 15.0, mean_storm_depth = 0.5,
-        ...     total_t = 100.0, delta_t = 1., random_seed=1)
+        >>> precip = PrecipitationDistribution(mean_storm_duration=1.5,
+        ...     mean_interstorm_duration=15.0, mean_storm_depth=0.5,
+        ...     total_t=100.0, delta_t=1., random_seed=1)
         >>> round(precip.storm_duration, 2) # diff't vals with diff't seed
         0.22
         >>> round(precip.interstorm_duration, 2)
--- a/landlab/components/uniform_precip/generate_uniform_precip.py
+++ b/landlab/components/uniform_precip/generate_uniform_precip.py
@@ -144,7 +144,7 @@ class PrecipitationDistribution(Component):
 
         # Test if we got a grid. If we did, then assign it to _grid, and we
         # are able to use the at_grid field. If not, that's cool too.
-        if grid is not none:
+        if grid is not None:
             assert isinstance(grid, ModelGrid)  # must be a grid
             self._grid = grid
 
--- a/landlab/components/uniform_precip/generate_uniform_precip.py
+++ b/landlab/components/uniform_precip/generate_uniform_precip.py
@@ -84,7 +84,7 @@ class PrecipitationDistribution(Component):
         'rainfall__flux':
             'Depth of water delivered per unit time in each storm', }
 
-    def __init__(self, *args,
+    def __init__(self, grid=None,
                  mean_storm_duration=0.0, mean_interstorm_duration=0.0,
                  mean_storm_depth=0.0, total_t=0.0, delta_t=None,
                  random_seed=0, **kwds):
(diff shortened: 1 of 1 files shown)


SiccarPoint:
Bump ( @kbarnhart ?). There was a Py2-Py3 compatibility issue regarding the generator treatment that I've now squashed.

I'm still seeing a fail on my local machine related to an attempted import of __init__.pyc within hybrid_alluvium (only in Py3), but I suspect that's an old file on my machine doing that and that these main tests will pass.

kbarnhart:
@SiccarPoint I think you are right about the local conflict. There is no longer a `HybridAlluvium` component (or a hybrid alluvium) folder. It is now called `Space`.

landlab/components/uniform_precip/generate_uniform_precip.py
```diff
@@ -13,7 +13,8 @@
 
 import random
 import numpy as np
-from landlab import Component
+from landlab import Component, ModelGrid
+from six import next
```
kbarnhart: A comment less about your changes and more about the documentation. It states in two places that the units are in mm/hr. Shouldn't that be changed to L/T, where T has the same units as the time step and L has the same units as ... whatever is using the generated precipitation (e.g. if precipitation is being used with cell area to calculate an effective, Q, then precipitation should have same L units as grid cells).

landlab/components/uniform_precip/generate_uniform_precip.py
```diff
@@ -26,13 +27,17 @@ class PrecipitationDistribution(Component):
...
                                   mean_interstorm_duration=0.0,
                                   mean_storm_depth=0.0, total_t=0.0,
                                   delta_t=0.0, random_seed=0)
 
     Parameters
     ----------
+    grid : ModelGrid or None
+        A Landlab grid (optional). If provided, storm intensities will be
+        stored as a grid scalar field as the component simulates storms.
     mean_storm_duration : float
```
kbarnhart: An anticipated forward compatibility question.

If in the future we make one of a few variants of spatially variable precipitation generator (e.g. the one @SiccarPoint discussed on the LL call, someone else talked about an orographic precipitation component..., I was just talking to someone at NCED about a very simple random number generator based one that I can't remember if someone else is thinking about making) then should we make sure that the name for this variable is appropriate for both OR give this variable a name that is unique to spatially uniform precipitation. 

That is, in the future, would a component that used rainfall flux look for `rainfall__flux` at grid and at node, and raise an error if both existed, and know that if it was provided at grid, that meant it was uniform, and if at node, it is spatially variable.

landlab/components/uniform_precip/generate_uniform_precip.py
```diff
@@ -26,13 +27,17 @@ class PrecipitationDistribution(Component):
...
+                                  mean_storm_duration=0.0,
                                   mean_interstorm_duration=0.0,
                                   mean_storm_depth=0.0, total_t=0.0,
                                   delta_t=0.0, random_seed=0)
 
     Parameters
     ----------
+    grid : ModelGrid or None
+        A Landlab grid (optional). If provided, storm intensities will be
+        stored as a grid scalar field as the component simulates storms.
```
kbarnhart: List the name of the default field here?

landlab/components/uniform_precip/generate_uniform_precip.py
```diff
@@ -361,10 +432,109 @@ def yield_storm_interstorm_duration_intensity(self,
...
+        with interstorms represented as intervals with rf_rate = 0, it yields:
+
+            (storm_duration, interstorm_duration)
+
+        When each tuple pair is yielded, the grid scalar field 'rainfall__flux'
+        is updated with the rainfall rate occuring during storm_duration.
+
+        This generator method is designed for direct equivalence with the
+        spatially resolved generators found elsewhere in Landlab.
+
```
kbarnhart: Give example of a spatially resolved generator here?

landlab/components/uniform_precip/generate_uniform_precip.py
```diff
@@ -361,10 +432,109 @@ def yield_storm_interstorm_duration_intensity(self,
...
+        ------
+        tuple of float
+            (storm_duration, interstorm_duration)
+
+        Notes
+        -----
+        One recommended procedure is to instantiate the generator, then call
+        instance.next() (in Python 2) or next(instance) (in Python 3)
+        repeatedly to get the sequence.
+
```
kbarnhart: Since you can't give a complete example below b/c of testing 2/3 compatablitiy, would you write more explicitly what is meant by this so that users not super familiar with using iterators can have a foothold.  

E.g. 

Instead of writing.

```python
>>> for (storm_dt, interstorm_dt) in precip.yield_storms():
 +        ...     storm_dts.append(storm_dt)
 +        ...     interstorm_dts.append(interstorm_dt)
 +        ...     intensities.append(mg.at_grid['rainfall__flux'])
```
write
```python
??
```

landlab/components/uniform_precip/generate_uniform_precip.py
SiccarPoint: Can you point out an example of a component already using this kind of notation? I've not seem one that doesn't commit to, e.g., `m` as the length scale (even though it might not have to be).

landlab/components/uniform_precip/generate_uniform_precip.py
SiccarPoint: Done

landlab/components/uniform_precip/generate_uniform_precip.py
SiccarPoint: Yes, I think so...?

SiccarPoint: Might also be a possibility to *force* this component to output `at_node`, with the values all the same? Seems like a waste of memory, but 1. the option to not have it do this is still there, and 2. this would be specifically compatible with spatially variable ones.

Let me know what you think? Happy to change this.

landlab/components/uniform_precip/generate_uniform_precip.py
SiccarPoint: Doesn't exist yet, so can't yet. Will add once there is one.

landlab/components/uniform_precip/generate_uniform_precip.py
SiccarPoint: I just got `??` as your recommendation there...? But I think I know what you mean (done).

kbarnhart: I wrote `??` since I don't know what to put... though later on, I saw some nice examples in the code...

kbarnhart: But the comment still stands since the intent of the comment is that a user wouldn't need to dig through the code for the example.

SiccarPoint: (I used a six import in the test to let it happen)

SiccarPoint: I think I did something sensible.

landlab/components/uniform_precip/generate_uniform_precip.py
SiccarPoint: @kbarnhart bump? Trying not to let this fall off my to-do list...

kbarnhart: @SiccarPoint Sorry, was backpacking (and now en route to the Boundary Waters) so my response time is somewhat delayed. 

You are right, I can't think of a component where this is implemented. I''ve looked around and found mostly units in meters and seconds. 

I suspect that units were a large topic of discussion before I joined the team, but maybe its worth revisiting in a future call since as far as I can tell landlab is agnostic, but most components list units in meters, seconds, etc.

kbarnhart:
Other than a import glitch on one of the appveyor builds, this all looks good.
I won't be able to comment in the next week or so, so feel free to merge without further comments/review from me.

landlab/components/uniform_precip/generate_uniform_precip.py
SiccarPoint: Sorry! Hadn't realised you were out of the office. I think the approach we've had until now is to assume standard units are m and yr. In lots of cases, it doesn't really matter, as long as the user checks all the unit conversions. _However,_ there definitely are components where this isn't the case, and things need to be geared correctly. This tends to happen anywhere where a component takes both a rapid flux as an input and produces a slow change as an output. In these cases, typically discharge [m**3/s], but dz/dt [m/y]. I think that because of these cases, we've basically just asserted `m` and `y` everywhere, to be on the safe side. The most obvious cases where the units could be different I think are generally flagged in the description.

Let's talk this over on a call, as you say.

kbarnhart: No worries... (we are on a multi-part holiday... first the eclipse... now canoeing!)

I'll put it on the call to-do list.

PR #520, opened 2017-08-02
Add event layers

mcflugen:
This pull request adds stratigraphic layers to landlab grids. Layers can be accessed through the new `layers` attribute of any landlab grid. I've created an [accompanying tutorial](https://github.com/landlab/tutorials/blob/mcflugen/add-eventlayers-tutorial/layers/eventlayers.ipynb) in a Jupyter Notebook in our tutorials repository.

This pull request only add the implementation of "event layers". That is, every event gets a single layer - there is no averaging of layers (those would be "binned layers" and will come as a separate pull request).


landlab/layers/eventlayers.py
```diff
@@ -0,0 +1,503 @@
...
+import numpy as np
+
+
+def _deposit_or_erode(layers, n_layers, dz):
+    """Add a new layer to the top of a stack.
+
+    Parameters
+    ----------
+    layers : ndarray of shape `(M, N)`
+        Array of layer thicknesses.
```
gregtucker: Would be useful to say in comment what M and N index (presumably one is nodes, one is layers?)

landlab/layers/eventlayers.py
```diff
@@ -0,0 +1,503 @@
...
+
+    if array.ndim > 0 and len(array) != nstacks:
+        values_per_stack = array.shape
+    else:
+        values_per_stack = array.shape[1:]
+        
+    return np.empty((nlayers, nstacks) + values_per_stack , dtype=array.dtype)
+    
+
+class EventLayersMixIn(object):
```
gregtucker: First time I've heard of MixIn's. Cool!

SiccarPoint:
Thought on naming - shouldn't all the layer properties stick with our agreed LL conventions? e.g., `layers.nlayers` -> `layers.number_of_layers`, `layers.nstacks` -> `layers.number_of_stacks`, etc.

This is awesome though.

SiccarPoint:
Also, a question on functionality. Are the properties in any way bound to layers that aren't issued with keywords? e.g., what is the output if I do this:

```
>>> grid.layers.add(3.)
>>> grid.layers.add(4., age=10.)
>>> grid.layers.add(5., d50=20.)
>>> grid.nlayers
```
[what is this output? I assume 3?]
```
>>> grid.layers['age']
```
??? - is this:
```
array([[ 0.,  0.,  0.],
       [10., 10., 10.],
       [ 0.,  0.,  0.]])
```
or just
``` 
array([10., 10., 10.])
```
If the latter, this seems like a recipe for indexing disaster. Perhaps there needs to be a mechanism to prevent the adding of layer properties piecemeal, or at least, to prevent unintuitive things happening, e.g.,
```
>>> grid.layers.add(3.)
>>> grid.layers.add(4., age=10.)
LayerError: Property 'age' does not exist in one or more lower layers that have already been created. Force the creation of the property 'age' set to zero in those layers using the keyword `override_properties=True`.
```
Or similar?

While I'm thinking about it, I assume it's trivial to modify the values of properties (or layer thicknesses?) in place, right? ->
```
>>> grid.layers.add((1., 1., 1., 1.))
>>> grid.layers.thickness[0, 2] = 0.
>>> grid.layers.thickness
array([[ 1., 1., 1.],
       [ 1., 1., 1.],
       [ 0., 1., 1.],
       [ 1., 1., 1.]])
```

mcflugen:
Thanks @SiccarPoint. These are good questions. I think the answer to your first question is that the attributes should be fixed when the first layer is added and after that if one of those attributes is not specified, an error should be raised.

I actually hadn't thought about changing layer thicknesses directly. I think that should work. Is there some reason that we wouldn't want users setting layer like that?

Issue #522, opened 2017-08-04
Component ErosionDeposition gives RuntimeWarnings if n<0 and flooded nodes are present 


kbarnhart:
This is because flooded nodes have slopes of < 0 and the erosion efficiency is calculated by raising the slope to n. 

@gregtucker I'll fix this today. Look out for a PR

@cmshobe This may be an issue for SPACE

PR #523, opened 2017-08-04
Fixing ErosionDeposition so that it can handle negative slopes with n<1

kbarnhart:
Addresses #522 

@gregtucker @cmshobe let me know if you have comments.

### DIFF
--- a/landlab/components/erosion_deposition/erosion_deposition.py
+++ b/landlab/components/erosion_deposition/erosion_deposition.py
@@ -215,6 +215,7 @@ class ErosionDeposition(Component):
         # power "m", and deposition rate
         self.qs_in = np.zeros(grid.number_of_nodes)
         self.Q_to_the_m = np.zeros(grid.number_of_nodes)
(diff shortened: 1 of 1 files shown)


gregtucker:
This looks good to me. One potential issue though: setting flooded nodes to S < 0 could have the unintended side effect of allowing erosion at nodes that are flooded but have S > 0. In theory, the solver should prevent S < 0 at non-flooded nodes, though the existence of a minimum dt could allow S < 0 to happen occasionally. Is there a way to mark "flooded" nodes as the set of those identified as flooded by the lake mapper as well as any additional ones with S < 0?

kbarnhart:
Good point @gregtucker . I'll fix that.

kbarnhart:
@gregtucker Now this should maintain the original flooded nodes and add any "new" flooded nodes with slopes of less than zero.

gregtucker:
Cool, thanks. I'll merge once tests finish.

kbarnhart:
Thanks!

Issue #521, opened 2017-08-02
function save_grid() fails when grid attribute _diag_fixed_links has been created


kbarnhart:
I'm working on pickling a very large and complicated grid in order to give a model built with Landlab restart capabilities. 

When I try to pickle the model grid with the function ``io.native_landlab.save_grid`` I get the following error:

```python
AttributeError: Can't pickle local object 'override_array_setitem_and_reset.__call__.<locals>._wrapped.<locals>.array'
```

Some sleuthing (using the function at the bottom, which I copied and modified slightly from [here](https://stackoverflow.com/questions/569754/how-to-tell-for-which-object-attribute-pickle-fails)) indicates that the issue is with the grid attribute ```_diag_fixed_links```.

To test this, I've made two minimally complete examples. In the first, I save a grid with elevation only. It succeeds. 

```python
from landlab import RasterModelGrid
from landlab.components import FlowAccumulator

from landlab.io.native_landlab import save_grid
mg = RasterModelGrid(4,5,2.)
z = mg.add_zeros('node', 'topographic__elevation')
z += mg.node_x.copy()

save_grid(mg, 'testsavedgrid.grid', clobber=True)
```

The second is identical except that I use FlowDirectorD8 (which is what I use in my complicated use-case). It fails with the same error message given above. 

```python
from landlab import RasterModelGrid
from landlab.components import FlowAccumulator

from landlab.io.native_landlab import save_grid
mg = RasterModelGrid(4,5,2.)
z = mg.add_zeros('node', 'topographic__elevation')
z += mg.node_x.copy()
fa = FlowAccumulator(mg, flow_director='D8')
fa.run_one_step()
save_grid(mg, 'testsavedgrid.grid', clobber=True)
```

I've started to look into the [documentation on what can and can't be pickled](https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled) and I **think** that this error may be coming because some aspect of the ```____dict___``` is un-pickle-able. I'm a bit confused, however, since the documentation indicates that this should raise a ```PicklingError``` exception and I'm getting an ```AttributeError``` exception. 

@mcflugen @SiccarPoint @gregtucker any ideas?

Function used to figure out which part of the grid was failing. 
```python
def get_pickling_errors(obj,seen=None):
    if seen == None:
        seen = []
    try:
        state = obj.__dict__
    except AttributeError:
        return
    if state == None:
        return
    if isinstance(state,tuple):
        if not isinstance(state[0],dict):
[...]

PR #525, opened 2017-08-04
Barnhark/get set state methods for grid

kbarnhart:
This addresses PR #521 

I've tried to identify only the necessary elements of a `RasterModelGrid` that are needed in order to recreate the `__dict__` upon un-pickleing. 

I've made new tests (and these are functionally used by the current `save_grid` and `load_grid` methods as they serve to provide instructions to `pickle` regarding how to pickle a grid. 

@SiccarPoint @mcflugen @gregtucker take a look?

If someone can see a way to make this more parsimonious, do tell. 

In the complicated application that motivated this improvement, these methods save a grid of size (415, 438),  181770 total nodes, 90088 core nodes, with 17 at node fields at 3 at link fields in 31.1 MB.

### DIFF
--- a/landlab/grid/base.py
+++ b/landlab/grid/base.py
@@ -3281,17 +3281,19 @@ class ModelGrid(ModelDataFieldsMixIn):
                           INACTIVE_LINK)
         inactive_links[self.link_dirs_at_node == 0] = False
         self._active_link_dirs_at_node[inactive_links] = 0
-
+        
         try:
             if self.diagonal_list_created:
                 self.diagonal_list_created = False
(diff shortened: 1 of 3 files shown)


kbarnhart:
for comparison, saving 1 at-node array in a netcdf for this grid costs 1.6 MB, 

so just saving the at-node arrays in a netcdf would cost 25 MB.

mcflugen:
@kbarnhart We probably aren't compressing the data when we write our NetCDF files - but we should. If we did, I imagine we could get close to 1.6 MB.

Issue #524, opened 2017-08-04
at_grid fields implemented as true scalars instead of as arrays with length 


kbarnhart:
Thus the following happens. 

```python
field = mg.add_field['grid', 'myfield', 0.]
# put a new value in the scalar field
field[0] = 1. 
INDEX ERROR
```

Similarly, the following yields an `INDEX ERROR`

```python
mg.add_field('grid', 'my_field', np.array([0., ]))
```

And @SiccarPoint points out that this is not ideal because it means the following is not possible. 

```python
[in init]
if one_case:
    self._var = mg.at_grid['my_field']  # self._var is used all over the place and frequently updated
else:
    self._var = np.array([0., ])  # some value not tethered to the grid
[in run_one_step]
self._var[0] = new_val
```

@mcflugen points out that since this happens.

```python
>>> grid.add_field('z', np.array([1,2,3]), at='grid')
array([1, 2, 3])
```
he would expect that this *should* happen,
```python
>>> grid.add_field('z', np.array([1]), at='grid')
array([1])
```
not,
```python
>>> grid.add_field('z', np.array([1]), at='grid')
array(1)
```
So @mcflugen concludes that this its a bug and not a feature.

I'll assign myself.

PR #526, opened 2017-08-04
Fix for scalar data fields and a new test

kbarnhart:
This PR starts to address Issue #524. This does not pass for me but gives strange errors.

e.g.
```python
AssertionError: array([ 1.,  2.]) is not array([ 1.,  2.])
```
This doesn’t make sense to me, since the arrays appear equivalent.


mcflugen:
@kbarnhart I need a little more to go on to debug what you have so far. I guess it's because an array was copied somewhere and so, even though the two arrays have the same values, that aren't the same array.

I'm not sure if this was a complete fix but I think changing the following line in `scalar_data_fields.py` from
```python
    if need_to_reshape_array(value_array, self.size):
```
to
```python
    if value_array.ndim > 1 and need_to_reshape_array(value_array, self.size):
```
should fix the problem of arrays of length 1 being converted to scalars - at least for most cases, it may not be a total fix and may cause other problems. Also please note that this is from memory so it may not be exactly correct but the flavor should be right.

kbarnhart:
Thanks @mcflugen . I improved the fix slightly based on your recommendations, but I still get some errors. I've pasted the entire error part of the land lab tests. 


```python
FAIL: Test adding an array as a grid field.
  File "/anaconda/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/barnhark/projects/landlab/landlab/field/tests/test_grouped_fields.py", line 204, in test_grid_field_as_array
    assert_is(val, fields.at_grid['const'])
AssertionError: array([ 1.,  2.]) is not array([ 1.,  2.])
    """Fail immediately, with the given message."""
>>  raise self.failureException('array([ 1.,  2.]) is not array([ 1.,  2.])')


FAIL: Test add field at grid maintains length property.
  File "/anaconda/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/barnhark/projects/landlab/landlab/grid/tests/test_raster_grid/test_fields.py", line 66, in test_at_grid_length
    assert_array_equal(np.array([0]).shape, grid.at_grid['myfield'].shape)
  File "/anaconda/lib/python3.6/site-packages/numpy/testing/utils.py", line 813, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "/anaconda/lib/python3.6/site-packages/numpy/testing/utils.py", line 694, in assert_array_compare
    raise AssertionError(msg)
AssertionError:
Arrays are not equal

(shapes (1,), (0,) mismatch)
 x: array([1])
 y: array([], dtype=float64)
>>  raise AssertionError('\nArrays are not equal\n\n(shapes (1,), (0,) mismatch)\n x: array([1])\n y: array([], dtype=float64)')


Ran 1451 tests in 22.968s

FAILED (failures=2)
```

kbarnhart:
@mcflugen I'll keep thinking about how to fix this. Going to bike home now, though.

PR #529, opened 2017-08-09
Gt try dyn ts space

gregtucker:
This PR adds an adaptive time-step solver to the SPACE component, and makes a few changes for speed (such as pre-allocating arrays in __init__ rather than creating them repeatedly under run_one_step).


kbarnhart:
@gregtucker assuming this passes tests, is this ready for use?

kbarnhart:
I assume yes.

landlab/components/space/space.py
```diff
@@ -295,14 +335,29 @@ def __init__(self, grid, K_sed=None, K_br=None, F_f=None,
...
+            self.run_one_step = self.run_one_step_basic
+        elif solver == 'celerity':
+            self.run_one_step = self.run_with_simple_time_step_adjuster
+        elif solver == 'adaptive':
+            self.run_one_step = self.run_with_adaptive_time_step_solver
+            self.time_to_flat = np.zeros(grid.number_of_nodes)
+        else:
+            raise ValueError("Parameter 'solver' must be one of: "
+                             + "'basic', 'adaptive', 'celerity'")
+
```
gregtucker: Yes, you're right, I forgot to update that. I decided the celerity solver wasn't ready for prime time, and would in practice be superceded by adaptive, so took it out.

gregtucker: Otherwise, it should be ready to use, with the caveat that it hasn't been tested against analytical solutions yet. Charlie has agreed to do that today or tomorrow.

kbarnhart: Sounds good. Looks like everything passes, so I'll wait for the updates to documentation and then merge.

landlab/components/space/space.py
```diff
@@ -327,6 +381,7 @@ def simple_stream_power(self):
                                 'not nnodes long!')  
                 self.Q_to_the_m[:] = np.power(self.drainage_area, self.m_sp)
             elif self.discharge_method == 'discharge_field':
+                print('how did we get here???')
                 if self.discharge_field is not None:
```
kbarnhart: I might also get rid of this comment line.

PR #545, opened 2017-09-21
watershed utility

nathanlyons:
This utility gets watersheds including those within grid boundaries. Functions identify the nodes that have flow received by a user designated outlet node. This relies on the field 'flow__receiver_node', so flow must be routed before the utility is used.


gregtucker:
Nice utility to have, thanks Nathan

margauxmouchene:
Great utility! And neat implementation!

A few comments:

- Line 32: Doc test (Examples) for get_watershed_array:

    `from landlab.utils.watershed import get_watershed_ids`

    should be:

    `from landlab.utils.watershed import get_watershed_arrays`

- Unit test (_test_watershed.py_) should go through every line of your code. At the moment I think it only tests _get_watershed_id_ (which also calls _get_watershed_array_) in a simple case, it should also test the if/else loops you have in _get_watershed_array_.

- I am going to ask help from @gregtucker or @mcflugen on this one: I think you have to modify the `_init_.py` file in the landlab/utils folder to have your utility import correctly within Landlab. Can Greg or Eric explain/confirm?

mcflugen:
@margauxmouchene, @nathanlyons Yes, it might be nice to add `from .watershed import get_watershed_ids` to `__init__.py`. The users will be able to do `from landlab.utils import get_watershed_ids`.

@nathanlyons, a couple other things:

* Is there a reason `get_watershed_ids` returns a `list` rather than a numpy array? If not, I suggest just returning the numpy array.
* What do you think about renaming `get_watershed_ids` to `get_watershed_nodes`? Since all grid elements have ids, I think it's clearer to say "nodes".
* I wonder if there is also a better name for `get_watershed_array`. Maybe something along the lines of `get_watershed_mask` (or `get_watershed_node_mask`)?
* Is `get_watershed_array` slow? Given the loop of the grid nodes, it seems like it might be. If so, we should try to cythonize this in another pull request.
* The one-line function descriptions should be in the active voice ("get" instead of "gets").

nathanlyons:
@margauxmouchene, thanks for the review. The import statements has been corrected, the test covers all except 1 line (the field error line), and the utility is import in util's init.

nathanlyons:
@mcflugen, thanks for the comments. In the future, I will aim to cythonize this in the near future. All other comments addressed.

nathanlyons:
@margauxmouchene The major change since your last review is the new function, get_watershed_masks_with_area_threshold.

margauxmouchene:
Everything seems fine to me, and the tests now all pass so I think we can merge it!

PR #546, opened 2017-09-21
Transport-length-model hillslope diffuser

margauxmouchene:
Hillslope diffusion component inspired by transport length-model equations by Carretier et al (2016) and Davy & Lague (2009)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -105,6 +105,7 @@ Hillslope geomorphology
 
    landlab.components.diffusion
    landlab.components.nonlinear_diffusion
+   landlab.components.transport_length_diffusion
 
 Fluvial geomorphology
 ---------------------
--- a/docs/landlab.components.transport_length_diffusion.rst.txt
+++ b/docs/landlab.components.transport_length_diffusion.rst.txt
@@ -1,0 +1,7 @@
+TransportLengthHillslopeDiffuser: Model hillslope diffusion using a transport-length equation
+---------------------------------------------------------------------------------------------
+
+.. automodule:: landlab.components.transport_length_diffusion
+    :members:
+    :undoc-members:
+    :show-inheritance:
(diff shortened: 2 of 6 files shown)


gregtucker:
Neat idea to apply this concept to hillslopes. If I understand right, the process is "local" when S << Sc, and becomes "nonlocal" as S => Sc.

margauxmouchene:
Yes, that's exactly right! When S<<Sc, most of the sediment entering a cell is deposited there (similar to pure diffusion), the sediment flux coming out of this cell will be "local" (does not include sediments from above). When S~Sc, the sediments 'bypass' the cell (or cells) and are deposited further downslope, the flux includes not only sediments detached locally but also sediments eroded further upslope (mimics mass wasting).

nicgaspar:
Looks good. I like the way you devised the unit tests.

Two things: I think you might want to change your field names to be more in line with our "standard", which sort-of, roughly follows CSDMS naming standards, and I think there should be plenty of text on CSDMS standard names. Standards might also be described in the Hobley et al ESURF paper. Here is a list (maybe out of date) of field names already in use https://github.com/landlab/landlab/wiki/Standard-names . The first word should be the thing that is being described, then two underscores, and then the attribute of the thing that is being described. E.g. soil__flux is one that is in use already. So maybe you want soil__flux_in? I think we also have a soil__unit_flux. You might need to dig a bit. I'm not sure that website is up to date.

The other question is - do you have stability issues? Do you need a time-step calculator? Or is that not an issue?

margauxmouchene:
I edited the names of some variables to match standard names  (e.g. `flux_in` changed to `soil__flux_in`).

I deleted `**kwds` from input because it was not used.

I implemented a test for timestep-related stability, it raises an error if the slopes are reversed by too much erosion/deposition during a too long time step. The diffusion equations I used do not allow for an adaptative time step to be calculated, so this seems to be the best solution.

gregtucker:
Increasing coverage: nice!

PR #551, opened 2017-09-23
incorporate porosity (phi) in SPACE's adaptive solver option

gregtucker:
Porosity factor 1 / (1-phi) had been included in SPACE's default solver, but not in the adaptive solver. Now corrected.


gregtucker:
Thanks, will merge when tests finish

Issue #560, opened 2017-10-02
Landlab does not store/write out out lower left corner and projection information


kbarnhart:
Projected data sources need lower left corner specified as well as projection information. 

Landlab model grids currently use an origin of (0,0) and do not store projection information. 

This should be addressed for both Esri Ascii and NetCDF formats.

PR #561, opened 2017-10-02
Barnhark/projection and llc storage on grid

kbarnhart:
This is a set of preliminary commits that begin to address Issue #560 

So far it reads projection information from a .proj file (if it exists), and writes it out when writing to esri ascii. 

It uses the lower left corner provided (adjusted based on the presence of a halo) as the origin of the landlab grid.

Still to do:

1) Make similar functionality and tests for netCDF reader/writer
2) Create the ability of the grid to check for projection incompatibilities.


kbarnhart:
I'm tagging @saisiddu @mcflugen @gregtucker @erkanistan @nicgaspar @ChristinaB since we discussed this morning. I'll re-post on this thread when the todo list items are finished and this branch is ready for discussion.

kbarnhart:
It appears as if all CI tests error out on two errors that I'm at a bit of a loss to deal with (since I'm not getting them). 

My inital guess is that the file "4_x_3.proj" is somehow missing.

@mcflugen any ideas?


```
ERROR: landlab.io.tests.test_read_esri_ascii.test_4x3_read_file_name
  File "/home/travis/miniconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/lib/python2.7/site-packages/landlab/io/tests/test_read_esri_ascii.py", line 69, in test_4x3_read_file_name
    assert_equal(grid.projection.strip(), '+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs')
AttributeError: 'NoneType' object has no attribute 'strip'

ERROR: landlab.io.tests.test_write_esri_ascii.test_4x3_write_projection
  File "/home/travis/miniconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/lib/python2.7/site-packages/landlab/io/tests/test_write_esri_ascii.py", line 124, in test_4x3_write_projection
    with open(os.path.join(_TEST_DATA_DIR, 'copy_4_x_3.proj'), 'r') as f:
IOError: [Errno 2] No such file or directory: '/home/travis/miniconda/lib/python2.7/site-packages/landlab/io/tests/data/copy_4_x_3.proj'
```

kbarnhart:
A follow-up comment regarding spatial information in NetCDF datasets. 

It appears that there the standard and favored way for coordinate information to be stored on a NetCDF is through a grid mapping variable (called `grid_mapping`):

```
A grid mapping variable is required if, in addition, it is desired to describe 
the mapping between the given coordinate variables and the true latitude 
and longitude coordinates, or to describe the figure of the Earth used to 
define the latitude and longitude coordinates, or to describe another
coordinate reference system definition used by some coordinates or auxiliary 
coordinates. 
```
From : http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#grid-mappings-and-projections

So I will plan to make the NetCDF reading and writing tools look for this variable, and if it exists, reproduce it upon writing out. 

@mcflugen @saisiddu @ChristinaB

Have any of you dealt with NetCDFs that have the reference system specified in a different way?

kbarnhart:
And a further follow-up comment regarding internal testing for projection inconsistency. 

Take as an example the present the way a grid and field would be read in from an esri ascii:

```python
(grid1, field1) = read_esri_ascii('file_name1.asc')
```

if a second grid (which for the moment, we'll assume has the unlikely but possible quality of the same dimensions but a different projection were read in, it would be done as follows:

```python
(grid2, field2) = read_esri_ascii('file_name2.asc')
```

If one then wanted to attach the `field2` to `grid1`, you would probably do the following:

```python
grid.add_field('node', 'field2', field2)
```

With a netcdf, the situation is a bit different. Fields are not returned as part of the reading function. So the equivalent of the line above is:
```python
grid2 = read_netcdf('file_name2.nc') 
field2 = grid.at_node['field2']
```

However, `field2` is just a `numpy` [ndarray](https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.ndarray.html), and so it doesn't really have the ability to communicate that it is associated with a particular projection. 

Some options:

1) Do nothing. Users must know that if they read in multiple grids, they all must be co-located. Communicating this may be improved by our IO documentation that reminds people that landlab does not do on-the-fly-projection. 

2) Identify a way to attach the projection information to the field (I have not yet come up with a way that I like.)

mcflugen:
@kbarnhart I think your tests are failing because you have not installed your test data files. Either you need to add them to `setup.py` as `package_data` or create them as part of your tests. If the files are rather small, which it looks like yours are, the second option might be best. You could do this as a [package-level fixture](https://nose.readthedocs.io/en/latest/writing_tests.html#fixtures).

kbarnhart:
@mcflugen , thanks!
I've added it to `package_data`, and re-submitted for testing.

kbarnhart:
@mcflugen looks like that `package_data` change fixed the issue. Thanks! 

I'm going to hold off on making further changes (and implementing the `grid_mapping` variable for the NetCDF case until we discuss this at our next call .

kbarnhart:
@SiccarPoint This is the PR/discussion about projections

@saisiddu - let me know what you think about options for testing projection compatability

saisiddu:
@kbarnhart:
- I have not dealt with contradicting projections at anytime. 
- I think it is nice to have capability to read and write projections.
- I am curious to see how we could attach projection data to a field? I am wondering about how a user suspect such capability and edit code accordingly? Will this not mean editing all our components? And doesn't all future components be aware of this while writing a component?
   - would we provide with a field' projection converter?
- Nice work Katy!

kbarnhart:
Thanks for the comment @saisiddu

kbarnhart:
Doing the projection part of this problem turned out to be really hard, so I've scaled back to just modifying the read/write esrii ascii to deal with lower left corners better (aka actually paying attention to them). 

I think these tests will pass, and then perhaps someone @saisiddu @mcflugen @SiccarPoint can merge?

Also, @ChristinaB, @jphuong I think someone told you this was coming down the pipeline on the last landlab call (which I missed to do some fieldwork). 

I hope this addresses your needs.

kbarnhart:
@frengers FYI this improvement has now occurred in Landlab. We discussed it many months (years) ago.

Issue #363, opened 2016-07-14
mg.nodes_at_link


SiccarPoint:
I really want a `nodes_at_link` property. At the moment we only have `node_at_link_head` and `node_at_link_tail`, in isolation. We can retain these, but we'd have to make it so that `node_at_link_tail = nodes_at_link[:, 0]` and `node_at_link_head = nodes_at_link[:, 1]` under the hood.

SiccarPoint:
@mcflugen - Greg likes this idea, but thought you might already be including it? If not, I'll do it now.

mcflugen:
I've totally already added this is my branch. I actually thought we would already have had this. I'm sorry I haven't had time to get all this stuff back into the master branch.

SiccarPoint:
Cool, I'll hold off.

SiccarPoint:
Bump?

SiccarPoint:
Bump!?!

PR #566, opened 2017-10-10
Add nodes_at_link

mcflugen:
This pull request adds a `nodes_at_link` property to `ModelGrid` (issued #363). `nodes_at_link` is a index array of shape `(n_links, 2)` where the first column is the node at each link tail and the second column the node at the link head. `node_at_link_tail` and `node_at_link_head` now return views of the `nodes_at_link` matrix.

### DIFF
--- a/landlab/grid/base.py
+++ b/landlab/grid/base.py
@@ -1050,6 +1050,25 @@ class ModelGrid(ModelDataFieldsMixIn):
         return self._core_cells
 
     @property
+    def nodes_at_link(self):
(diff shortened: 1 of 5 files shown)


mcflugen:
@SiccarPoint, does this look good to you?

SiccarPoint:
This looks great, but I'm not 100% sure there's a good unit test/doctest for this. It might be worth adding one. Your call. Other than that, merge away.

mcflugen:
@SiccarPoint thanks! I added some unit tests.

PR #572, opened 2017-10-17
OverlandFlow component TypeError fix

amanaster2:
Fixed TypeError that pops up when running OverlandFlow component.


amanaster2:
@mcflugen I'm unsure if I did this correctly -- please let me know if I need to do something differently next time.

mcflugen:
@amanaster2 Thanks! I think this should be fine. I'll wait for the tests to complete before merging - just in case.

mcflugen:
@amanaster2: Will you need a new release that contains this fix or is just having this on the master branch good enough?

amanaster2:
@mcflugen Would having it on the master branch be good enough if people use HydroShare to run Landlab? Because I believe some people will be using HydroShare to run Landlab, and I'm unsure if HydroShare pulls from the master branch or not.

mcflugen:
@amanaster2: I'm not sure how landlab is installed on HydroShare. I believe they build landlab from the master branch but I'm not sure. This would be a question for @saisiddu. Regardless if they use the development version from GitHub or a release, you will have to work with them to update the version they are using.

amanaster2:
@mcflugen Sure thing -- that shouldn't be a problem.

mcflugen:
@amanaster2: Great. Please let me know if you need a new release as it will take a bit of time to create and deploy it.

PR #574, opened 2017-10-21
new kw arg FastscapeEroder: discharge_name

gregtucker:
This is a simple update to FastscapeEroder that allows it to take a discharge field instead of using drainage area. The implementation is simple: the component stores a string discharge_name, which defaults to 'drainage_area'. In the erode() method, the hard-coded name 'drainage_area' is now replaced by variable discharge_name. If the user passes something else for discharge_name, such as 'surface_water__discharge', this is used instead of the default 'drainage_area' field. If it doesn't exist, the user gets a key error.

Note that testing for existence of the passed field in __init__ would make some of our tests fail, because sometimes this component is instantiated before the requested field (such as 'drainage_area') is created. This is the reason for storing the name rather than the field itself.

### DIFF
--- a/landlab/components/stream_power/__init__.py
+++ b/landlab/components/stream_power/__init__.py
@@ -4,4 +4,5 @@ from .stream_power_smooth_threshold import StreamPowerSmoothThresholdEroder
 from .sed_flux_dep_incision import SedDepEroder
 
 
-__all__ = ['StreamPowerEroder', 'FastscapeEroder', 'SedDepEroder', 'StreamPowerSmoothThresholdEroder']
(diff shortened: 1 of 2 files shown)


gregtucker:
While we're at it, also make analysis easier by adding the landlab.plot function analyze_channel_network_and_plot to plot's __init__, so it can be imported via landlab.plot (rather than needing to know the specific file in the module)

mcflugen:
@gregtucker It looks like the docs need updating for the new keyword.

gregtucker:
Good catch, doc added for new parameter.

PR #580, opened 2017-10-27
Add citations

mcflugen:
This pull request adds citations to landlab components (and landlab itself). The idea is that components should have a `_cite_as` attribute, which is a string (or a list of strings), that gives a citation that users of the component can use in publications. All instantiated components are added to `landlab.registry` from which citations can be gathered.

For example,
```
>>> import landlab
>>> print landlab.registry.format_citations()
    @article{hobley2017creative,
    title={Creative computing with Landlab: an open-source toolkit
        for building, coupling, and exploring two-dimensional
        numerical models of Earth-surface dynamics},
    author={Hobley, Daniel EJ and Adams, Jordan M and Nudurupati,
        Sai Siddhartha and Hutton, Eric WH and Gasparini, Nicole M and
        Istanbulluoglu, Erkan and Tucker, Gregory E},
    journal={Earth Surface Dynamics},
    volume={5},
    number={1},
    pages={21},
    year={2017},
    publisher={Copernicus GmbH}
    }
```


nicgaspar:
This is cool! Thanks Eric. Maybe we should talk about what is appropriate for citations? I think the Hobley et al. citation is a good one for any Landlab model, but should it also be in all of the components? Maybe it could be in the grid classes, which would then, defacto, make it come up in any land lab model?

mcflugen:
@nicgaspar: The Hobley et al. reference will be added as soon as you import landlab. You don't need to even instantiate a component or a grid. It might be nice to have another general landlab reference that is more dynamic, though. Like a user manual or something?

PR #581, opened 2017-10-29
Gt ca uplift

gregtucker:
This PR started as a bug fix for the cellular automata "lattice uplifter" but has ended up making an important change to HexModelGrid. Hex grids with a rectangular shape have two "ragged" edges in which the positions of the outermost nodes are staggered. Because the boundary flag was set by VoronoiDelaunay, which uses a convex hull to find boundary points, the "inner" of the staggered edge nodes were NOT flagged as boundaries, even though they properly should be. That is now fixed (and the fix method is tested).
This has a side effect on two tests in PotentialityFlowRouter: there are now fewer core nodes, and the solution is slightly different for a hex grids. The new numbers in these tests looked reasonable, so I simply copied them, but @siccarpoint should take a look to make sure all is well.


SiccarPoint:
Looks good to me, esp. from point of view of the changes to the potentiality FR. Merge at your leisure.

mcflugen:
@gregtucker: Your Python 3 errors are because Python 3 uses [absolute imports](https://www.python.org/dev/peps/pep-0328/). This means the following is looking for a package named `cts_model` at the top level (the current folder is ***not*** searched)
```python
from cts_model import CTSModel
```
Instead, this should be,
```python
from .cts_model import CTSModel
```
This tells Python that `cts_model` is in the current folder.

The `int64` error is because 64bit Windows returns 64 bit integers when referring to memory (so functions like `where`) but the default `int` is only 32 bit. We have a function to help with this called `as_id_array` that converts (if necessary) integer arrays to landlab id arrays (arrays of `int`). So,
```python
self.links_to_update = where(logical_or(lower_active, active_bnd))[0]
```
could be
```python
self.links_to_update = as_id_array(where(logical_or(lower_active, active_bnd))[0])
```

landlab/ca/tests/grain_hill.py
```diff
@@ -6,17 +6,15 @@
 _DEBUG = False
 
 import sys
-from .cts_model import CTSModel
-from .lattice_grain import (lattice_grain_node_states,
+from cts_model import CTSModel
+from lattice_grain import (lattice_grain_node_states,
```
mcflugen: Use relative imports. So, 
```python
from .cts_model import CTSModel
from .lattice_grain import lattice_grain_node_states, lattice_grain_transition_list
```

landlab/grid/hex.py
```diff
@@ -222,10 +222,54 @@ def _initialize(self, base_num_rows, base_num_cols, dx, orientation,
...
+            self.status_at_node[bot_row] = self.status_at_node[0]
+            top_row = numpy.where(self.y_of_node
+                                  >= (self._nrows - 1) * dx)[0]
+            self.status_at_node[top_row] = self.status_at_node[0]
+        else:  # horizontal orientation; left & right staggered
+            left_row = numpy.where(self.x_of_node <= 0.5 * dx)[0]
+            self.status_at_node[left_row] = self.status_at_node[0]
+            right_row = numpy.where(self.x_of_node
+                                    >= (self._ncols - 1) * dx)[0]
+            self.status_at_node[right_row] = self.status_at_node[0]
```
mcflugen: `HexModelGrid` should implement the following methods,
* `nodes_at_top_edge`
* `nodes_at_bottom_edge`
* `nodes_at_left_edge`
* `nodes_at_right_edge`
These are also the names we use in other grids like `RasterModelGrid`.

There should also be a better way than examining every node's *x* and *y*. In addition, this doesn't take into account an origin that isn't *(0, 0)*.

landlab/ca/boundaries/hex_lattice_tectonicizer.py
```diff
@@ -473,6 +475,40 @@ def __init__(self, grid=None, node_state=None, propid=None, prop_data=None,
...
+        g = self.grid
+        nc = g.number_of_node_columns
+        max_link_id = (3 * (nc - 1) + 2 * ((nc + 1) // 2) + nc // 2
+                       + (nc - 1) // 2)
+        lower_active = logical_and(arange(g.number_of_links) < max_link_id,
+                                   g.status_at_link == 0)
+        boundary = logical_or(g.status_at_node[g.node_at_link_tail] != 0,
+                              g.status_at_node[g.node_at_link_head] != 0)
+        active_bnd = logical_and(boundary, g.status_at_link == 0)
+        self.links_to_update = where(logical_or(lower_active, active_bnd))[0]
```
mcflugen: The array returned by `np.where` should be converted to `int` (it will be `int64` on some machines). This can be done with,
```python
self.links_to_update = as_id_array(where(logical_or(lower_active, active_bnd))[0])
```
`as_id_array` is in `landlab/core/utils.py`.

PR #585, opened 2017-11-12
Remove cruft from lake_mapper

mcflugen:
This pull request just removes some old and unused code from `DepressionFinderAndRouter` (and uses landlab's `error_message` and `warning_message` for printing messages).


SiccarPoint:
Not me...

Issue #600, opened 2017-12-05
Depth Dependent Diffuser boundary conditions


kbarnhart:
Testing has indicated that the Depth Dependent Diffuser (and probably the Taylor Depth Dependent Diffuser)makes soil on open boundary nodes and considers them in some flux calculations. This should not be the case and can create instabilities.

PR #601, opened 2017-12-05
Address depth dependent boundary conditions

kbarnhart:
This PR addresses Issue #600. It covers both the DepthDependentDiffuser and the TaylorDepthDependentDiffuser. 

@Glader011235 @gregtucker can you check it out and see if it fixes the problems you are having. 

If it does I'll address failing tests and and create any necessary new ones.

### DIFF
--- a/landlab/components/depth_dependent_diffusion/hillslope_depth_dependent_linear_flux.py
+++ b/landlab/components/depth_dependent_diffusion/hillslope_depth_dependent_linear_flux.py
@@ -15,7 +15,7 @@ class DepthDependentDiffuser(Component):
     This component implements a depth and slope dependent linear diffusion rule
     in the style of Johnstone and Hilley (2014).
 
-    This component will ignore soil thickness in non core nodes.
+    This component will ignore soil thickness located at non-core nodes.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


gregtucker:
This looks like the right fix to me. In general, components should only modify fields at core nodes, not boundaries (though there may be exceptions). 

Redefining self._active_nodes is a good way to test the fix, but if it works I think it would be clearer to remove self._active_nodes and directly use self.grid.core_nodes.

Glader011235:
Sounds like that’ll work- I’ll let you know as soon as I test it out. 

Rachel


--
Rachel Glade
University of Colorado, Boulder
PhD Candidate
rcglade@gmail.com
rachel.glade@colorado.edu

kbarnhart:
thanks @Glader011235 and @gregtucker 

I've cleaned up the references to core nodes and have fixed tests. 

if tests pass and @Glader011235 finds this works in her application, I'll go ahead and merge.

Glader011235:
@kbarnhart Your fix works and keeps soil thickness constant on the boundary nodes- but unfortunately, I'm still seeing the same odd behavior in the analytical test, so there must be something else going on. Will continue trying to figure out what's going on...

That said, it would still be good to merge your changes, since we do want to avoid fluxing/weathering on the boundaries and the component should have been doing that all along (mea culpa).

kbarnhart:
@Glader011235 I figured out your mismatch issue. Will send the notebook shortly.

Issue #604, opened 2017-12-19
channel profiler only plots one channel per watershed. 


kbarnhart:
Ideally, a channel profiler would either plot the one largest channel in a watershed or would plot all channels with a drainage area below a threshold.

PR #605, opened 2017-12-19
improvements to channel profiler

kbarnhart:
addresses issue #604 

now multiple channel segments in a drainage can be plotted by specifying main_channel_only=False

also added some documentation, probably not enough, but at there is now more than there was. 

also adds a plot map-view function.


kbarnhart:
@gregtucker @SiccarPoint any comments on this?

It shouldn't change the behavior of the channel profiler (current behavior is set as defaults).

@mcflugen it looks like the only not-passing is half the travis tests where landlab was imported correctly. This means that (assuming @gregtucker & @SiccarPoint , or others) don't have a problem, its OK to merge, right?

gregtucker:
Looks great to me, thanks @kbarnhart

kbarnhart:
@gregtucker thanks! I'm now about halfway through making some some additional documentation improvements, and will push up and merge once done.

SiccarPoint:
This looks great to me. On a related note though (& a wee bit off topic), could you weave this component into the reference manual docs too? Probably sits under "Terrain analysis"? We have a whole heap of new components now that don't appear to exist from the point of view of a user...

kbarnhart:
Good point Dan! I’ll do that (ASAP...) I’ll also bring up the reference
manual point today on the call unless you are there to do it.

kbarnhart:
@SiccarPoint I've now updated this components documentation

@mcflugen a question about best practices for doing test on plotting functions. I've done my best to test all parts of this code that doesn't actually create a plot. Is there a way to test the plotting?

Tagging @nathanlyons and @margauxmouchene since I know they are working on plotting capabilities and may either already know the answer or be interested in the answer.

Issue #603, opened 2017-12-19
Travis test fails for watershed utility


nathanlyons:
Travis tests are failing for #545.
- Python 2 tests fail with ImportError: No module named landlab.
- Python 3 tests have this error: "The command "conda install python=$TRAVIS_PYTHON_VERSION" failed and exited with 120 during ." This appears to happen when python is installing.

@mcflugen, any suggestions?

nathanlyons:
@mcflugen, the majority of tests for #545 are passing after #607 was merged. One Mac test of #545 (stalled build) and all tests of #608 (PackagesNotFoundError for landlab) are failing.

mcflugen:
@nathanlyons I think it was a problem with a new version of *conda* (4.4.1). They've released a newer version (4.4.4), which seems to fix the bug that caused the error we were seeing (not being able to install using `--use-local` - conda/conda#6520). I restarted your builds and they are now completing successfully.

nathanlyons:
@mcflugen The watershed utility tests continue to fail. It seems due to dependencies for Python version <3.6. Is this a concern or is it OK if we move forward?

mcflugen:
@nathanlyons It's not just the watershed utility. Everyone's tests were failing (see #633). It looks like you're working on a pretty old branch. If you're able, I would update your branch with what's on master.

PR #606, opened 2017-12-20
only apply ero/dep to core nodes

gregtucker:
This PR fixes an issue with Taylor Nonlinear Diffuser, in which erosion/deposition was being applied to open boundary nodes. Now only applied to core nodes.


kbarnhart:
Looks like all Linux builds failed. @mcflugen do you know if this is a Travis issue?

Other than that, @gregtucker , it looks good to me.

mcflugen:
@kbarnhart Yes, the errors are probably Travis related. See #603 and #607.

kbarnhart:
thanks @mcflugen !

@gregtucker I'll merge once/if this passes.

PR #608, opened 2017-12-22
field profiler

nathanlyons:
FieldProfiler plots the values of fields along a profile with endpoints defined by nodes or coordinates. This functionality is needed for teaching notebooks.


kbarnhart:
@nathanlyons I wanted to check in about data structures in the profile components/utilities. 

#878 is getting close to ready. And we had talked (or I had proposed... I forget) that we think about whether there should be parallels in the data structures used by `ChannelProfiler` and `FieldProfiler`

Looking more closely at `FieldProfiler` I'm not entirely sure this makes sense... but thought I'd post here to discuss. 

`ChannelProfiler`  creates a datastructure `self.profile_structure` that is a  a list of length number_of_watersheds. Each element of that list is itself a list of length number of stream segments within that watershed (one if the flag main_channel_only is True, otherwise, potentially with length more than one).  Each  stream segment list contains the node ids of a stream segment from downstream to upstream.

I *think* that `FieldProfiler` was built to just make one profile... correct? If so, then making these coordinate makes less sense... What do you think?

nathanlyons:
@kbarnhart It would be nifty for other profilers to be part of the same component, although how they work may be too different to justify it.

The fundamental difference between ChannelProfiler and FieldProfiler is how profiles are defined: flow routing for the former, user-defined endpoints for the latter.

I envisioned only 1 profile created at a time and a profile is made of only 2 endpoints. This could be extended (and made similar to the channel profile data structure) by allowing (1) multiple profile lines and (2) multiple profile segments by allowing users to define vertices between endpoints for zig-zaggy lines.

profile_structure hierarchy from outer to inner lists

channels | user-mapped
|---|---|
watersheds | profile lines
stream segment | line segment
stream node | profile node

Plotting may be different for the segments. Are stream segments plotted continuously along the stream/profile distance or is each segment plotted separately? For user-mapped profiles, all line segments of a given profile are plotted continuously. Alternatively:

channels | user-mapped
|---|---|
watersheds | n/a
stream segment | profile line
stream node | profile node

The columns on each row of this table are more similar. (There's no reason to retain line segments once the profile nodes are determined.) The catch is the outer list is always length of 1 (a list of profiles). Alternatively, would it be problematic to have different data structures? Would it be problematic to make the base profiler class (_NetworkProfiler) more general, eg not check for flow receiver fields at least when a user-mapped profiler will be created?

Issue #609, opened 2017-12-27
if dt is type numpy.float64, StreamPowerEroder uses incorrect Brent method. 


kbarnhart:
The following snippet
```python
from landlab import RasterModelGrid
from landlab import FIXED_VALUE_BOUNDARY
from landlab.components import FlowAccumulator, StreamPowerEroder, LinearDiffuser
import numpy as np

exp = np.arange(0, 8)
dts = 10 * 2**exp

K_sp = 0.001
m_sp = 0.5
n_sp = 1    
uplift_rate = 0.01

for dt in dts:

    mg = RasterModelGrid((10,10))
    z = mg.add_zeros('node','topographic__elevation')

    for edge in (mg.nodes_at_left_edge, mg.nodes_at_right_edge, mg.nodes_at_bottom_edge, mg.nodes_at_top_edge):
        mg.status_at_node[edge] = FIXED_VALUE_BOUNDARY
        z[edge] = 0
    
    fa = FlowAccumulator(mg, depression_finder='DepressionFinderAndRouter')
    sp = StreamPowerEroder(mg, K_sp=K_sp, m_sp=m_sp, n_sp=n_sp, use_Q='surface_water__discharge')
    ld = LinearDiffuser(mg, linear_diffusivity=0.0001)
 
    for i in range(100):
    
        fa.run_one_step()
        sp.erode(mg, dt=dt)
        ld.run_one_step(dt)
    
        # uplift the core nodes.
        mg.at_node['topographic__elevation'][mg.core_nodes] += dt * uplift_rate # add the uplift
```

gives the following error

```python
Traceback (most recent call last):

  File "<ipython-input-1-9013b5e84488>", line 34, in <module>
    sp.erode(mg, dt=dt)

  File "/Users/barnhark/projects/landlab/landlab/components/stream_power/stream_power.py", line 578, in erode
    self._n, z)

TypeError: Argument 'threshsxdt' has incorrect type (expected numpy.ndarray, got numpy.float64)
```

This is because the line at 571 states;
```python
if isinstance(threshdt, float):
```

A `np.float64` is not a `float` so this yield false. This is easily fixed by using `if isinstance(threshdt, float):` since `np.float64` inhertits from `float`

PR #610, opened 2017-12-27
Fixes type checking issue

kbarnhart:
Addresses issue #609


landlab/components/stream_power/stream_power.py
```diff
@@ -568,7 +568,7 @@ def erode(self, grid, dt, elevs='topographic__elevation',
             threshdt = self.sp_crit * dt
 
         # solve using Brent's Method in Cython for Speed
-        if type(threshdt) == float:
+        if isinstance(threshdt, float):
```
mcflugen: Looks good. We should never be using `type(...) == ...`. I've tried to remove them when I see them but I'm sure there's more still.

PR #613, opened 2017-12-31
Tidy up continuous integration and versioning

mcflugen:
This pull request tidies up our continuous integration (both Travis and Appveyor).

The highlights:
* Fixed the failing builds on 64 bit WIndows
* Updated the Anaconda Cloud token for uploading Windows distributions (we haven't been uploading Windows distributions for a while).
* Removed separate builds for different versions on numpy. Now we build with numpy 1.11 but require anything greater than that at runtime.
* Use [versioneer](https://github.com/warner/python-versioneer) for tracking versions. This means we only need to update version numbers in one place and that version numbers are automatically generated in between releases.


gregtucker:
Looks great to me @mcflugen. Nice to have just one place to keep version. Only question is that I don't really understand "we build with numpy 1.11 but require anything greater than that at runtime" -- does this mean we require 1.11 OR greater?

mcflugen:
@gregtucker Correct. We compile with numpy 1.11 but, since subsequent versions are backward compatible, we are able to require numpy equal to or greater than 1.11 when landlab is actually installed. Previously we had built with different numpy versions and required that particular version when landlab was installed. Since we hadn't kept up with new versions of numpy we didn't have a build for the latest version (1.13), which meant that if someone installed landlab, their numpy version would have been downgraded to match what we built with.

Issue #614, opened 2018-01-12
Raster Model Grid seems to rebuild diagonal and d8 quantities -- issue with @cache_result_in_object()?


kbarnhart:
In re-running a model that I've run before an extensively profiled I noticed that it was taking quite a bit (at least 2-3x) longer than I expected. I re-profiled it and one thing that seems to be taking longer than I expected was the construction of lists and/or numpy arrays associated with the new diagonals functionality. 

I've included a super simple version of my model (complete with profiling code). Obviously this example doesn't take that long to run, but I've verified that the order of the profiling is similar to my more complicated example. Profile output listed below python code. 

It appears to me that this may have something to do with the `@cache_result_in_object()` decorator not working as expected since the name of the decorator makes me think it was intended to solve this problem. The high ranking of `_wrapped()` in the profile results may also support this suspicion (1.643 seconds out of 2.017 seconds in the toy example). 

@mcflugen any thoughts on how to fix this?

Some of the functions that appear to be taking longer than I expected:
_wrapped
diagonal_adjacent_nodes_at_node
adjacent_nodes_at_node
diagonals_at_node
create_diagonals_at_node
d8s_at_node
nodes_at_diagonal
create_nodes_at_diagonal

```python
import cProfile, pstats, io

from landlab import RasterModelGrid
from landlab import FIXED_VALUE_BOUNDARY
from landlab.components import FlowAccumulator, StreamPowerEroder, LinearDiffuser

dt = 10
K_sp = 0.001
m_sp = 0.5
n_sp = 1    
uplift_rate = 0.01

mg = RasterModelGrid((30,30))
z = mg.add_zeros('node','topographic__elevation')

for edge in (mg.nodes_at_left_edge, mg.nodes_at_right_edge, mg.nodes_at_bottom_edge, mg.nodes_at_top_edge):
    mg.status_at_node[edge] = FIXED_VALUE_BOUNDARY
    z[edge] = 0

fa = FlowAccumulator(mg, flow_director='D8', depression_finder='DepressionFinderAndRouter')
sp = StreamPowerEroder(mg, K_sp=K_sp, m_sp=m_sp, n_sp=n_sp, use_Q='surface_water__discharge')
ld = LinearDiffuser(mg, linear_diffusivity=0.0001)
 
pr = cProfile.Profile()
pr.enable()
for i in range(100):

    fa.run_one_step()
    sp.erode(mg, dt=dt)
    ld.run_one_step(dt)

    # uplift the core nodes.
[...]

PR #615, opened 2018-01-14
Fix issue #614 - Values not being cached

mcflugen:
@kbarnhart this pull request should fix the problem you identified in #614. Thank you so much for finding this! Good eye!

The problem was a small typo in the `cache_result_in_object` decorator, just as you thought. The decorator was checking for the cached value in the wrong place, so never found it, and so the value was always being recalculated.

Please see if this fixes your problem and if things are running at the same speed as they were before.

I also added some tests to make sure this doesn't happen in the future.


kbarnhart:
@mcflugen this seems to have solved my problem (I get a ~40x speedup as a single timestep for my complex model goes from 41.462 seconds to 0.937 seconds).  

However,  I get 43 test failures when I build and test this branch (I'm on commit 7faad8cc61df2b414642df21ccaeb5990db87b22 which I believe is the most recent). I haven't been able to figure out the theme of the failures, so I've attached a text file that has the test output in it. Maybe you'll see the pattern...

[build_log.txt](https://github.com/landlab/landlab/files/1629824/build_log.txt)

My standard workflow for rebuilding is:

```python
python setup.py develop -u
conda install --yes --file=requirements.txt
python setup.py develop
python scripts/test-installed-landlab.py
```

mcflugen:
@kbarnhart Thanks. I see the problem. It should be another easy fix.

kbarnhart:
@mcflugen super!

mcflugen:
@kbarnhart Not quite as easy as I thought but these latest changes should get us pretty close. Please have a look and see if they fix things for you.

kbarnhart:
@mcflugen this passes tests on my computer and profiles to be nearly identical to the last profile test.

mcflugen:
@kbarnhart Great! Since all the tests are passing on your machine and mine as well as AppVeyor, I think things are working.

Something strange is going on with our Travis builds that is causing them to fail. I've rerun builds that had previously passed and now they're failing so I don't think it has anything to do with these latest commits.

kbarnhart:
@mcflugen I looked through the code and as best I can tell it looks good. I'm not super familiar with code you've modified related to testing and conda requirements, though.

I'm OK with this being merged whenever you think it is OK.

thank you!

PR #627, opened 2018-03-05
Add NetworkModelGrid

mcflugen:
This pull request adds a new model grid, `NetworkModelGrid`, that's just a graph of nodes, and links. There's also data fields, of course. This isn't at all ready to be merged but I figured I would get things started to see what folks think.

Create a *Y* network of two links that join into a single link.
```python
>>> from landlab.grid.network import NetworkModelGrid
>>> grid = NetworkModelGrid(((0, 1, 2, 2), (0, 0, -1, 1)), ((1, 0), (2, 1), (3, 1)))
```
The grid has most all of the same methods as other landlab grids,
```python
>>> grid.x_of_node
array([ 0.,  0., -1.,  1.])
>>> grid.y_of_node
array([ 0.,  1.,  2.,  2.])
>>> grid.nodes_at_link
array([[0, 1],
       [2, 1],
       [1, 3]])
```
Notice that nodes, links, and link directions have been changed to match the usual landlab ordering. For this case, we may want to disable this reordering.

Existing functions that work on links and nodes on unstructured grids should also work with network grids. For example,
```python
>>> from landlab.grid.gradients import calc_grad_at_link
>>> grid.at_node['z'] = [1,2,3,4]
>>> calc_grad_at_link(grid, 'z')
array([ 1.        , -0.70710678,  1.41421356])
```

Because `NetworkModelGrid` is based on the graph data structures, there are a couple extras,
```python
>>> from landlab.plot import graph
>>> graph.plot_graph(grid, at='node,link')
```
Netcdf writing is also a little nicer. The following dumps the grid to a netcdf file in UGRID format and writes fields for all locations (even if they happen to have the same name).
```python
>>> grid.at_link['z'] = [10,20,30]
>>> grid.at_node['z'] = [1,2,3,4]
>>> grid.to_netcdf('network.nc')
```


kbarnhart:
Following along I'm going to write some notes here on improvements/things to identify:

1) Identify input mechanisms/formats. Some things to consider:
a) Shapefile
b) RasterModelGrid and a drainage area threshold
c) Ugrid
d) Formats used by DHSVM preparation tools
e) USGS HUC 
f) others?

2) Identify some common use cases that this model grid can work with
a) Sediment transport
b) Routing a kinematic wave hydrographic
c) Coupled long profile evolution
d) Others

3) Determine which attributes/mappers/calculations are necessary and supported
a) gradients (done), 
b) more here....

4) Determine which data storage formats are possible
a) at element
b) ragged arrays 

5) Create `NetworkModelGrid` with the intention of having it interface and map back and forth with a fully-determined with-area grid. 
a) identify a mechanism to map network and area grid elements to one another. 
b) Determine if nodes need to be co-located.

kbarnhart:
Ahhh. @mcflugen that makes sense. I figured that there was a reason, but it wasn't immediately apparent.

kbarnhart:
A long awaited update from the NetworkModelGrid call from Friday 3/9.

tagging: @mcflugen @saisiddu @SiccarPoint @gregtucker @nicgaspar @erkanistan @margauxmouchene @nathanlyons @awickert @christinaB

Need to add tags to Allison and John but I don't think they are on Github yet.

**Update**
On the call we spent most of the time going over the [Jupyter notebook that I've made to highlight the capabilities of NetworkModelGrid](https://github.com/landlab/tutorials/blob/barnhark/nmg_tutorials/network_model_grid/Example_Usage_of_Network_Model_Grid.ipynb). The major topics of discussion were:

a) Goals for input and output capabilities
b) Discussion of additional capabilities

_IO_
Regarding IO we determined that we should be able to support import of a shape file of the river network (polyline type shapefiles). If a river segment includes a more complicated line than from the head node to the tail node, the NMG should also save the full path of the channel segment. 

We should consider compatibility with output form the [National Hydrography Dataset](https://nhd.usgs.gov)

A second option is to take an input DEM and a drainage area threshold. We'd have two options,  keep all nodes and just keep stream segment ends.

There was enthusiastic support for the ability to map back and forth between a NetworkModelGrid and a RasterModelGrid. 

_Capabilities_

The discussion about NMG capabilities had two main points. 
1) The NMG should have all of the abilities and attributes that a RasterModelGrid currently has (for just links and nodes). 

This is because most of the things that John's model does needs to move up and down the network based on node adjacency. Being able to use the FlowDirectorSteepest and FlowAccumulator is also a need. 

To this end, I made a first attempt at splitting the current ModelGrid (the parent of RasterModelGrid) into two pieces, one that just has the node and link capabilities and one that has everything else. This is currently in development on the [barnhark/network_model_grid branch](https://github.com/landlab/landlab/tree/barnhark/network_model_grid). As best as I can tell I've done this without breaking the tests. I will bring it up on this weeks Landlab call for discussion. Some of the additional pieces of functionality are now visible in the tutorial mentioned above. 

The second thing we discussed is the ability have sediment parcels with attributes like volume and grain size that move around on the grid. John needs to do things like take the mean of the grain size of all the parcels at link or the sum of the sediment volume. 

I talked with @margauxmouchene  about making this compatible with her clast tracker and with @mcflugen  about the feasibility of using layers for this application. Based on these conversations I decided that a new datastructure made the most sense. It is not yet complete, but it is presently called ItemCollection and is located on  the branch [barnhark/create_item_collection](https://github.com/landlab/landlab/tree/barnhark/create_item_collection). This is just a Pandas Dataframe that enforces a few requirements: items must live on a part of the grid (e.g. link, node), and items can't live on parts of the grid that don't exist (e.g. if there are 100 links, an item can't live on link 200). This will allow us to take advantage of Pandas [split-apply-combine capabilities](https://pandas.pydata.org/pandas-docs/stable/groupby.html) which are exactly what we need for things like  "average grainsize at link".

ChristinaB:
@kbarnhart   To add on to the list above
Identify some common use cases that this model grid can work with
a) Sediment transport
b) Routing a kinematic wave hydrographic
c) Coupled long profile evolution
d) Energy Transport  (Temperature modeling)
e) Contaminant Transport (Water Quality modeling)

Can I join the mini-hack before CSDMS remotely?  I am not planning to attend CSDMS, but I am tempted to join just for the Monday before to join you.

kbarnhart:
@ChristinaB I don't see why not. It would be great to have you. 

Thanks also for adding the other use cases.
 
I don't have much experience running hack-a-thons with remote participants. Have you done this? Maybe you can propose the best way to do this (slack + long running zoom call?)

awickert:
@kbarnhart @ChristinaB This outline and this set of capabilities for NetworkModelGrid sound excellent to me, and I wanted to comment that it seems that this framework will also be able to be used (without a network) to run **1D (e.g., long profile, cross-sectional, hillslope) models** in the Landlab framework.

Do we need any more discussion on a convention to connect links to one another in a network? In my work, which looks broadly similar to @mcflugen's approach above, each link has an ID and a "to_link" attribute that tells me which link the previous link enters. I include a "distance downstream" attribute that tells me which end of the link is the start, and what the distance of the points on the link are (Instead of just nodes between links, each link in my world has its own set of (x,y) points), and therefore, where upstream flow enters the link. I assume that each link sends flow/sediment/etc. to only one other link; this would have to be changed to allow for bifurcations in networks.

kbarnhart:
@awickert great thoughts. I think that Landlab currently has sufficient conventions. I've summarized them below so all who are working on this are aware of them. If anyone thinks that additional information needs to be added, lets put it on the agenda for the Wednesday teleconference. 

Thus far, the Landlab stores connectivity information in two main ways. The first way is inherent to the grid (with no components, and no assumptions about what the grid might be used for). This convention is that the head node and tail node are defined based on geometric relation between the nodes in x and y space rather than by _flow direction_. 

The attributes: `links_at_node`, `node_at_link_head`, `node_at_link_tail`, and `link_dirs_at_node` provide information about the connectivity structure of the links and nodes within the landlab ordering and direction reference frame. No assumptions are made about maximum number of links at a node, or flow divergence/convergence. [This link](http://landlab.readthedocs.io/en/latest/landlab.grid.base.html#information-about-links) is to the part of the documentation that describes information about links. 

Then, if a user uses the component `FlowAccumulator` (which has been created to replace `FlowRouter`, a new set of fields that define connectivity are created. 

If nodes have an surface elevation, provided by a at-node model grid field (typically `topographic__elevation`), running the `FlowAccumulator` component on the model grid creates the fields relevant for the flow of water and sediment. As of the current working version of NetworkModelGrid (branch [barnhark/nmg_third_attempt](https://github.com/landlab/landlab/tree/barnhark/nmg_third_attempt)) the NetworkModelGrid can work with the FlowAccumulator correctly _and_ can read in shapefiles based on examples Jon has provided me. 

`flow__receiver_node` indicates the node ID of the flow reciever (and gives a nodes own ID if it is an outlet node or a sink). This makes it easy to find outlet nodes.  If route to multiple methods are used, then `flow__receiver_nodes` and `flow__receiver_proportions` provide information about flow receivers and proportions, respectively. 

While a _distance upstream_ attribute currently does not exist. This is something that is being developed in response to Issue #638 and will definitely be done before CSDMS. It requires that the `FlowAccumulator` has already been run as it uses the data structures this component creates. 

Depending on if flow is directed by a route-to-one (e.g. D4, D8), or route-to-multiple method (Dinf, MFD), different fields are created. The documentation for [`FlowAccumulator`](http://landlab.readthedocs.io/en/latest/landlab.components.flow_accum.html) is pretty compressive, but I've summarized a bit here. 

FlowAccumulator stores as ModelGrid fields:

-Node array of drainage areas: *'drainage_area'*
-  Node array of discharges: *'surface_water__discharge'*
-  Node array containing downstream-to-upstream ordered list of node IDs: *'flow__upstream_node_order'*. 
-  Node array of all but the first element of the delta data structure:  *flow__data_structure_delta*. The first element is always zero. 
-  Link array of the D data structure: *flow__data_structure_D*

The FlowDirector component will add additional ModelGrid fields. DirectToOne Methods (Steeptest/D4 and D8) store the following as ModelGrid fields:

- Node array of receivers (nodes that receive flow), or ITS OWN ID if there is no receiver: *'flow__receiver_node'*
-  Node array of steepest downhill slopes:  *'topographic__steepest_slope'*
-  Node array containing ID of link that leads from each node to its receiver, or BAD_INDEX_VALUE if no link:  *'flow__link_to_receiver_node'*
-  Boolean node array of all local lows: *'flow__sink_flag'*

DirectToMany Methods (MFD) store the following as ModelGrid fields:

-  Node array of receivers (nodes that receive flow), or ITS OWN ID if there is no receiver: *'flow__receiver_nodes'*. This array is 2D, and is of dimension (number of nodes x max number of receivers).
-  Node array of flow proportions: *'flow__receiver_proportions'*. This array is 2D, and is of dimension (number of nodes x max number of receivers).
-  Node array of links carrying flow:  *'flow__links_to_receiver_nodes'*. This array is 2D, and is of dimension (number of nodes x max number of receivers).
-  Node array of the steepest downhill receiver. *'flow__receiver_nodes'*
-  Node array of steepest downhill slope from each receiver: *'topographic__steepest_slope'*
-  Node array containing ID of steepest link that leads from each node to a receiver, or BAD_INDEX_VALUE if no link: *'flow__link_to_receiver_node'*
-  Boolean node array of all local lows: *'flow__sink_flag'*

ChristinaB:
A note on ROLES for Christina - - My idea for simplifying things (after we talked on 4/16) is to use Jon's shapefile and make it look like a DHSVM input file, that way we can test two inputs have the same results/outputs.  If you want me to do that, just send/share Jon's file.     
We an test DHSVM pre-processor generated shapefiles as a next step. 
Task 1: Test that Katy's current shapefile I/O works for DHSVM shapefiles.  Elwha, Sauk, Puyallup (Allison's next paper).  How do these examples fit (or not) the assumptions that fit for Jon's shapefile?
Task 2: Identify terrain analysis methods that generates shapefiles that fit assumptions. Identify what terrain features break the I/O.
Task 3: Write up assumptions Katy made.  This all may be redundant and so well designed we will have clear sailing, but I'll believe it when I see it (in writing)

kbarnhart:
@mcflugen as best as I can tell these commits fix all the errors I see! thanks!

kbarnhart:
@mcflugen. I think there is one additional error left in this branch:

I get 

```
ERROR: landlab.field.tests.test_graph_fields.test_delete_field
  File "/anaconda3/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/barnhark/projects/landlab/landlab/landlab/field/tests/test_graph_fields.py", line 173, in test_delete_field
    fields.delete_field('link', 'vals')
  File "/Users/barnhark/projects/landlab/landlab/landlab/field/graph_field.py", line 901, in delete_field
    ds._ds = ds._ds.drop(name)
  File "/anaconda3/lib/python3.6/site-packages/xarray/core/dataset.py", line 2255, in drop
    return self._drop_vars(labels)
  File "/anaconda3/lib/python3.6/site-packages/xarray/core/dataset.py", line 2266, in _drop_vars
    self._assert_all_in_dataset(names)
  File "/anaconda3/lib/python3.6/site-packages/xarray/core/dataset.py", line 2234, in _assert_all_in_dataset
    raise ValueError('One or more of the specified variables '
ValueError: One or more of the specified variables cannot be found in this dataset

```

do you see this one?

kbarnhart:
I'm presently working on an updated version of this PR. I'm going to close this one.

Issue #639, opened 2018-03-15
No NormalFault component 


kbarnhart:
Landlab doesn't have a component that makes a rock uplift pattern consistent with a normal fault.

PR #637, opened 2018-03-14
Updates to channel profile and fastscape

kbarnhart:
I started to get testing errors related to the channel profiler and `FastscapeEroder`. 

these mostly had to do with running either on a HexGrid which doesn't have the attribute `grid.length_of_d8`. I added a type check to fix this. 

a similar issue came up with the stream profiler which was using `_length_of_link_with_diagonals`, the precursor to` length_of_d8`. I also added a type check here. 

While doing this I cleaned up the lems used in the tests to make them simpler

PR #640, opened 2018-03-15
Barnhark/normal fault

kbarnhart:
This PR addresses Issue #639 and creates a normal fault component. 

I believe it will break until PR #637 is updated (and merged into this branch) as it uses a HexModelGrid with the FastscapeEroder component. 

Associated with this branch is a tutorial PR.


kbarnhart:
thanks @gregtucker.

I've now modified the component so there can be more than one uplifted field, added tests for this and
added a bit to the tutorial. If it passes I'll merge.

PR #643, opened 2018-03-20
Barnhark/exponential weatherer docstring

kbarnhart:
Addresses Issue #642


kbarnhart:
hmmm. it should probably be:

`soil_production__maximum_rate`

and

`soil_production__decay_depth`

These were what I came up with for bento as well.

PR #653, opened 2018-03-24
Fix names of packages deployed to Anaconda Cloud

mcflugen:
This pull requests fixes a couple small bugs that caused package names deployed to Anaconda Cloud to be not quite right. On Mac and Linux, package versions didn't drop the "v" (so, `v1.3.0` instead of just `1.3.0`). On Windows were were using the new-style conda-build build strings (these are the ones with a hash for the dependencies) and incorrectly contained the `dev` tag.

Basically, the changes are:
*  Tell versioneer to ignore the "v" at the start of tags
*  Remove the BUILD_STR from Windows builds and use `--old-build-string`


.travis.yml
```diff
@@ -57,20 +58,22 @@ virtualenv:
...
-  provider: pypi
-  user: mcflugen
-  password:
-    secure: V15DceMv4DUmjgxToaxrP3KaQxsLbfrTkB1s6D8YNQZvYJiVvNlBpC1nF8yWbOt3IWJQSGjaQEnObd2ZhHe2Kq/H+QsEFyBKlYF7SNp35FyS5zLuf2qhnNhmzhqkspwBpiGe88c1PQPFzfqdIUQXp7ZAv6L8YRvNy2sbzvgxh6I=
-  on:
-    tags: true
-    branch: release
-    repo: landlab/landlab
-  distributions: sdist bdist_wheel
+# deploy:
```
kbarnhart: @mcflugen just checking that you meant to keep all of the deploy portion commented out. 
assuming so, this looks oK to me.

mcflugen: Yes, that is fine. We deploy to both Anaconda Cloud and PyPI in the `after_success` step so this isn't necessary.

Issue #656, opened 2018-03-27
Coveralls doesn't seem to be getting data for coverage


kbarnhart:
Seems like the last commit that was tested was [this one](https://coveralls.io/builds/14850976) on 30 Dec 2017.

I might have a solution developed dealing with this in `terrainbento`

@mcflugen , maybe you can suggest if it is good enough. I'll make a pr

kbarnhart:
@mcflugen I fixed this in `terrainbento` mostly in [this PR](https://github.com/TerrainBento/terrainbento/pull/39).

A note to you, me, and whoever else might find this useful. 

Summary:
1) Switched to pytest, including moving all unit tests into a adjacent folder **WITH** an __init__.py file
2) Removed nose dependency (including all use of nose tools in testing)
3) Use pytest-cov with doctests for running and reporting on coverage.

PR #657, opened 2018-03-27
changing location of coveralls call

kbarnhart:
perhaps addresses Issue #656

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,10 +55,10 @@ script:
   if [[ "$COVERALLS" == "1" ]]; then
   pip install coveralls
   conda install -q landlab --use-local -c $(dirname $(conda build --output ./.conda-recipe/ --old-build-string))
-  python scripts/test-installed-landlab.py
+  bash .ci/travis/run_travis_test.sh
   coveralls
   fi
-  # - bash .ci/travis/run_travis_test.sh
+
 virtualenv:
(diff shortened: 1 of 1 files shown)


mcflugen:
@kbarnhart I'm not sure this fixes the problem. I think the issue is that we switched to building landlab for testing with `conda build` rather than `python setup.py install` and so coveralls doesn't see the code that was tested.

I would like to overhaul our Travis builds but, in the meantime, perhaps having just one of the builds (maybe linux and Python 2.7?) build with the old `python setup.py` and then submit the results to coveralls like we did before? What do you think?

kbarnhart:
@mcflugen , you are right, this doesn't fix the problem... It had seemed to on my `terrainbento` PR, but now it isn't fixed there either. 

I'm not sure what the best way to solve it is, but I'm ears for any ideas you have.

kbarnhart:
@mcflugen I *might* have gotten this to work on `terrainbento` by using the `.ci/travis/run_travis_test.sh` file. 

I think that coveralls can see the coverage, and reports it can on the GitHub PR, though it still shows a `coverage:unknown` badge in the coveralls page. 

coveralls build: https://coveralls.io/builds/16204569

I made this change in Landlab but I'm waiting to push it up until I know if you think this was a reasonable approach.

Issue #673, opened 2018-04-30
No Landlab Tool/Component to make working with Layered Rocks easy


kbarnhart:
Many interesting questions in Earth Surface processes deal with having spatially variable lithology (in x, y and z). 

Landlab has components (e.g. FastscapeEroder, LinearDiffuser) that can take spatially variable coefficient. Landlab also has a EventLayers tool meant to efficiently create layers. 

However, no tool yet exists to help create a block of rock (generic or layered) with spatially variable properties.

PR #674, opened 2018-04-30
Create Lithology and LithoLayers

kbarnhart:
This PR addresses Issue #673 

It should be ready for review in a few days (along with a helpful tutorial notebook).

I'll make a comment at that time to let you guys know. 

I'd also appreciate comments from @amforte and @byanites and @mwrossi


nicgaspar:
@kbarnhart but maybe more for @mcflugen 
I am simultaneously learning layers and rock block for this review, so sorry for the slowness. I am confused by the doc string for both '_deposit_or_erode' and '_get_surface_index', but I'll focus on '_get_surface_index' as that seems to have been changed for this pull. I ran the doc string and it works as suggested. 
This starts out as node 1 having 5 layers, each with depth 1. unit. Then it looks like the code wants to reduce layer 5 (of node 1) by -1. What happens is that layer 5 (of node 1) gets changed to zero, as expected, but layer above it also gets changed to zero. Why?
I tried creating my own examples, but I remain confused. For example, in the attached, it would imply to me that the depth of only layer 5 should get reduced, but instead layer 5 gets eliminated, and then the depth of layer 4 is reduced.

I assume this is doing what is desired, but maybe we need more in the doc string because I don't understand what is desired. The examples in the doc string for '_deposit_or_erode' also confused me, because using layers of depth -1 confused me.


screen shot 2018-05-02 at 2 47 24 pm

kbarnhart:
Alrighty. A few major changes have now been made. A summary:

1) Renamed to Lithology and LithoLayers.
2) Created a new type of layers called MaterialLayers that does not create layers of zeros representing erosion _and_ combines layers when the entire layer stack surface values are the same as the new layer values. Lithology can use either EventLayers or MaterialLayers (and docstring includes a description of the difference). 
3) Changed Lithology to only add rock type ID to the attributes tracked by the layers. This makes material layers compatibility checking faster AND reduces memory usage. Surface values of any rock property are still accessible and set to model grid fields through dictionary mapping. 

I'm going to hold off on making any incremental changes for a bit to let people look at this and give comments. 

The tutorial has been updated to work with this version of the PR. 

And a final note. We are planning to submit this to the Journal Open Source Software relatively soon to try out that submission process.

kbarnhart:
@mcflugen @nicgaspar I don't know if you get notified when I mention you in the commit message. 

Based on a conversation I just had with @nicgaspar I've revised this docstring of the ``_deposit_or_erode`` function in eventlayers.py`. 

Obviously this is the docstring of a private function, so you might ask, what;s the point. 

The intention here is that the usage of `_deposit_or_erode` may be confusing to someone who wants to develop with Layers. 

I've tried to improve it based on what I thought would help @ nicgaspar. Let me know if it helps. 

@mcflugen, let me know if you think anything is incorrect

kbarnhart:
Some notes on fixing the memory allocation error tests.

running the test three times in a row give three different results:

1) no error, 2) Segmentation fault 11, and 3) the error I see on travis.

```engr2-2-75-245-edu:lithology barnhark$ nosetests --with-doctest
...................................
Ran 35 tests in 0.081s

OK
Segmentation fault: 11
engr2-2-75-245-edu:lithology barnhark$ nosetests --with-doctest
........python(2362,0x7fffa27ae380) malloc: *** error for object 0x7fa62d12df38: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
engr2-2-75-245-edu:lithology barnhark$ ngr2-2-75-245-edu:lithology barnhark$ nosetests --with-doctest
-bash: ngr2-2-75-245-edu:lithology: command not found
engr2-2-75-245-edu:lithology barnhark$ ...................................
-bash: ...................................: command not found
engr2-2-75-245-edu:lithology barnhark$ ----------------------------------------------------------------------
-bash: ----------------------------------------------------------------------: command not found
```

I was able to figure out that the error had to do with using MaterialLayers and was occurring in the doctests of Lithology. 

After turning off the `@boundscheck(False)` flag off (temporarily). I determined that the issue had to do with situations in which only erosion occurred, and thus deposit_and_erode was attempting to put a zero in the new EventLayer (which didn't exist) ... Thus creating a memory allocation failure. 

Fix (I hope) was to create a new cython method for MaterialLayers that just erodes.

nicgaspar:
Some notes on the notebook...

When I run the first code block I get a warning. It's no big whoop, but I think it will cause the tutorial tests to fail? (see attached)

In the second code block, here are some questions that I had when I looked at it:
What are z0s? (maybe just add a comment saying that you are making the layer depths at the anchor point. is that correct?)
Why are some of the values negative? 
For those who don't now what np.tile is, you could just state what that line is doing in the notebook.

The truth is I'm maybe too simple an audience for this tutorial. Starting out with the Anticline was way too much for my brain. I can't envision that function. It does make for a beautiful example. But my tiny head needed to work on a 5 by 5 grid, make layers that only dipped in one direction (so even more simple than your shallow or steep dip examples, and then play around to see what the values were to try to understand them. (I'm still confused about what happens to the negative values... but I wanted to get you a few more comments.)

So, if it were me, I would not get rid of this gorgeous anticline example. Rather, before it, I would make a very small grid, with a very easy example of layers varying, and illustrate that first so people get their confidence up. I know people can read the docstrings, but I think the point of the notebook is that people should not need to read the docstring with the notebook?

(sorry, I keep looking in 15 minute chunks, so I never have enough time to actually figure things out.)

screen shot 2018-05-08 at 4 36 51 pm

kbarnhart:
@nicgaspar THANK YOU! for the comments and the info about the warning. Super helpful to improve the code, comments, and text. I agree that someone should not need to read the API to understand the tutorial. I'll revise on Thursday. 

@gregtucker BinnedLayers is not in master yet (as far as I can see...) so I didn't make any statements about it. As soon as it exists in Landlab, I expect that Lithology will be compatible with it (and we can change the docstring).

kbarnhart:
Todo: Need to fix python 2.7 error:

```
File "/home/travis/miniconda/envs/_testing/conda-bld/landlab_1525843103120/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/python2.7/site-packages/landlab/components/lithology/lithology.py", line 591, in landlab.components.lithology.lithology.Lithology.add_rock_type
Failed example:
    lith.ids
Expected:
    frozenset({1, 2, 4, 6})
Got:
    frozenset([1, 2, 4, 6])
```

kbarnhart:
I've now revised the tutorial to include @nicgaspar 's comments. I've also added additional tests and a bit of functionality to the component. 

I'd very much appreciate comments on the tutorial/usability of the component from @amforte and @byanites and @mwrossi before I submit this to JOSS.

kbarnhart:
@gregtucker @nicgaspar @mcflugen 

I've prepared a `paper.md` and `paper.bib` to submit this submodule to [JOSS](https://joss.readthedocs.io/en/latest/index.html). 

@gregtucker and I discussed authorship of this submission and we thought that the four of us would be the most appropriate. If you think the author list should be expanded, let me know. 

I'd appreciate any comments you have on the `paper.md` file (and associated references). 

I've reproduced some information about JOSS here so that it is easier to make comments.

Its unclear to me how/if JOSS handles submission of a submodule (rather than an entire repository). For example, Landlab has contribution guidelines, a README, installation instructions, etc. However, we will see how they want to handle this and will then know more about whether JOSS is an appropriate place to submit Landlab components/submodules.

I'm about to be on vacation (presently in the Munich airport) so there is not a huge rush on providing comments/edits to the `paper.md`. Ideally I'd like to submit this no later than Aug 27-29. Let me know if that timeline is a problem for you. 

If you have comments I'd propose that you include them as commits directly on this PR. 

Between now and then I'll finalize the tests for MaterialLayers and EventLayers (they are not 100% tested at the moment). 

-----------------------
Information from JOSS
-----------------------
### What should my paper contain?

JOSS welcomes submissions from broadly diverse research areas. For this reason, we request that authors include in the paper some sentences that would explain the software functionality and domain of use to a non-specialist reader. Your submission should probably be somewhere between 250-1000 words.

In addition, your paper should include:

A list of the authors of the software and their affiliations
A summary describing the high-level functionality and purpose of the software for a diverse, non-specialist audience
A clear statement of need that illustrates the purpose of the software
A list of key references including a link to the software archive
Mentions (if applicable) of any ongoing research projects using the software or recent scholarly publications enabled by it

### Review criteria

[JOSS page here](https://joss.readthedocs.io/en/latest/review_criteria.html)

kbarnhart:
@nicgaspar @gregtucker @mcflugen if you could take a look at the paper.md by Wednesday, that would be great.

kbarnhart:
@mcflugen do these Mac errors make any sense to you?

nicgaspar:
@kbarnhart Looks great Katy. Thanks for including me. Some thoughts, although nothing is pressing, maybe besides adding the one reference.

1. [Philosophical] Why do we say water and ice? Ice is water. No need to comment on that, but it always gets me.
2. Given that this is supposed to be for a non-specialist, I wonder if you need to say "inherent strength and fracture spacing of rock". Do non-specialists think about rock being fractured? This could be simplified to just "Rock properties influence erosion rates by changing the processes responsible for erosion and the rate at which rock is weathered, detached, and turned into mobile sediment."
3. Along with the first reference to the Forte 2016 paper I would also include a reference to the Perne paper. It's very relevant.
@article{perne2017steady,
  title={Steady state, erosional continuity, and the topography of landscapes developed in layered rocks},
  author={Perne, Matija and Covington, Matthew D and Thaler, Evan A and Myre, Joseph M},
  journal={Earth Surface Dynamics},
  volume={5},
  number={1},
  pages={85--100},
  year={2017},
  publisher={Copernicus GmbH}
}
4. I wonder if the comment about Landlab superceeding CHILD might make some people uncomfortable? I think the fact that the earth is not homogeneous and we want Landlab to be able to represent, to some degree, portions of earth, we need layering.
5. Event layers and material layers did not make sense to me when I first started exploring. So I wonder if in the description of event layers it's worth saying event layers, in which the data structure stores each time-step as an event EVEN IF THERE IS NO MATERIAL IN THE LAYER", or "material layers, in which LAYERS REPRESENT MATERIAL OF THE SAME PROPERTY, NOT OF THE SAME AGE".

Again, feel free to ignore my teeny points.

kbarnhart:
@nicgaspar thanks for the comments! The will all improve the `paper.md`. I'll make a commit with improvements. 

Another update on JOSS submission. I brought up some questions about how to handle the details of this submission [here](https://github.com/openjournals/joss/issues/443). 

I will also need to get the docs to build correctly before I can submit. That is currently being addressed by PR #776

kbarnhart:
@nicgaspar I think this is a good point

> [Philosophical] Why do we say water and ice? Ice is water. No need to comment on that, but it always gets me.

My justification for using this phrase has always been that water and frozen water (and water as a gas...) erode rock with very different mechanisms resulting in different landforms.

landlab/components/lithology/litholayers.py
```diff
@@ -0,0 +1,191 @@
...
+
+        >>> lith['K_sp']
+        array([ 0.0001,  0.001 ,  0.0001,  0.001 ,  0.0001,  0.001 ,  0.0001,
+                0.001 ,  0.0001])
+        """
+        self._grid = grid
+
+        function_args = function.__code__.co_varnames
+        if len(function_args) != 2:
+            msg = 'LithoLayers: function must take only two arguments, x and y.'
```
gregtucker: take exactly two
(possible a caller would give 0 or 1)

PR #681, opened 2018-05-03
SiccarPoint/SingerRainfall

SiccarPoint:
This PR implements a LL version of Mike Singer’s STORM code to generate spatially resolved, physically realistic storm sequences through time.

Aspects of this remain somewhat unavoidably linked to MS’s original calibration site at Walnut Gulch. However, the only significant burned in assumption relates to the intensity-duration curves, which are calibrated at the site. Everything else can be modified at will. I believe this is adequate for the time being - a future version could relax this assumption if I can figure out a good way of doing it.

Cautiously optimistic this is pretty robust and more or less ready to go. Testing is largely (and deliberately) via pretty extensive doctests; the same doctests are designed to serve as illustrative examples for the user (hence no explicit tutorial).

Still need to link this into the main documentation. Will do this post-merge as a separate PR.


SiccarPoint:
Assigning @jadams15 on the basis you wrote the uniform_precip component this is broadly based off... @mcflugen also here since Github recommended you ;-)

SiccarPoint:
In addition, what do you think to an experimental submission of this component to the [Journal of Open Source Software](https://joss.theoj.org/about)? Think it would make a nice, self-contained package for review, and let us see what they say about doing this for LL components individually. Re our conversations about credit for coding, would be good to see if this works...?

jadams15:
I put this on my to-do list and will get to it by Monday. We discussed JOSS in an unrelated context during our CSDMS/Delta Force meeting this morning so I am still very new to the journal. I'll look over JOSS before the review and give you my thoughts when I submit the review!

kbarnhart:
@SiccarPoint , I think the submission to JOSS is a great idea. 

I'm planning to submit the Lithology and LithoLayers (PR #674) and @Glader011235 and I just discussed submitting the four or five hillslope related components she made as another submission. 

Lets coordinate timing of submission of this component and the Lithology component. They are very different things, so I assume that they'll be happy to consider them individually. But finding out that they would consider each of the things individually would be a good think to know.

SiccarPoint:
Yes! Presumably we should leave them just as PRs before submission to JOSS? Not clear on this.

I'm happy to try to dig up more info on this stuff if people think it would be helpful.

kbarnhart:
@SiccarPoint , I was planning on leaving it as a PR and leaving the `_citation` field set to something like 'Submitted to JOSS, will be updated before final merge.'

SiccarPoint:
@kbarnhart I'm on board with all this. Hi, future JOSS reviewer skimming back through our discussion!

SiccarPoint:
@jadams15 Thanks for doing this! In reply to your questions:

1. I *believe* the generation is quite happy to proceed ad infinitum without chewing through too much memory. In practice, I think in this version there is a hard-wired limit though, *per year*, which equates to 5000 storms. This effectively results in a catchment area limit (more area- more storms). We maintain two nnodes*5000 arrays, which isn't too bad for memory limit I think. This is a classic trade-off of speed for memory usage. I'll tweak to make this a bit more explicit.

2. Will clean up these.

3. Yeah, this is my debug code. Was keeping it for convenience until close to final version. It'll come out.

4.I agree this method is cleaner...! Mike's code is very tailored to his very specific application, and it hasn't been as straightforward to pull it into a LL-ready version as I'd hoped. Because he basically always runs on the same grid, storms that don't hit any gauges are rare anyway. That wouldn't always be the case for a LL run. The GMD paper is still sat on the editor's desk. I think they're still trying to find a reviewer. As always.

jadams15:
Thanks @SiccarPoint for answering all my questions - it sounds like we aren't doing a merge of this PR until after the manuscripts are submitted, so it sounds like my work here is done. If anything comes up or you need another review for any reason, let me know :)

SiccarPoint:
@nicgaspar @gregtucker I believe this is good to go, and a JOSS submission will still be possible. Given the Singer paper is now out, would be good to get this out the door ASAP (per your email!). 

@jadams15 has completed her review, above - though this has been debugged a bit more since then. Tests are improved, but not totally comprehensive. I think that's OK for now, but I'll tweak them later. Not sure why there's a fail on one of the travis tests, but I don't think it's related to this code...?

SiccarPoint:
I'm going to merge this.

Issue #682, opened 2018-05-03
No tool to make dealing with field, array of at ____, or single value easy


kbarnhart:
Many landlab components support the option of providing an input value as one of three things:
a) String of field name
b) Array of at grid element size
c) Single value (typically int or float). 

This often is implemented by using a series of `isinstance` or `if` `else` statements. 

For example, in the Space component

```python
#K's and critical values can be floats, grid fields, or arrays
if type(K_sed) is str:
    self.K_sed = self._grid.at_node[K_sed]
elif type(K_sed) in (float, int):  # a float
   self.K_sed = float(K_sed)
elif len(K_sed) == self.grid.number_of_nodes:
    self.K_sed = np.array(K_sed)
else:
    raise TypeError('Supplied type of K_sed ' +
                    'was not recognised, or array was ' +
                    'not nnodes long!')
```

Based on conversations with @mcflugen I had previously dealt with this issue by creating a private convenience function that used the `@use_field_name_or_array` decorator. 

For example:

```python

@use_field_name_or_array('node')
def _return_surface(grid, surface):
    """
    Private function to return the surface to direct flow over.

    This function exists to take advantange of the 'use_field_name_or_array
    decorator which permits providing the surface as a field name or array.
    """
    return surface
```

This approach has two problems:
1) I created the convenience function in each place I used it (FlowAccumulator, FlowDirector, NormalFault, flow_direction_dinf.py). 
2) The decorator cannot deal with being passed a float, as is required by space. 

Proposed solution:
1) Create a utility filed called `return_array` that contains convenience functions like the one shown above. At the moment, only one related to nodes is needed, but if ones associated with other grid elements are needed, it could be expanded. 
2) Create a new decorator that can deal with field names, arrays, or single values.

PR #683, opened 2018-05-03
Barnhark/use field name array or float

kbarnhart:
This PR addresses Issue #682 

It creates a new decorator `use_field_name_array_or_value` that is similar to `use_field_name_or array` but can also take as input a single value. If this is the case, that value is broadcast to an array of the correct shape. 

It creates a new utility file called return_array that holds a function  `_return_array_at_node` that uses the new decorator. 

It removes the use the old convenience function in the four places it was previously used. 

I've submitted this PR at this state to see what people think about this approach. 

If we want, it could be merged at this point (with revisions) or extended to apply the `_return_array_at_node` function in other parts of the codebase.


SiccarPoint:
This is very clever. Looks good, but two more general questions spring to mind:
1. Isn't de-facto broadcasting a float into an array every time before doing operations inefficient? I could have this wrong, but I would have thought `myarray * 4.` would be faster than `myarray * array([[4., 4.], [4., 4.]])`...?
2. I don't like the idea of having anything inside the utils folder with a leading underscore getting imported out to anywhere else (i.e. `_return_array_at_node`). If it's code that can and should be called from elsewhere, it shouldn't be flagged as ostensibly private.

Otherwise, conceptually, I like this a lot. I find myself needing to do this quite a bit.

kbarnhart:
thanks for the comments @SiccarPoint 

To answer your questions
2. I have no opinion on this one, so I'm happy to change it to not be private. 
1. I don't know but I suspect @mcflugen might know. 

Some musings on item 1)
My impression is that this function is typically used in the `__init__` phase of creating an instance of and using a component. The values passed as array, field, or int, are typically constants that may or may not be spatially variable. So the broadcast happens once, and then is subsequently used. 

Broadcasting is probably the best in terms of memory (as I understand it, numpy stores only the smallest version of the value, creates a read-only version of the array that may be stored in discontinuous parts of memory, AND that multiple "elements" of the array point to the same bits in memory). How this works in terms of speed is another question... I don't know.

If the thing is temporally variable, they it typically will be passed to init as a field names (e.g. spatially variable runoff in FlowAccumulator needs to be done via the `'water__unit_flux_in'` field, as you can't pass a new `runoff_rate` to the `run_one_step` method). Then this method will just provide a pointer to the field (which is what we do now).  

**FURTHER**
@gregtucker @mcflugen and I had a conversation yesterday about this in which @mcflugen pointed out that this may be a case of supporting too many options rather than the more pythonic "one right way to do something" (@mcflugen correct me if I misrepresented your thoughts). 

However if we had to choose one way only @mcflugen and @gregtucker supported arrays, and I supported field names... And neither of these fix the issue of supporting either single value or at-grid-element-sized-thing.

kbarnhart:
Summarizing the discussion we had about this in the landlab call

1) This approach is fine. 

2) When this function is used, all possible uses need to be tested in the component (even though there are tests of the function itself). This is because the use of the combination of array and/or value may be different within each component. 

3) This change should not be made across all of landlab as part of this pull request. It should be up to individual component authors to make this change across the entire package. 
a) Importantly, @SiccarPoint brings up that there are numerical implications of spatially variable fields. Sometimes upstream, downstream, or mid-point values may be appropriate to use. This is a decision that an individual  component author would need to think about and implement correctly.

SiccarPoint:
(...which is why I never tried stuff like this in a number of places before!)

kbarnhart:
@amanaster2 : @erkanistan recommended that I let you know that I'm developing this. It is not 100% finished, but is close (and should be done in the next few days). It sounds like it might be useful in the work you are doing.

amanaster2:
@kbarnhart Great, thank you!!

mcflugen:
```python
>>> import sys
>>> import numpy as np
>>> x = np.broadcast_to(np.array(5), (3, 3))
>>> sys.getsizeof(x)
112
>>> sys.getsizeof(np.empty((3, 3)))
184
```
So it seems to me that a new array isn't actually created with the scalar copied to every element. Right?

kbarnhart:
@mcflugen Yes, that is my understanding. And the StackOverflow issue ran into memory problems because some slicing resulted in temporary arrays (which we are unlikely to run into).

kbarnhart:
Tests related to masked arrays that have recently cropped up exist (@mcflugen and I just discussed).
Other than that and the one addressed by PR #706 , tests pass, so I'm merging.

Issue #684, opened 2018-05-07
HexModelGrid lacks watershed methods


kbarnhart:
For example 

```python
from landlab.plot import imshow_grid
from landlab import HexModelGrid

hmg = HexModelGrid(5, 4)
z = hmg.add_zeros('node', 'topographic__elevation')
z += hmg.x_of_node + 1.0


hmg.set_watershed_boundary_condition_outlet_id(9, z, nodata_value=-9999)
AttributeError: 'HexModelGrid' object has no attribute 'set_watershed_boundary_condition_outlet_id'

hmg.set_watershed_boundary_condition(z, nodata_value=-9999)
AttributeError: 'HexModelGrid' object has no attribute 'set_watershed_boundary_condition_outlet_id'
```

PR #685, opened 2018-05-07
Barnhark/give hex models watershed methods

kbarnhart:
Addresses Issue #684


kbarnhart:
Looks like other than Appveyor Python 3.5 tests failing while packages are being installed, this at least passes...

kbarnhart:
@mcflugen do you have any thoughts on these apveyor 3.5 tests failing? or do you think this is OK to merge?

PR #689, opened 2018-05-09
Make rock block tweaks

mcflugen:
@kbarnhart a pull request (into your branch) for your consideration.

Based on our quick discussion this afternoon, I made some changes to your branch that simplifies things. By tweaking `deposit_or_erode` in `eventlayers.pyx`, we're able to use it for both `EventLayers` and `MaterialLayers` (so no need for `just_erode`, etc.). In the end there's less repeated code and just less code overall, which is a good thing.

In addition, notice that the tests aren't actually all passing on Travis. If you have a closer look at the output for some of the builds, even ones that get a green check mark, there are still errors reported in the test section. I thought this was a bug I had fixed but apparently not.


kbarnhart:
@mcflugen this looks great. thank you! I'll merge. 

it looks like there is one error on Python 2.7 related to what frozensets look like:
```
File "/home/travis/miniconda/envs/_testing/conda-bld/landlab_1525843103120/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/python2.7/site-packages/landlab/components/lithology/lithology.py", line 591, in landlab.components.lithology.lithology.Lithology.add_rock_type
Failed example:
    lith.ids
Expected:
    frozenset({1, 2, 4, 6})
Got:
    frozenset([1, 2, 4, 6])
```
I'll fix that in my rock block branch.

PR #692, opened 2018-05-13
Gt/hex and ca

gregtucker:
Update CellLab's hex_lattice_tectonicizer to properly shift link and transition info when doing normal-fault offset.


SiccarPoint:
OK, this looks good content wise to me. However, you need more documentation - all the method docstrings (and `__init__`) need "Parameters" sections, and more thorough descriptions of what they actually do. It's very opaque on both fronts, and I'm still not sure I fully grasp the distinction between propid and propdata. Surely only one of these is necessary?

SiccarPoint:
Incidentally, these changes don't resolve the bug in the grain_hill model. I'll raise a separate issue there, with a note for the minimum changes to your code to reproduce it. Will work forward with the mountain_front_model in preference.

SiccarPoint:
Oooh, also: It's not clear whether/how you advect the values around as material gets heaved. Might want to be clearer about that?

Issue #705, opened 2018-05-17
Critical Landlab install issues


SiccarPoint:
I have attempted to downgrade my Python 2 numpy version to 1.13 to enable testing.
```
>>> conda install numpy=1.13
>>> python setup.py develop
```
I can launch Python just fine. However, my Landlab install now appears very broken. An attempt to test, or to even `import landlab`, kicks off:
```
Unable to import landlab. You may not have landlab installed.
Here is your sys.path
/Library/Frameworks/GDAL.framework/Headers
/Library/Frameworks/GDAL.framework/Programs
/Applications/QGIS.app/Contents/Resources/python
/Users/danhobley/anaconda/lib/python27.zip
/Users/danhobley/anaconda/lib/python2.7
/Users/danhobley/anaconda/lib/python2.7/plat-darwin
/Users/danhobley/anaconda/lib/python2.7/plat-mac
/Users/danhobley/anaconda/lib/python2.7/plat-mac/lib-scriptpackages
/Users/danhobley/anaconda/lib/python2.7/lib-tk
/Users/danhobley/anaconda/lib/python2.7/lib-old
/Users/danhobley/anaconda/lib/python2.7/lib-dynload
/Users/danhobley/anaconda/lib/python2.7/site-packages
/Users/danhobley/anaconda/lib/python2.7/site-packages/aeosa
/Users/danhobley/git_landlab
/Users/danhobley/grain_hills
/Users/danhobley/anaconda/lib/python2.7/site-packages
Traceback (most recent call last):
  File "test-installed-landlab.py", line 25, in <module>
    import landlab
  File "/Users/danhobley/git_landlab/landlab/__init__.py", line 37, in <module>
    from .field.scalar_data_fields import FieldError
  File "/Users/danhobley/git_landlab/landlab/field/__init__.py", line 4, in <module>
    from .graph_field import GraphFields
  File "/Users/danhobley/git_landlab/landlab/field/graph_field.py", line 2, in <module>
    import xarray as xr
  File "/Users/danhobley/anaconda/lib/python2.7/site-packages/xarray/__init__.py", line 6, in <module>
    from .core.alignment import align, broadcast, broadcast_arrays
  File "/Users/danhobley/anaconda/lib/python2.7/site-packages/xarray/core/alignment.py", line 10, in <module>
    from . import utils
  File "/Users/danhobley/anaconda/lib/python2.7/site-packages/xarray/core/utils.py", line 13, in <module>
    import pandas as pd
[...]

PR #706, opened 2018-05-17
SiccarPoint/fix issue 702

SiccarPoint:
Cleans up the failing test in line with new docs in the func.


SiccarPoint:
Unable to test this locally; see #705 , but should work fine

Issue #701, opened 2018-05-16
Travis gives a green check when tests don't pass


kbarnhart:
I've seen this happen in #658 and #674. It may have happened elsewhere. 

There seem to be other cases in which it does behave as expected (throws a red x when tests fail). 

I believe that @mcflugen knows about this and is generally updating the testing implementation. 

With CSDMS meeting coming up, I don't anticipate this will be resolved in short order, but the issue mostly exists to remind us all to actually look at the results of the tests on Travis and Appveyor between now and when this is fixed. 

@mcflugen @SiccarPoint @nicgaspar @gregtucker @saisiddu @erkanistan

mcflugen:
I believe #707 fixed this.

PR #707, opened 2018-05-17
Mcflugen/fix conda tests

mcflugen:
This pull request addresses #701 where Travis was reporting that all tests passed when, in fact, they did not. In addition, @kbarnhart noticed that our AppVeyor tests were not being run at all (oops 🤦‍♂️ )! Anyway, that's fixed in this pull request too and now all builds are failing (and being reported as such). Yay!

I'll leave it up to future pull requests to fix the failing tests. They look to be fairly minor.

### DIFF
--- a/.conda-recipe/run_test.py
+++ b/.conda-recipe/run_test.py
@@ -1,4 +1,8 @@
 #! /bin/bash
-
+import sys
 import landlab
-landlab.test()
+result = landlab.test()
+if result.wasSuccessful():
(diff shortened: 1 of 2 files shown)


kbarnhart:
Since we don't expect that the tests will pass on this PR, I'll assess as follows:
do the tests do run AppVeyor? If they fail on AppVeyor and Travis, do we get red xs instead of happy green checks.

And if these things are the case, I'll approve this.

Issue #708, opened 2018-05-17
Cleanup and improvement of SPACE and Erosion Deposition


kbarnhart:
in discussions with @cmshobe and @gregtucker about SPACE and Erosion/Deposition we identified some things:

1. Space and ErosionDeposition should share a common base class. This will make it easier if others care to make similar components that define Es and Er differently.

2. Both components should clean up the inits and the calc_ero_rate function.

3. All different methods of instantiation (and the components overall) should be tested.

4. There are some behaviors that we want that neither component does. Two examples:

a) We should be able to provide an "external" input sediment flux. This would be added to `qs_in`. For example a fluvial input of sediment at a grid boundary (alluvial fan???). 

b) When a node is flooded, how much sediment should be deposited? All of it? Or the amount associated with the deposition rate x dt. 

I'm going to submit a PR very shortly that deals with some of these issues.

cmshobe:
I agree with these changes, with the most important in my mind being the cleaning up of the inits.

On #4b:
In lines 603 and 608 don't we already have a rule for flooded or back-tilted nodes in which sediment is added to the alluvial thickness at a rate of $D_s dt$?

nreitman:
@kbarnhart as we discussed I'm really excited about this: "a) We should be able to provide an "external" input sediment flux. This would be added to qs_in. For example a fluvial input of sediment at a grid boundary (alluvial fan???)." and am willing to help with testing!

kbarnhart:
Thanks @nreitman, I'll work on this and tag you when its ready for tests to be made!

Basalte:
I have seen this discussion about ""external" input sediment flux. I'd like to model evolution of alluvial fan with landlab. I am very new in Landlab and I wonder if thsi option  is now avaliable. If not, how model alluvial fan with landlab ?

kbarnhart:
@Basalte thanks for the question. This functionality is not yet present in either SPACE or ErosionDeposition. Like all of landlab, we would welcome pull requests adding functionality. 

The best way to learn about landlab is to check out the documentation (https://landlab.readthedocs.io/) and the library of tutorials (https://landlab.readthedocs.io/en/master/user_guide/tutorials.html). These can be run locally or through binder.

gregtucker:
[...]

PR #709, opened 2018-05-17
Barnhark/major cleanup to space and erodepo init

kbarnhart:
Addresses Issue #708 

Also tagging @cmshobe.

This cleans up the inits, simplifies the hydrology, and makes a base class. 

It creates a new field `external_sediment__flux` that can be used to supply an external sediment flux. 

Existing tests pass locally on my machine. 

It does NOT: make any new tests, assess if the `external_sediment__flux` method works as expected, or verify that the notebooks and examples from the paper operate as expected. 

Assignments:

Everyone, look at what I did and see what you think. Think about deposition in a flooded node. When a node is flooded, how much sediment should be deposited? All of it? Or the amount associated with the deposition rate x dt.

@gregtucker try out the `external_sediment__flux` approach. Does it work as you think it should?

@cmshobe verify that this works for the space paper examples. Maybe make them into unit tests if they don't take too long to run.

Once we've hammered out the guts of the revised code into something that we all like, we can divy up revising the documentation and adding more tests.


kbarnhart:
@cmshobe I've now merged the changes from PR #719. Have at your next steps.

kbarnhart:
I did a quick check and just saw the 8 masked array errors. Merging. 
thanks for making those tests @cmshobe

Issue #711, opened 2018-05-21
Mapper request


pfeiffea:
Hoping for a new mapper called something like 'map_upwind_node_to_link', which would map from a node to its downstream link. This is for use in the NetworkSedimentTransporter, particularly thinking about taking drainage area from an upstream node and applying it to the downstream link. 

Thanks!!

SiccarPoint:
Hold up, @kbarnhart. I believe we already have a version of this functionality. See `map_max_of_link_nodes_to_link`.

PR #713, opened 2018-05-21
Barnhark/make map upwind node to link

margauxmouchene:
Addresses issue #711:
Create a mapper that assigns a value to a link from its upstream node.

### DIFF
--- a/landlab/grid/mappers.py
+++ b/landlab/grid/mappers.py
@@ -1429,6 +1429,29 @@ def map_link_vector_sum_to_patch(grid, var_name, ignore_inactive_links=True,
 
     return out
 
+def map_upstream_node_to_link(grid, var_name):
+    """Map upstream node to link.
+
+    Based on 'flow__links_to_receiver_nodes'
+
+    Examples
+    --------
+    >>> from landlab import RasterModelGrid, FlowAccumulator
+    >>>
+    """
+    # we want to map the node value to the downstream link.
+
+    # values
+    values = grid['node'][values]
+
+    # link_id downstream of each node.
+    link_to_reciever = grid['node']['flow__links_to_receiver_nodes']
+
+    vals = np.zeros(grid.size('link'))
+    for i in range(len(vals)):
+        upstream_node = np.where(link_to_reciever == i)[0]
+        vals[i] = values[upstream_node]
+
 
 def dummy_func_to_demonstrate_docstring_modification(grid, some_arg):
     """A dummy function to demonstrate automated docstring changes.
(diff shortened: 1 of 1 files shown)


SiccarPoint:
I believe this method duplicates existing functionality. See `map_max_of_link_nodes_to_link`. Consider junking it, unless you're happy it's different enough.

kbarnhart:
@gregtucker and recently discussed this. 

Based on that discussion, I think what we want may be 

`map_value_at_upwind_node_to_link(grid, control_name, value_name)`
and 
`map_value_at_downwind_node_to_link`(grid, control_name, value_name)`

In which upwind vs downwind is determined by `control_name` (so we could use `topographic__elevation` to have the effect of doing this based on flow direction). 

And `value_name` represents the value to map to links. 

We don't currently have these mappers, but they would be generic than what @margauxmouchene and I cooked up yesterday, but would map from nodes to links. 

Also, as links only have one node on either end of them, we don' need to think about the min/max/mean of the upwind/downwind values as we we do for the map upwind link to node-type mappers.

SiccarPoint:
(I believe when these were made we were not using "upwind"/"downwind", for consistency with wider-spread use of "min"/"max" elsewhere in the code base, and also because those terms are clearer if you aren't already familiar with the jargon.)

kbarnhart:
@SiccarPoint , thanks for clarifying, You are right, that function is what we want. 

I will think about how (if at all) I might think about revising the mappers documentation section in order to make sure that I would have been about to figure this out based on the documentation alone).

SiccarPoint:
Yes, we need to work on that! I got here by:
1. Being pretty sure we had such a thing in the first place, then
2. Instantiating a grid, then doing `grid.map_[tab]` in iPython.
This is not a particularly advisable workflow for a user in the wild!

kbarnhart:
@SiccarPoint @margauxmouchene I just had a further thought... Sometimes the elevation differences are not sufficient to map based on flow (e.g. when there is a depression). 

Option A) A user/developer needs to be savvy enough to figure out that they should use `map_value_at_max_node_to_link(control_name, value_name, out=None)` on a pit filled field, or 

Option B) We make something like this that is flow-routing specific. 

I'm pretty split between these two. Do either of you have opinions?

SiccarPoint:
For a grid method, I think we should keep it simple. Maybe add a router-specific alternative as part of the routing tool sets?

kbarnhart:
I'm closing this issue since I think @SiccarPoint resolved it in the spring.

Issue #479, opened 2017-05-01
Try/except error - solved by restarting computer:


amanaster2:
Hi,
I was working with the KinwaveImplicitOverlandFlow component last week and I was getting the following error:
![image](https://cloud.githubusercontent.com/assets/26778806/25597824/e6dcc5de-2e85-11e7-9ac9-99f522673d4b.png)
Restarting my computer solved the error. I figured I should inform the Landlab team of this error in case anyone else has the same problem!

Best,
Amanda Manaster

kbarnhart:
Hi @amanaster2, thanks for making this an issue! We talked about this a bit in our recent developer call. One topic that came up was that this would be hard to ask for a code snippet that would reproduce this error, since you seem to have successfully made it go away. 

A few questions, however, that may be helpful for us and for someone else who may have this problem in the future. 

1) Before this error occurred, did landlab pass all of its internal tests?
2) Did you make any other changes before or after restarting your computer?

amanaster2:
Hi @kbarnhart,
I (somehow) reproduced this error yesterday... It seems that the error pops up after I've run the same bit of code multiple times... But it's not every time, so it's hard to guarantee that that snippet of code would reproduce the error!

To answer your questions:
**1.** Landlab passed all but a few of its internal tests. (The internal tests that it did not pass were related to netCDF4, which I've been having issues with in the most recent version of Python anyways.)
**2.** I uninstalled and reinstalled the developer version of Landlab about 30 minutes before I actually restarted my computer (unsure if that helped or not!), but I did not make any changes after restarting my computer.

kbarnhart:
Sorry to hear it came back! But it sounds like you were able to reproduce your restart fix, yes?

Re-running ```python setup.py develop``` often fixes problems (especially if a recent update needs cython compiling), but its hard to tell if that is exactly what solved your problem. 

I'd recommend getting landlab to a place where it has no test errors and then if the error comes back, trying to make a bit of code that reproduces it, and posting that snippet. 

if you are using conda, 
```
conda install --file=requirements.txt
```
followed by ```python setup.py develop``` should fix the netcdf issue. Let me know if it doesn't.

saisiddu:
@kbarnhart 
Just writing my observations as I was trying to help her on this error: I don't think this error was solved due re-compiling. Her netCDF4 errors seems to be due to her installation of Python 3.6. I remember her saying that few people are seeing this error due to recent upgrade to netCDF4. 
[...]

PR #716, opened 2018-05-31
Replaced try/except with if

amanaster2:
Issue #479 is addressed

### DIFF
--- a/landlab/components/flow_accum/flow_accumulator.py
+++ b/landlab/components/flow_accum/flow_accumulator.py
@@ -715,9 +715,7 @@ class FlowAccumulator(Component):
                 assert runoff_rate.size == grid.number_of_nodes
 
         # test for water__unit_flux_in
-        try:
-            grid.at_node['water__unit_flux_in']
-        except FieldError:
+        if 'water__unit_flux_in' not in grid.at_node:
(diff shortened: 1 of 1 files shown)


saisiddu:
@kbarnhart and @mcflugen: Amanda and I edited 'next' branch to address issue #479. We didn't see any change in the number of failures (after making edits). We did run local tests to confirm this. Can you please confirm that this pull request is good for merging? Thanks!

kbarnhart:
@saisiddu As I understand the Landlab git workflow, you'll want to branch off of `master` and merge into `master`. Can you please do that instead?

`next` appears to be a branch that @mcflugen is using for testing. I might propose that we get rid of it for confusion (e.g. use `mcflugen\next` for such things. 

We might also consider making the git workflow for the tutorials branch and the main code base branch the same (e.g. rename `master` in this repo to `next`) @mcflugen what do you think?

mcflugen:
@kbarnhart you're correct. @saisiddu please branch off of `master`, not `next`. I created `next` to set up our workflow like it is in our tutorials repository but then decided against it. I should have removed it. Sorry for the confusion!

@kbarnhart I am still all for renaming `master` to `next`, though. If people are ok with it.

Issue #718, opened 2018-06-01
SPACE needs porosity (phi) in cfuncs sediment flux calculation


cmshobe:
@kbarnhart @gregtucker I believe that erosion of sediment Es should be multiplied by (1-phi) in the sediment flux function but isn't. Can someone sanity check me on this?

Issue #717, opened 2018-06-01
SPACE cfuncs file has incorrect cell area value for qs calculation


cmshobe:
@kbarnhart @gregtucker  I apologize for not noticing this sooner, but I think the Qs calculation in the cfuncs file is incorrect. The sed flux needs to be calculated using the area of a cell, NOT the square of link length. When a diagonal pops up the Qs value is then incorrect. This area value should always be dx**2 for a rectangular grid, never diag_length^2. It's important that we fix this before doing any further testing or modifications, but I'm not sure how to fix it and also keep the hex grid compatibility. @kbarnhart any ideas?

PR #719, opened 2018-06-04
Fix cell area and porosity in Space and Erosion Deposition

kbarnhart:
addresses Issues #717 and #718 
Tagging @cmshobe  as well. 

As part of the main space overhaul we plan to make more comprehensive  tests


kbarnhart:
Failures look like they are associated with not being able to get to the point of installing landlab. Spot check of ~4 of the passes look like they just have the 8 known failures related to masked errors.

@gregtucker OK to merge?

PR #720, opened 2018-06-04
Gt/ca top hit bug

gregtucker:
Fix "top hit" bug in hex_lattice_tectonicizer, which manifested in GrainHill as a freeze that sometimes occurred after nodes with rock or sediment state entered the top row of the lattice. The fix is simply to not have any changes to the top row nodes (which is consistent with the CellLab philosophy of leaving boundary states unchanged through time).


gregtucker:
CI test issues seem to be conda-related rather than substantive, so I am merging this.

PR #722, opened 2018-06-07
Adding 3 steady-state tests for SPACE (from the GMD paper examples)

cmshobe:
1) detachment-limited solution (slope-area model vs analytical)
2) transport-limited solution (slope-area model vs analytical)
3) mixed bedrock-alluvial solution (slope-area and sediment depth model vs analytical)

ALSO enhanced documentation for 'dicharge_field' param
ALSO removed qs_in_ext capability because @kbarnhart and I agreed to deal with that in a separate PR.

@gregtucker and @kbarnhart please review!

### DIFF
--- a/landlab/components/space/tests/test_space.py
+++ b/landlab/components/space/tests/test_space.py
@@ -1,8 +1,70 @@
 from landlab import RasterModelGrid, HexModelGrid
 from landlab.components import Space, FlowAccumulator
 import numpy as np
-from numpy.testing import assert_equal
+from numpy import testing
 
+def test_matches_detachment_solution():
+    """
+    Test that model matches the detachment-limited analytical solution
+    for slope/area relationship at steady state: S=(U/K)^(1/n)*A^(-m/n).
+    """
+    
+    #set up a 3x3 grid with one open outlet node and low initial elevations.
+    nr = 5
+    nc = 5
+    mg = RasterModelGrid((nr, nc), 10.0)
+    
+    z = mg.add_zeros('node', 'topographic__elevation')
+    br = mg.add_zeros('node', 'bedrock__elevation')
+    soil = mg.add_zeros('node', 'soil__depth')
+
+    mg['node']['topographic__elevation'] += mg.node_y / 10000 \
+        + mg.node_x / 10000 \
+        + np.random.rand(len(mg.node_y)) / 10000
+    mg.set_closed_boundaries_at_grid_edges(bottom_is_closed=True,
+                                           left_is_closed=True,
+                                           right_is_closed=True,
+                                           top_is_closed=True)
+    mg.set_watershed_boundary_condition_outlet_id(0,
+                                                  mg['node']['topographic__elevation'], 
+                                                  -9999.)
+    br[:] = z[:] - soil[:]
+        
+    # Create a D8 flow handler
+    fa = FlowAccumulator(mg, flow_director='D8')
+    
+    #Instantiate DepressionFinderAndRouter
+    
+    # Parameter values for detachment-limited test
+    K_br = 0.01
+    U = 0.0001
+    dt = 1.0
+    F_f = 1.0 #all detached rock disappears; detachment-ltd end-member
+    m_sp = 0.5
+    n_sp = 1.0
+
+    # Instantiate the Space component...
+    sp = Space(mg, K_sed=0.00001, K_br=K_br,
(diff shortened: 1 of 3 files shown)


landlab/components/space/tests/test_space.py
```diff
@@ -1,8 +1,234 @@
...
+from numpy import testing
 
+def test_matches_detachment_solution():
+    """
+    Test that model matches the detachment-limited analytical solution
+    for slope/area relationship at steady state: S=(U/K_br)^(1/n)*A^(-m/n).
+    """
+    
+    #set up a 3x3 grid with one open outlet node and low initial elevations.
+    nr = 5
```
kbarnhart: nitpick comment. comment should say 5x5

landlab/components/space/tests/test_space.py
```diff
@@ -1,8 +1,234 @@
...
+    
+def test_matches_transport_solution():
+    """
+    Test that model matches the transport-limited analytical solution
+    for slope/area relationship at steady state: S=((U * v_s) / (K_sed * A^m) 
+    + U / (K_sed * A^m))^(1/n).
+    """
+    
+    #set up a 3x3 grid with one open outlet node and low initial elevations.
+    nr = 5
```
kbarnhart: ditto comment above 5x5

landlab/components/space/tests/test_space.py
```diff
@@ -1,8 +1,234 @@
...
+                         sp_crit_br=0)
+
+    # ... and run it to steady state (4000x1-year timesteps).
+    for i in range(4000):
+        fa.run_one_step()
+        flooded = np.where(df.flood_status==3)[0]
+        sp.run_one_step(dt=dt, flooded_nodes=flooded)
+        br[mg.core_nodes] += U * dt #m
+        soil[0] = 100.
+        z[:] = br[:] + soil[:]
```
kbarnhart: I don't think I understand why you need to enforce the outlet node soil depth here but not in the detachment case. could you add a comment that states why.

landlab/components/space/tests/test_space.py
```diff
@@ -1,8 +1,234 @@
...
+    for slope/area relationship at steady state: 
+    S=((U * v_s * (1 - F_f)) / (K_sed * A^m) + U / (K_br * A^m))^(1/n).
+    
+    Also test that the soil depth everywhere matches the bedrock-alluvial
+    analytical solution at steady state: 
+    H = -H_star * ln(1 - (v_s / (K_sed / (K_br * (1 - F_f)) + v_s)))
+    """
+    
+    #set up a 3x3 grid with one open outlet node and low initial elevations.
+    nr = 5
```
kbarnhart: woooo. 5x5

landlab/components/space/tests/test_space.py
```diff
@@ -1,8 +1,234 @@
...
+        + np.random.rand(len(mg.node_y)) / 10000
+    mg.set_closed_boundaries_at_grid_edges(bottom_is_closed=True,
+                                           left_is_closed=True,
+                                           right_is_closed=True,
+                                           top_is_closed=True)
+    mg.set_watershed_boundary_condition_outlet_id(0,
+                                                  mg['node']['topographic__elevation'], 
+                                                  -9999.)
+    soil[:] += 0. #initial soil depth of 100 m
+    br[:] = z[:]
```
kbarnhart: comment should no soil.

landlab/components/space/tests/test_space.py
```diff
@@ -1,8 +1,234 @@
...
+                         sp_crit_br=0)
+
+    # ... and run it to steady state (4000x1-year timesteps).
+    for i in range(10000):
+        fa.run_one_step()
+        flooded = np.where(df.flood_status==3)[0]
+        sp.run_one_step(dt=dt, flooded_nodes=flooded)
+        br[mg.core_nodes] += U * dt #m
+        soil[0] = 0.
+        z[:] = br[:] + soil[:]
```
kbarnhart: similar to above, I don't totally understand this boundary condition. Why not enforced at the analytical steady state soil depth?

I see that you are comparing only the core nodes, but the slope downstream impacts qs-out (via Es and Er), so I'm having a hard time figuring out why it wouldn't matter.

kbarnhart:
A thought... Should we put in  an optional test a timestep checker (that can be turned on to see if the numerical solution is being overshot with the basic solver).

cmshobe:
OK @kbarnhart this should be good to go!

Issue #714, opened 2018-05-21
Additional fields created by FlowDirection


kbarnhart:
Right now FlowDirectors create two at-node fields `flow__receiver_node` and `flow__link_to_receiver_node`. Ideally, they would have an option to create other fields. 

A comprehensive list of fields might be:

@ link
`flow__direction` (-1, 0, 1) # indicates if flow direction is the same as topology direction
`flow__upstream_node`
`flow__downstream_node`
`flow__upstream_link_at_link`
`flow__downstream_link_at_link`

@ node:
`flow__link_to_receiver_node`
`flow__link_to_upstream_node`
`flow__receiver_node`
`flow__upstream_node`

SiccarPoint:
`link_dirs_at_node` was designed so if you sum `[fluxes_on_links] * mg.link_dirs_at_node` for a given set of `links_at_node`, you get a net flux *INTO* the node. i.e., a negative number is assigned to any link pointing out of the node. So this means that where the node sits at the link TAIL, it gets a -1. I think this definition is consistent with what you have in parentheses, but not what you have outside them...???

Regardless, your proposed new definition 100% makes sense. It's just the sign on the fluxes that would be defined by the subsequent flow accumulation, right? I note this seems to be a fundamentally different thing from `link_dirs_at_node`: those are to do with link-node topology, but `flow__link_direction` is inherently a process definition...? (i.e., you could make a distributary network out of a catchment drainage by flipping `flow__link_direction`, but this would make zero difference to `link_dirs_at_node`)

kbarnhart:
@SiccarPoint you are right. I reversed what is in the parentheses. Convention should be:

* 0 = No Flow
* -1 = Flow is going AGAINST the topology (flow going from head node to the tail node)
* 1 = Flow is going WITH the topology (flow going from the tail node to the head node)

Thanks for thinking about it carefully!

nicgaspar:
@kbarnhart 
`flow__link_direction` sounds super helpful to me. I'd love if @jadams15 could chime in so let's give her a minute to respond too. She has thought about this SO much.

jadams15:
I think there is a lot of benefit to having this - as Dan describes - as a process definition. As Nicole notes, I spent a lot of time thinking about ``link_dirs_at_node`` and often struggled to explain to new users how the link direction could be used to get at total discharge. It makes complete sense to me that for the network model you would move away from link direction and instead have a function for flow direction.

[...]

PR #725, opened 2018-06-10
Barnhark/flow adjacency info

kbarnhart:
This branch addresses Issue #714 

Its not quite done.


kbarnhart:
This branch now also addresses #860.

kbarnhart:
This PR accomplishes a few things:
Mostly, it adds functionality requested by #714 (as discussed with @pfeiffea).

In FlowAccumulator it adds a few new properties: 
- link_upstream_order : the order of links based on the stack
- headwater_nodes : the nodes that have no upstream nodes and contribute flow.

In FlowDirectorSteepest it adds the ability to determine if flow is going with or against the topologic link direction (in the field flow__link_direction). Based on this field there are a few new properties
- flow__link_direction_at_node
- flow__link_incoming_at_node
- flow__link_direction
- upstream_node_at_link
- downstream_node_at_link
These were developed to support work on the NetworkSedimentTransporter.

You might wonder why this only exists for FlowDirectorSteepest. It is because it is not trivial to do this right with diagonals and links (and for to-one and to-many methods). Not that it can't be, but I didn't want to go through the pain of creating the functionality if no one needed it (and @pfeiffea only needs it for Steepest).

In testing this last example I determined that DepressionFinderAndRouter does not update the links correctly if there is a lake of only one node. So I fixed this and have a test. 

In order to make NetworkModelGrid fully work with FlowAccumulator, it adds "grid" as a location in GraphFields.

It also deals with the deprecation warnings about FlowRouter from #860 and some lingering xy_spacing deprecation warnings in DrainageDensity.

kbarnhart:
@gregtucker @mcflugen @SiccarPoint could one of you review this? I summarized this PR in the comment above.

landlab/components/flow_accum/flow_accumulator.py
```diff
@@ -548,6 +550,25 @@ class DepressionFinderAndRouter to the parameter `depression_finder`.
...
+    this type of grid, an at-node array called ``cell_area_at_node`` must be
+    present.
+
+    >>> from landlab.grid.network import NetworkModelGrid
+    >>> y_of_node = (0, 1, 2, 2)
+    >>> x_of_node = (0, 0, -1, 1)
+    >>> nodes_at_link = ((1, 0), (2, 1), (3, 1))
+    >>> nmg = NetworkModelGrid((y_of_node, x_of_node), nodes_at_link)
+    >>> area = nmg.add_ones('node', 'cell_area_at_node')
+    >>> z = nmg.add_field('topographic__elevation',
```
mcflugen: Why not use black formatting in docstrings too?
```python
    >>> z = nmg.add_field(
    ...     'topographic__elevation',
    ...     nmg.x_of_node + nmg.y_of_node,
    ...     at = 'node',
    ... )
```

landlab/components/flow_accum/flow_accumulator.py
```diff
@@ -732,6 +753,70 @@ def node_order_upstream(self):
         """Return the upstream node order (drainage stack)."""
         return self._grid["node"]["flow__upstream_node_order"]
 
+    @property
+    def link_order_upstream(self):
```
mcflugen: Since this isn't something being saved in the grid, I wonder if it should be a function rather than a property.

landlab/components/flow_accum/flow_accumulator.py
```diff
@@ -732,6 +753,70 @@ def node_order_upstream(self):
...
+        array([ 5, 14, 10,  6, 11,  7, 23, 19, 15, 20, 16, 28, 24, 29, 25])
+        """
+        downstream_links = self._grid["node"]["flow__link_to_receiver_node"][
+            self.node_order_upstream
+        ]
+        out = downstream_links.flatten()
+        return out[out != BAD_INDEX_VALUE]
+
+    @property
+    def headwater_nodes(self):
```
mcflugen: Same comment as with `link_order_upstream`.

landlab/components/flow_director/tests/test_flow_director.py
```diff
@@ -379,3 +379,18 @@ def test_change_bc_post_init():
         ]
     )
     assert_array_equal(new_true_reciever, fd.receiver)
+
+
+def test_link_flow_direction_hex():
```
mcflugen: Are you just trying to pad our testing stats?

landlab/components/flow_routing/lake_mapper.py
```diff
@@ -1139,16 +1133,44 @@ def _route_flow(self):
...
+                        outlet_receiver
+                        == self._grid.adjacent_nodes_at_node[outlet_node, :]
+                    )
+                    new_link = self._grid.links_at_node[outlet_node, find_recs]
+
+                if new_link.size == 0:
+                    new_link = LOCAL_BAD_INDEX_VALUE
+                # print('outlet_receiver', outlet_receiver)
+                # print('outlet_node', outlet_node)
+                # print('new_link', new_link)
```
mcflugen: Remove debug print statements.

landlab/components/flow_director/flow_director_steepest.py
```diff
@@ -155,6 +260,18 @@ def __init__(self, grid, surface="topographic__elevation"):
...
+        if "flow__link_direction" not in self._grid.at_link:
+            self._flow__link_direction = grid.add_field(
+                "flow__link_direction",
+                grid.zeros(at="link", dtype=int),
+                at="link",
+                dtype=int,
+            )
+        else:
+            self._flow__link_direction = grid.at_link["flow__link_direction"]
+
```
mcflugen: I think you're using the name "*flow__link_direction*" to be compatible with existing names, is that right? Since it is not a CSDMS Standard Name, I think we should drop the double underscore. In my opinion, it just confuses things. The same goes for "*flow__link_direction_at_node*" and "*flow__link_incoming_at_node*".

In addition, I don't think we should *ever* use double underscores in python variable names (i.e. `self._flow__link_direction`) even if they are proper Standard Names.

Should the data type for *flow__link_direction* be `int`? Could it, instead, be `int8`?

landlab/components/flow_director/flow_director_steepest.py
```diff
@@ -228,13 +345,330 @@ def direct_flow(self):
...
+               [ 0,  0,  0,  0],
+               [ 0,  0,  0,  0],
+               [ 0,  0,  0,  0],
+               [ 0,  0,  0,  0],
+               [ 0,  0,  0,  0],
+               [ 0,  0,  0,  0]])
+
+        Next with DepressionFinderAndRouter:
+
+        >>> mg2 = RasterModelGrid(5,5)
```
mcflugen:
```python
        >>> mg2 = RasterModelGrid((5, 5))
```

mcflugen:
👍 @kbarnhart Looks good.

landlab/components/flow_accum/flow_accumulator.py
kbarnhart: Can do. One thing to note, however, as we've been starting to be much better about lint... Is that ([to the best of my knowledge](https://github.com/ambv/black/issues/144)) black does not yet support formatting docstrings... @mcflugen correct me if you know better.

But there is something called [docformatter](https://github.com/myint/docformatter) which does. I've tried it out with mixed success.

landlab/components/flow_accum/flow_accumulator.py
kbarnhart: This makes sense. Will do.

landlab/components/flow_accum/flow_accumulator.py
kbarnhart: Ditto above. Will do.

landlab/components/flow_director/tests/test_flow_director.py
kbarnhart: nope! I think that was an old test that I outlined but never wrote... removing.

landlab/components/flow_director/flow_director_steepest.py
kbarnhart: This has more to do with Issue #864, right? I assume this means you prefer spacing.

landlab/components/flow_director/flow_director_steepest.py
kbarnhart: Good point about the names. Will change. 

Also, I think it could be int8 for sure!

landlab/components/flow_accum/flow_accumulator.py
kbarnhart: My attempt at the middle path resulted in changing the formatting for FlowAccumulator, and all FlowDirectors... Definitely looks nicer!

landlab/components/flow_accum/flow_accumulator.py
kbarnhart: I also determined that a few other new weren't really properties (so I fixed them too).

landlab/components/flow_director/flow_director_steepest.py
kbarnhart: I got rid of them in all landlab/components/flow_* folders... It will make finding and removing deprecation warnings easier when I eventually deal with #864.

PR #745, opened 2018-07-10
SiccarPoint/fix issue 744

### LINKED ISSUE #744: Negative runoffs/infiltration
Not sure whether to label this as a bug or an opportunity:

At present the runoff accumulator quite happily takes negative numbers as its ``runoff`` argument, or as `water__unit_flux_in`. If you do this, the output permits negative flow accumulations, which is clearly nonsense.

If we treat this as a pure bug, there needs to be an input check that runoff >= 0., and a test.

BUT

Better surely to treat this as an opportunity to implement simple infiltration/transpiration losses - which I don't believe is currently something we can do, right? So, keep the option for runoff < 0., but ensure that surface_water__discharge >= 0. at all nodes, as it's calculated. And appropriate test.

Seems like a simple fix either way, presumably modifying the Cython loop in the second case. Happy to do this myself, though want @kbarnhart 's input beforehand - and you may well be able to do it faster, too.

SiccarPoint:
This PR fixes #744 by forbidding negative discharge but still permitting negative runoff, i.e., transmission losses.

Tests to be added once approval from @kbarnhart is obtained! Qualitative testing indicates the new approach is working fine.


SiccarPoint:
This also appears to break some tests. Think I now understand better how this is set up. Retracting pull.

SiccarPoint:
This now works. I was being a total moron before. Three tests now fail, but these are floating point errors easy enough to clean up. New tests still needed.

kbarnhart:
I double checked a few of the failing travis builds and I can confirm that the errors are floating point ones as @SiccarPoint indicated

SiccarPoint:
From my perspective this is now good to go, but there are some failing tests for 32 bit machines only. This appears to be related to the Cython, and the interpretation of long and double formats.It fails with `ValueError: Buffer dtype mismatch, expected 'DTYPE_INT_t' but got 'long long'` in the cython.

Ideas to resolve this, @mcflugen? It's clearly to do with the declaration of DTYPE_INT, but I don't know enough about Cython to make this architecture agnostic...

SiccarPoint:
Bump @mcflugen ?

mcflugen:
@SiccarPoint I think somewhere along the way one or both of the arrays `s`, and `r` were created with the wrong data type (probably `int64` instead of `int32`). This is usually caused by an array being created using `np.where`. On 32 bit systems `np.where` will return an array of 64 bit integers, even though regular ints on that system are only 32 bit. I created the function `as_id_array` (located in `landlab.utils`) to help with this problem.

If you did something like `s = as_id_array(s)` before calling the cython function, you would probably fix the error but you would most likely be creating a copy of the array on those platforms that are seeing the current error.

Issue #750, opened 2018-07-11
Qs in external for SPACE and ErosionDeposition


kbarnhart:
as discussed in Issue #708 with @gregtucker @cmshobe and @nreitman

cmshobe:
Is there a possibility of building this as some sort of "utility" component? I.e. Not adding this functionality directly to the SPACE and ED classes? I think that a) that would keep things cleaner b) keep SPACE more similar to its published form and c) it would more easily allow for expansion to other components such as sediment flux dependent river incision or any future transport components.

kbarnhart:
@cmshobe check out the changes in the PR, they are pretty clean and minimal. 

More importantly, I don't think its possible to accomplish this unless the baseline value for qs_in (e.g. before the calculate_qs_in step) is set _within_ the current component. For that reason, its not clear to me that this can be done with an external component. 

I think that you have a good point regarding thinking about other sediment-flux dependent components. But by using good field names, I think we will accomplish that.

SiccarPoint:
I already do this in the SedDepEroder. Agree it's going to be necessary to embed it in the component. My solution was to "deposit" the sediment in transport after each timestep, then remobilise it at the start of the next. Will check your modifications a bit later to see if they're similar! My field names are:
```
        'topographic__elevation',
        'channel_sediment__depth',
        'channel__bed_shear_stress',
        'channel_sediment__volumetric_transport_capacity',
        'channel_sediment__volumetric_discharge',
        'channel_sediment__relative_flux',
        'channel__discharge'
'''
where channel_sediment__depth is the key one for controlling external sediment inputs (i.e., it gets dumped into the node from e.g., hillslopes, then we mobilise as much of that as we can when the SDE run method gets called).

PR #751, opened 2018-07-11
Barnhark/space qs in external

kbarnhart:
Starting to address #750 

Not yet ready for review or tests


kbarnhart:
@nreitman you can now have at making some analytical solution test for the modified version of SPACE and ErosionDeposition

PR #752, opened 2018-07-16
Use pytest for testing

mcflugen:
This pull request primarily switches our tester from `nose` to `pytest`. Since `nose` is no longer being developed and `pytest`, generally, seems to be better, it seemed like a good idea to switch.

The highlights:
* Use `pytest` instead of `nose`
* Tests pass with numpy 1.14 as well as lower versions
* Fixed coverage reporting
* Fixed a failing doctest in `diffuse_by_discharge` (it seems that this test must not have even been run previously?)
* Fixed a warning caused by iterating over a pandas `Dataset`


kbarnhart:
@mcflugen NICE WORK!

mcflugen:
@kbarnhart Hmm. Something's not quite right. I've already made those changes (`setup.py` and `setup.cfg`). Let me have a look, maybe some changes were inadvertently undone?

kbarnhart:
@mcflugen, you are right the changes are there. I missed them in the sea of changed files.

mcflugen:
@kbarnhart Do you think this is ready for a merge?

I made a few additional changes:
* There are now no references to *nose*.
* I removed the entire old testing framework (including the `test-installed-landlab` script).
* Replaced the yield tests, which are supported by *nose* and the current version of *pytest*, but will not be supported in the future by *pytest*. These now use test parameterization provided by *pytest*.
* Updated references to `nose` in the docs.

margauxmouchene:
FYI, I just modified the `Develop your own component page <https://github.com/landlab/landlab/wiki/Develop-your-own-component>`_ to mention pytest and not nose. I am not sure about the pytest vs. py.test mentioned by @kbarnhart

mcflugen:
Thanks @margauxmouchene! `py.test` and `pytest` are the same thing. Since version 3, the recommended usage is `pytest`. `py.test` is still supported and there are no plans to remove it, but sometime in the future, it could be removed.

landlab/components/sink_fill/tests/conftest.py
```diff
@@ -0,0 +1,162 @@
+import pytest
+import numpy as np
```
kbarnhart: ~with a filename of conftest.py is this discoverable? Or is there somewhere I missed that indicated that this file should be tested?~

kbarnhart: never mind. I've now learned what conftest.py. I've left my comment and this link to info

https://docs.pytest.org/en/2.7.3/plugins.html?highlight=re

so that if others have the same questions  they might find this.

Issue #774, opened 2018-08-21
field duplication in route-to-many FlowDirectors


SiccarPoint:
I think I've found a bug in the route-to-many FlowDirectors. By definition these produces arrays of outputs that are 2D, right? For FlowDirectorDINF at least, this gets saved in 'flow__receiver_node**s**'.

(I can see the naming logic of this, but I also don't like it, as it's inconsistent with the naming of the route-to-one routers, and since you can invoke a single FlowDirector or FlowAccumulator, I would say the fields produced by these should remain the same no matter what router you're using...)

That aside though, the actual problem is that somewhere along the line (during inheritance somewhere??), at least the FlowDirectorDINF is producing **both** 'flow__receiver_nodes' AND 'flow__receiver_node'. This is not right, surely; it should be one or the other. Haven't checked other route-to-many ones, but could be there too. Try:

```
>>> mg = RasterModelGrid((5, 5), dx=1.)
>>> mg.add_zeros('node', 'topographic__elevation', dtype=float)
>>> fd = FlowDirectorDINF(mg, 'topographic__elevation')
>>> fd.run_one_step()
>>> 'flow__receiver_node' in mg.at_node.keys()
True
>>> 'flow__receiver_nodes' in mg.at_node.keys()
True
```

This affects necessary type testing in components that only work with route-to-one routers.

kbarnhart:
@SiccarPoint a quick response (again from the Munich airport).

This was intentional. Whether or not it is a bug, I'm not sure.

If you look at the docstrings [for MFD for example](https://landlab.readthedocs.io/en/latest/landlab.components.flow_director.html#landlab.components.flow_director.FlowDirectorMFD) you can see that the fields created are listed and include both fields. 

I did this because multiple flow direction algorithms need a (n-nodes, number of receivers) size array but expectation that 'flow__receiver_node' has a shape of (n-nodes,) for was very entrenched in other components (e.g. the lake mapper).

Looking briefly at other components that use the field `flow__receiver_node`, they may not work correctly with route-to-multiple components (e.g. I'm not sure if the stream power eroders only erode the steepest link or all active links). About to board the next flight so I'll check in about this later (at worst a week from today, after vacation and fieldwork).

SiccarPoint:
[...]

PR #775, opened 2018-08-22
Address Issue 774

kbarnhart:
This PR starts to address Issue #774 

I've removed the duplicated field names and the tests pass (locally). 

There are a few things that still need to be done (@SiccarPoint feel free to add commits to this PR if you are so inspired). 

As best I can tell, three fields that matter here are: flow__receiver_node, flow__link_to_reciever_node, and topographic__steepest_slope.

The two big things that I think we need to do are:

a) write the tests for the new NotImplementedErrors I've made (this is easy, I just didn't get a chance)

b) identify all other components that use these three fields, and either:
    - write a test that demonstrates that they can work with a route to multiple
    - raise a NotImplementedError and write a test for that. 

This one is harder and will take some consistent checking across the code base.

I propose that we brainstorm what all the components are first and then move forward from there.


SiccarPoint:
No worries - this isn't causing any crucial problems, so we can happily kick this down the road.

A little light searching indicates the key components to look at will be:
`draininage_density`
`erosion_deposition`
`generalised_erosion_deposition`
`lake_mapper`
`lake_mapper_barnes`
`space`
`channel_steepness`
`stream_power`
`fastscape_stream_power`
`stream_power_smooth_threshold`
`sed_flux_dep_incision`
`transport_length_hillslope_diffusion`

And the various `flow_accumulator`s and `flow_director`s (obviously). Utils are:

`channel_profile`
`drainage_plot`
`flow_distance`
`source_tracking_algorithm`
`watershed`

I suspect the fast solution here might be to by default introduce the assertions, unless we (realistically, you, as designer of the route-to-manys?) know for a fact the component is happy with route-to-many. AFAIK they are all guaranteed to work with one-to-one, and we can relax this constraint piecemeal later.

SiccarPoint:
OK, this looks great. Only comment might be that you've added a couple of helper properties in the flow directors as part of this (viz. `node_steepest_slope, and `link_to_flow_receiving_node`), and I don't believe (?) they're tested. Clearly these are trivial, but it's lines we aren't covering as-is... I suspect the same probably applies to their adjacent, existing friends (e.g. `proportions_of_flow`).

I assume you have the failing tests here in hand.

Good to go when you are.

kbarnhart:
Thanks @SiccarPoint! I'll add those tests (and merge if all passes).

The travis errors seem to have to do with issue #782 and should be resolved on this branch soon.

kbarnhart:
@SiccarPoint I took this opportunity to bring the `flow_accum` and `flow_director` component submodules to 100% coverage. Thanks for pointing out the coverage gap.

Issue #734, opened 2018-06-20
Docs broken due to new sphinx


kbarnhart:
The RTFD docs error with the error described [here](https://github.com/rtfd/readthedocs.org/issues/4057).

I've fixed this in TerrainBento [here](https://github.com/TerrainBento/terrainbento/pull/32).  

I can fix it the same way for Landlab... but the Landlab docs are a bit more complicated than TerrainBento's and I wasn't the one to set them up. So I want to make sure its OK to fix in the same way before I go ahead and do it. 

@mcflugen @SiccarPoint @gregtucker @jennyknuth feel free to weigh in.

gregtucker:
Looks like a simple fix re: versioning. Go for it!

kbarnhart:
In a landlab call we punted on deciding how to fix this. The short story is that the issue is with `numpydocs` which is no longer supported. Instead one can used `napoleon`. This was easy for me to change in terrainbento, but it was unclear if it would be so simple to transition away from `numpydocs` in landlab. 

However, Issue #762 has occurred because the docs are broken so we should make a decision and fix this.

SiccarPoint:
Not clear to me what the problem might be... What are the key differences? Nothing I've built differs from "standard" Sphinx/rtd styling, so hopefully no problems.

kbarnhart:
I've just been asked about the functionality of the `channel_profile` plotting utility as the docs still list `plot_channels_in_map_view` but the master branch doesn't support this.

kbarnhart:
This has been fixed by PR #776. @mcflugen I think it makes sense for us to make a release so that both the `release` and the `master` docs are working.

What do you think?

Then we can close this issue.

kbarnhart:
Fixed now in both `release` and `master`. closing

PR #776, opened 2018-08-27
Barnhark/fix broken docs

kbarnhart:
This PR addresses Issue #734 

Summary of things this does:
- removes numpydoc
- uses napoleon instead of numpydoc in building docs
- fixes two bits of the `build_grid_func_docs.py` file that had python 2/3 string/byte issues.
- turns building docs back on in travis

On my computer docs still don't build correctly and `make html` ends with this error. @mcflugen any ideas?

```
/Users/barnhark/projects/landlab/landlab/landlab/graph/object/at_node.py:docstring of landlab.graph.object.at_node.get_links_at_node:: WARNING: more than one target found for cross-reference 'Graph': landlab.graph.graph.Graph, landlab.graph.Graph

Exception occurred:
  File "/anaconda3/lib/python3.6/site-packages/sphinx/domains/cpp.py", line 3543, in matches
    if s.identOrOp != identOrOp:
AttributeError: 'Symbol' object has no attribute 'identOrOp'
The full traceback has been saved in /var/folders/d_/zvv65dd52_g_k8xht41v676r0000gn/T/sphinx-err-hctxp_pm.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [html] Error 2
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
 [submodule "docs/scipy-sphinx-theme"]
 	path = docs/scipy-sphinx-theme
 	url = https://github.com/scipy/scipy-sphinx-theme
-[submodule "docs/sphinxext"]
-	path = docs/sphinxext
-	url = https://github.com/numpy/numpydoc
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,8 +59,7 @@ script:
   fi
 - |
   if [[ "$BUILD_DOCS" == "1" ]]; then
-    echo "WARNING: NOT BUILDING DOCS"
-    # (cd docs && make html)
+    (cd docs && make html)
   fi
 - |
   if [[ "$DEPLOY" == "1" ]]; then
--- a/docs/build_grid_func_docs.py
+++ b/docs/build_grid_func_docs.py
@@ -71,7 +71,7 @@ def create_dicts_of_cats():
  fails_allgrid) = create_dicts_of_cats()
 
 for grid_to_modify in grid_name_to_class.keys():
-    f = open('./text_for_' + grid_to_modify + '.py.txt', "rb")
+    f = open('./text_for_' + grid_to_modify + '.py.txt', "rt")
     text = f.read()
     f.close()
     for LLCAT in LLCATS:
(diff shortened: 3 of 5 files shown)


kbarnhart:
@mcflugen can you turn on building the docs on rtfd for this branch so we can test it fully and see what they would look like?

mcflugen:
> can you turn on building the docs on rtfd for this branch so we can test it fully and see what they would look like?

I believe I've turned them on for this branch and it looks like they are building. Does building and viewing them locally not show the problem you're having? That's a pain.

kbarnhart:
It appears to me that I have a local problem for building the docs BUT there is no problem when they are built in our Linux, python 3.6 build on Travis CI (the only build where we now test the docs...)

It looks like you turned it on, but it isn't working (https://readthedocs.org/projects/landlab/builds/7706749/)

Can you give me access to our read the docs backend?

mcflugen:
@kbarnhart One thing that seems could be a problem is that `docs/sphinxext` is missing but the submodule is still listed in `.gitmodules`. Maybe you didn't properly remove the `sphinxext` submodule?

There's not a lot of information in the *readthedocs* logs for this build but, what does seem to be there, indicates RTFD might be having trouble with this submodule.

kbarnhart:
Alrighty. This looks like it works. You can see the docs [here](https://landlab.readthedocs.io/en/barnhark-fix_broken_docs/)

I've done some spot checking and they look good to me. I have not pored through every page of the docs... I don't anticipate doing that. 

In order for this to totally take effect I think we will need to merge them _and_ make a release. That way all three versions of the docs on the website (`latest`, `release`, and `master`) will all be updated.

@mcflugen if we make a release for this, there are a few other things that I suspect should be in it (e.g. the results of PR #775, and maybe #780).

kbarnhart:
@mcflugen @SiccarPoint OK to merge this?

mcflugen:
> OK to merge this?

👍

SiccarPoint:
Yup from me

Issue #779, opened 2018-08-29
Travis OSX builds all fail


kbarnhart:
This seems to be an issue with at least PR #674 and #776. 

I can vouch for testing both these PRs locally with OSX and Py 3.6 (though not in an identical build environment).

Looking at the TravisCI log, it appears that in progress work by @margauxmouchene and @saisiddu is also affected. Ditto the [recent build of 1.5.2](https://travis-ci.org/landlab/landlab/builds/421853471).

In all cases the builds that complete with errors take ~50 minutes (while a landlab build typically takes less than 10 minutes). 

Somehow @mcflugen's PR #778 does not seem to be affected. 

@mcflugen maybe you understand this.

If not, this PR gives us a single place to discuss.

margauxmouchene:
I was just looking at this, I get the same errors as you. Did not get them just a few days ago.
I don't get any errors when testing locally either (OSX, Python 3.6 also).

mcflugen:
@kbarnhart This is a [problem](https://github.com/pytest-dev/pytest/issues/3888) with the newest release of pytest (v3.7.3). You should be fine if you limit pytest to less than that (I only tried 3.7.1 but, from the sounds of it, 3.7.2 should work). It looks like they've fixed the problem but they don't have a release with the fix yet.

If, in `.travis.yml` and `meta.yaml`, you install `pytest==3.7.2` your tests should pass.

kbarnhart:
OK. I made a PR that makes a fix.

kbarnhart:
This was fixed by #777.

PR #780, opened 2018-08-29
Address Issue 779: Travis fails due to the pytest version

kbarnhart:
Only thing changed here is the pytest version in the `.conda-recipe/meta.yaml` and the `.travis.yaml` files. 

Should address #779.

### DIFF
--- a/.conda-recipe/meta.yaml
+++ b/.conda-recipe/meta.yaml
@@ -31,7 +31,7 @@ requirements:
 
 test:
   requires:
-    - pytest
+    - pytest 3.7.4
   commands:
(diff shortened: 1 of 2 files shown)


kbarnhart:
@mcflugen OK to merge this? at this point it only changes the version number.

mcflugen:
> OK to merge this?

Sure. However, version 3.7.4 just came out, which should fix this problem. Maybe check if 3.7.4 works first?

kbarnhart:
Since 3.7.4 is now out (and is the latest version), just removing the version info should work, right?

mcflugen:
Correct. I don't see why we should specify the version number any more.

With pytest working again, there now shouldn't be any need to merge this, correct? I think we can close this pull request.

kbarnhart:
But, I agree, no need to merge this as 3.7.1 works from your PR #778.

Eventually we can test if removing the version number works.

Issue #785, opened 2018-09-11
FlowRouter should be deprecated


kbarnhart:
In PR #422 we revised the FlowRouter and functionally replaced it with the FlowAccumulator and FlowDirectors. 

However, the FlowRouter is commonly present in the docstrings, does not have a deprecation warning, and its documentation is not discoverable. 

We have discussed removing it in landlab 2.0 and thus it makes sense to formally deprecate it, make it less discoverable, etc.

PR #786, opened 2018-09-11
Barnhark/deprecate flow router

kbarnhart:
This PR addresses Issue #785.

It adds a deprecation warning to FlowRouter, changes all use of FlowRouter to FlowAccumulator in docstrings and tests (except in the FlowRouter component and tests themselves), and removes FlowRouter from the docs. 

A few other notes:

I'm nearly certain that when we (me with lots of guidance from @mcflugen @SiccarPoint @gregtucker) redesigned FlowAccumulator out of FlowRouter that we agreed that we should do what is in this PR... I just never got around to actually doing it.


kbarnhart:
Looks like there is an error associated with python 2.7 versions not seeing the DeprecationWarning test. Will try to fix.

kbarnhart:
Alrighty. It looks like this all passes.

kbarnhart:
@SiccarPoint @mcflugen thoughts on this?

SiccarPoint:
Yes, I think this makes sense. That said, this thing is pretty deeply embedded in quite a lot of code, so we probably ought to be generous with the deprecation time if in doubt.

Also probably a publicity blitz would be a good idea.

I can review Tuesday if no-one else has got to it by then.

kbarnhart:
I made the warning state that it would be deprecated in v2.0

Which I thought was the earliest reasonable time (given expectations of backwards compatibility).

Thanks for reviewing on Tuesday. 

I follow on thought is that we (me?) should do a clean look through of ALL the tutorials to make sure no examples remain there.

mcflugen:
Looks good to me.

SiccarPoint:
Also good to me.

Issue #771, opened 2018-08-06
DrainageDensity documentation is not discoverable


kbarnhart:
It needs to be listed in its own automodule .rst file and linked to by index.rst. Tagging @cmshobe (talk with me if you have questions on how to do this).

kbarnhart:
It would also be great if the docstrings included a description of how the supported parameters (e.g. slope exponent) are used.

kbarnhart:
@jdgemperline you and I just discussed this. Apparently I made an issue of it three weeks ago.

cmshobe:
ok @kbarnhart @jdgemperline I'm going to address this now.

PR #787, opened 2018-09-19
adding drainage density component to index.rst

cmshobe:
Addresses #771. There was already a drainage density .rst file, but there was no line for drainage density in the index.rst file. @jdgemperline this should make the Dd docs work (assuming the docs are working generally).

@kbarnhart check and merge as the spirit moves you


kbarnhart:
this looks great @cmshobe !
if you have additional comments @jdgemperline please make them here. otherwise I'll merge as is when the tests pass (which they should because this is only about docs).

Issue #788, opened 2018-09-20
plt.gca().set_axis_bgcolor(color_for_background) deprecated


kbarnhart:
As of mpl 2.0.

PR #789, opened 2018-09-20
changing to set_facecolor

kbarnhart:
addresses #788


kbarnhart:
@SiccarPoint this is a one line fix. Check it out and merge?

Issue #798, opened 2018-09-26
Lithology JOSS proof citations are not formatted correctly


kbarnhart:
The [PDF proof of the Lithology manuscript](https://github.com/openjournals/joss-papers/blob/joss.00979/joss.00979/10.21105.joss.00979.pdf) has citations that are not formatted correctly.

This Issue is associated with [JOSS Review # 979](https://github.com/openjournals/joss-reviews/issues/979)

kbarnhart:
This now fixed.

PR #799, opened 2018-09-26
Address #798 : Fix citations in Joss Proofs

kbarnhart:
This PR updates the citations in the manuscript. It also updates the read-the-docs URLs (which were missing .html) and adds a URL to the tutorial. 

It addresses #798 

I'm not 100% sure that the citations will work perfectly, but the way to find out is to merge this into master, and then have JOSS `whedon` recompile the proof.


mcflugen:
👍 

(Because the paragraphs were on a single line it was difficult to see what you changed. A few more line breaks (at 88 characters) would have been helpful. Just something to keep in mind.)

Issue #802, opened 2018-09-27
FlowAccumulator MFD on flat surfaces


SiccarPoint:
This isn't exactly urgent, but wanted to flag it while it's in front of me.

The FlowAccumulator (and so, presumably, the FlowDirectors as well) Error confusingly if called on a flat surface. The error message is opaque, and it's an IndexError. IMO calling on a flat surface is a potential use case, albeit rare.

```
    mg = RasterModelGrid((5, 5), 1.)
    z = mg.add_zeros('node', 'topographic__elevation')
    fa = FlowAccumulator(mg, flow_director='MFD')
    fa.run_one_step()
Traceback (most recent call last):
  File "/Users/danhobley/git_landlab/landlab/components/stream_power/tests/test_TL.py", line 17, in test_multidirection
    fa.run_one_step()
  File "/Users/danhobley/git_landlab/landlab/components/flow_accum/flow_accumulator.py", line 1034, in run_one_step
    self.accumulate_flow()
  File "/Users/danhobley/git_landlab/landlab/components/flow_accum/flow_accumulator.py", line 988, in accumulate_flow
    s = flow_accum_to_n.make_ordered_node_array_to_n(r, p)
  File "/Users/danhobley/git_landlab/landlab/components/flow_accum/flow_accum_to_n.py", line 458, in make_ordered_node_array_to_n
    construct_it(baselevel_nodes)  # don't think this is a bottleneck, so no C++
  File "/Users/danhobley/git_landlab/landlab/components/flow_accum/flow_accum_to_n.py", line 169, in construct__stack
    visited_enough = num_visits[visited] == self.num_receivers[visited]
IndexError: arrays used as indices must be of integer (or boolean) type
```

In this case, the Error is resolved by roughening the surface.

kbarnhart:
Thanks for pointing this out @SiccarPoint 

Here is a MCE:

```python
from landlab import RasterModelGrid
from landlab.components import (FlowAccumulator)


for fd in ['FlowDirectorSteepest', 
           'FlowDirectorD8', 
           'FlowDirectorDINF', 
           'FlowDirectorMFD']:
    mg = RasterModelGrid(40, 60, dx=10)
    z = mg.add_zeros('topographic__elevation', at='node')
    fa = FlowAccumulator(mg, flow_director=fd)
    try:
        fa.run_one_step()
        print(fd + ' success')
    except IndexError:
        print(fd + ' failed')
    del mg, z, fa
FlowDirectorSteepest success
[...]

Issue #667, opened 2018-04-17
D infinity not interacting correctly with low elevation closed nodes


kbarnhart:
The `FlowDirectorDINF` first runs the standard algorithm, and then adjusts for closed boundary nodes. 

PR #658 tests identified this issue with a grid in which core nodes are surrounded by closed nodes with low elevations. 

``` python
import numpy as np
import matplotlib.pylab as plt

from landlab import RasterModelGrid

from landlab.components import FlowDirectorDINF, FlowDirectorMFD
from landlab.plot.drainage_plot import drainage_plot

fs = (9, 4)
mg = RasterModelGrid((5, 4), spacing=(1, 1))
z = np.array ([[0,0,0,0],[0,21,10,0],[0,31,20,0],[0,32,30,0],[0,0,0,0]],dtype='float64')
mg.add_field('node','topographic__elevation', z)
mg.set_closed_boundaries_at_grid_edges(bottom_is_closed=True, left_is_closed=True, right_is_closed=True, top_is_closed=True)

fd = FlowDirectorDINF(mg)
fd.run_one_step()

fig = plt.figure(figsize=fs)
drainage_plot(mg)
fig.savefig('dinf_closed.png')
```
![dinf_closed](https://user-images.githubusercontent.com/6589339/38884567-3f12aa1e-422d-11e8-9644-1f89233582be.png)

This contrasts with what you would get if the boundary nodes were closed. 

```python
mg = RasterModelGrid((5, 4), spacing=(1, 1))
z = np.array ([[0,0,0,0],[0,21,10,0],[0,31,20,0],[0,32,30,0],[0,0,0,0]],dtype='float64')
mg.add_field('node','topographic__elevation', z)

fd = FlowDirectorDINF(mg)
fd.run_one_step()

fig = plt.figure(figsize=fs)
drainage_plot(mg)
fig.savefig('dinf_closed.png')
```

![dinf_open](https://user-images.githubusercontent.com/6589339/38884623-5dd46f28-422d-11e8-81c6-0ec8aed8baca.png)

Or with what running MFD would create:

```python
mg = RasterModelGrid((5, 4), spacing=(1, 1))
z = np.array ([[0,0,0,0],[0,21,10,0],[0,31,20,0],[0,32,30,0],[0,0,0,0]],dtype='float64')
mg.add_field('node','topographic__elevation', z)
[...]

PR #803, opened 2018-09-28
Fix Flats in Accumulate-to-Multiple, BC and closed node issues in Dinfinity

kbarnhart:
This starts to address #802 

Not done yet.


kbarnhart:
Now ready for review. 

@SiccarPoint and @amanaster2 could you both review this pull request (I can only formally add @SiccarPoint in GitHub).

This addresses a few different Issues (#802, #667, and #811). 

- It makes sure that the FlowDirectorDINF and FlowDirectorMFD work correctly with perfectly flat terrain (Issue #802)
- It modified FlowDirectorDINF to correctly handle flow next to closed boundary nodes correctly (Issue #667) and makes it so it does not appear the flow is coming from closed nodes (#811). 

I've added a number of new tests for MFD and Dinf that provide analytical tests to both of these flow directing methods (for sloped and flat surfaces, with and without closed boundaries). This is where I'd focus looking.

kbarnhart:
@amanaster2 - @saisiddu recommended I make some comments being a bit more specific about what I'd like you to do (if you are able). 

1) check out this pull request branch and see if it fixes the problem you posted in #811 
2) look at the new tests you can see them most easily be clicking the "files changed" tab above. I built each of these tests as an "analytical" solution -- a situation in which I think I know what the right answer is. I'd be very interested for you and @SiccarPoint to determine if you think these tests are correct. If you think there is a place were I got the test wrong, or where another tests is needed, just make a comment. 

Any amount of either of these items would be great! Thank you!

amanaster2:
@kbarnhart Okay, great. Thanks for the specificity! I will try to do so this week, but I'll be going into the field Wed-Sat. If not this week, I'll look at it early next week.

kbarnhart:
@amanaster2 sounds great! thanks for having a look! early next week is totally fine.

kbarnhart:
@SiccarPoint thanks@ @amanaster2 let me know what you think.

amanaster2:
@kbarnhart Everything looks good to me!

Issue #812, opened 2018-10-16
tons of inspect.getargspec() is deprecated warnings


kbarnhart:
In testing (on py 3.6) I get tons of warnings associated with decorated functions. 
```
/Users/barnhark/projects/landlab/landlab/landlab/utils/decorators.py:134: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
  argspec = inspect.getargspec(func)
```

It would be lovely not to have to scroll through them while looking at pytest results.

PR #813, opened 2018-10-16
Fix argspec warnings

kbarnhart:
Addresses #812 

Passes tests locally (OSX/py3.6)

### DIFF
--- a/landlab/utils/decorators.py
+++ b/landlab/utils/decorators.py
@@ -20,7 +20,7 @@ import inspect
 
 try:
     from inspect import getfullargspec
-except ImportError
+except ImportError:
     from inspect import getargspec as getfullargspec
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


kbarnhart:
@mcflugen Looks like this fails on Py 2.7 builds... If you have any ideas on how to have it pass but not have all the warnings...

mcflugen:
@kbarnhart There is no `getfullargspec` in the Python 2.7 `inspect` module. Maybe we need something like,
```python
try:
    from inspect import getfullargspec
except ImportError
    from inspect import getargspec as getfullargspec
```
and then further down use,
```python
argspec = getfullargspec(func)
```
Since `getargspec` isn't deprecated in 2.7, there should be any warning issued.

kbarnhart:
Thanks! for catching the typo. @mcflugen. It eluded me...

Issue #769, opened 2018-07-31
New data structure to record time- and/or space-dependent variables 


margauxmouchene:
This is an outline for the new data structure that @nathanlyons and I are working on, 
The issue is to keep track of what we need to put in this new tool.

Fell free to add/comment!

1. Tentative name: DataRecord (courtesy of @nathanlyons)

2. Aim:
to combine @kbarnhart 's ItemCollection and @nathanlyons 's RecordCollection
will use xarray, 
will record time- and/or space-dependent variables describing the model and/or a set of items

3. Structure

The xarray DataSet created by DataRecord will have the following characteristics:

- Dimensions (one or both):
   - time: (size=number_of_timesteps)
   - item_id (size=number_of_items)
- Coordinates:
   - time (time) float [dt, 2dt, 3dt, ... number_of_timesteps * dt]
   - item_id (item_id) int [0, 1, 2, ... number_of_items - 1]

- Data variables (if Item_id is specified, the following two variables must exist; if not, no constraints on data variables):
   - Grid_element 		(item_id, time) str 	[‘node’, ‘node’, ‘link’, …]
   - Element_id		(item_id, time) int 	[0, 3, 3, …]

- Other possible variables (examples):
   - Mean_elevation	(time) float		[100, 110, 113, …]
   - Item_type		(item_id) str		[‘granite’, ‘limestone’, ‘granite’, …]
   - Item_size		(item_id, time) float	[5, 0.3, 2, …]

4. Methods and properties
- Methods we need:
   - “Checking methods”: if items are defined: 
	 - Check that they all have grid_element AND element_id specified
	 - Check that all grid_elements exist in the grid type that is used (e.g. Network grid doesn’t have cells)
        - Check that all element_ids exist on the grid (e.g. cannot be node 9 on a 3x3 grid)
    -	Add/delete:
         - Add/delete item (concat/drop?)
         - Add/delete time (concat/drop?)
         - Add/delete data variable 
    - Get/set values (do we need specific methods for this? Indexing makes this easy to do without calling a method)
    - Split/Groupby etc. (inherited from xarray?)
    - From ItemCollection:
        - Get_items_on_grid_element(at=’node’, element_id=0)
        - Calc_aggregate_value(func, var, at='node', fill_value=_FILL_VALUE, args=(), **kwargs)

- Properties:
From RecordCollection:
   - Number_of_timesteps
   - Model__times (array of all recorded timesteps)
   - Model__earliest_time (and latest_time, and prior_time)

[...]

PR #816, opened 2018-10-18
DataRecord init options

nathanlyons:
The change I made to init options described  [in this comment](https://github.com/landlab/landlab/issues/769#issuecomment-421528527) seem to be covered by the tests you have already created, so I did not make additional tests. 

The run time improvement described in the linked comment is due to this line change in the `set data` method:
```python
# previous line
time_index = int(self.time_coordinates.index(time[0]))

# faster
time_index = np.where(self.time.values == time)[0][0]
```

Questions and other minor things I hoped to save you from having to doing:
- Editted first bit of method docstrings to be limited to one line of text.
- Are the adjacent double parenthesis needed? For example:
```python
raise TypeError(('time must be a list or a 1-d array'))
```
- Are `_input_var_names`, etc necessary given that this is not a component?
- avoided inline comments (PEP8)
- sometimes its written 'Datarecord' and sometimes 'DataRecord'. I modified occurences in print statements to consistently be 'DataRecord'. I advocate for it to always to be this only because I as a user wonder if different capitalization has significance. (Maybe there is a significance I'm not getting.)

### DIFF
--- a/landlab/data_record/data_record.py
+++ b/landlab/data_record/data_record.py
@@ -73,12 +73,12 @@ class DataRecord(Dataset):
         ----------
         grid : ModelGrid
         time : list or 1-D array of float or int (optional)
-            The initial time coordinates to add to the recorded. A time
-            coordinate is not created if the value is 'None' (default).
+            The initial time(s) to add to the record. A time dimension is not
+            created if the value is 'None' (default).
         items : dict (optional)
-            No item is created if the value is 'None' (default). Otherwise,
-            dictionary describes the position of generic items on the grid. The
-            structure is:
+            Generic items that live on grid elements. No item is created if the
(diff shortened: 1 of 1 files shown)


margauxmouchene:
Thank you for these improvements!

I will merge that and then address your other comments in the main branch.

You are right:
- double parentheses are not needed,
- `input_var_names` and co. are not relevant here, will delete those lines.

pfeiffea:
Hi y'all. Jon Czuba and I are finally circling back to this project, and it seems like DataRecord is ready for use. Is that true? 

@margauxmouchene would you be able to chat briefly sometime in the next week or two about how DataRecord operates differently from ItemCollection?

kbarnhart:
@pfeiffea can we set up a call too. I want to make sure I'm not holding you guys back either...

PR #818, opened 2018-10-22
Remove diff_2D.py

### LINKED ISSUE #817: Kill diff_2D?
Consensus is that the `diff_2D.py` component is 100% vestigial and can be removed with little risk. Let's try.

SiccarPoint:
This PR resolves #817


kbarnhart:
We should probably decide if this should just be removed... or actually deprecate it like we are supposed to in order to keep backwards compatibility until LLv2.0

SiccarPoint:
I *think* this is entirely untested, and de facto broken for use. All the tests being picked up by pytest-cov are inherited, from elsewhere, if I'm reading the code right. My confidence that no-one is using this at all exceeds 99%. And you know how twitchy I am about removing stuff wholesale in general...!

SiccarPoint:
It also doesn't exist as a component, or even its own `__init__.py` declarations. Nor any user facing documentation. So can't be imported easily, nor can users learn about it with actually diving into the code. It's only going to be power users (i.e. developer users) who might feasibly have even seen this, and TBH, they know enough to resolve any problems this will cause for them in a worst case scenario.

kbarnhart:
@SiccarPoint you've convinced me. 
I've restarted your AppVeyor build that failed...

PR #819, opened 2018-10-25
Margauxmouchene/data record

margauxmouchene:
Left to do:

- one last round of checks by @nathanlyons 
- do we want someone else to look at it?
- should we put xarray in cite_as? See [here](http://xarray.pydata.org/en/stable/faq.html#how-should-i-cite-xarray), they have a paper and DOIs for specific versions.


nathanlyons:
It looks good and all still runs well for me. Also, its covered 100% by tests!

> should we put xarray in cite_as? See here, they have a paper and DOIs for specific versions.

I think it's OK to cite or not. The Xarray link mentions citing only in the context of academic publications. (I made note to cite it an upcoming SpeciesEvolver paper.) I suppose not including the citation in DataRecord makes it less likely that Xarray is cited when DataRecord is used by others in the future. Conversely, I have not seen numpy and many other dependancies cited elsewhere in Landlab, and it seems that it would be a lot to cite each dependency in each module its used. We could instead consider citing things like Xarray for Landlab in general, if there is a sensical place to do this.

> do we want someone else to look at it?

It seems to me to be in good shape, so I think you can merge at will, but please do ask others if you wish. I am not in-the-know with item collection compatibility.

margauxmouchene:
Thank you @nathanlyons. 
Discussed citing xarray with @kbarnhart and @mcflugen: it probably makes sense to advertise our use of xarray at a higher level (it also used in other parts of Landlab), not specifically in DataRecord so I removed it here.
I created a Tutorial for DataRecord [here](https://github.com/landlab/tutorials/tree/margauxmouchene/DataRecord_tutorial/data_record), @nathanlyons can you have a look at it? And if anyone else wants/can look at the code and/or tutorial, please let us know what you think!

Otherwise we're ready to merge!

nathanlyons:
I am thrilled that DataRecord is being merged and that you took it on. I will get to the tutorial asap.

PR #822, opened 2018-10-29
SiccarPoint/lossy FlowAccumulator

SiccarPoint:
This PR adds a LossyFlowAccumulator to the components, which permits dynamic loss or gain of discharge of water during flow downstream. This inherits heavily from the standard FlowAccumulator, but the innovation is in the passing of a loss_function, a true Python function that describes how water should be gained or lost from node-to-node.

This exists due to a use case one of my PhD students is developing, but hopefully useful in plenty of other places too.

I believe the component is 100% tested, wired into the docs, documented, and hopefully largely good to go. Comments on clarity extremely welcome!


SiccarPoint:
@mcflugen - any ideas why all my 64bit Python 3.6 builds are failing here!? All fine locally.

kbarnhart:
@SiccarPoint can you give me a preferred review timeline?

SiccarPoint:
Kind of up to you. Would be nice in the next few weeks if poss. Got a student who wants to use this, so getting it in a release would be good in the not-too-distant future. (Or I can make him use a dev version...!)

kbarnhart:
@SiccarPoint a couple of super big picture clarification questions. 

As I understand it, this component implements a loss of discharge during accumulation. I'm going to take a big step back and think about what this means in terms of hydrology to make sure I have everything right. 

If I have it right, I have some more specific comments. They might not apply if I don't have it right, so let me know if I got it right and I'll proceed. 

Our very simplistic hydrology:

First, it rains, this creates a precipitation water unit flux, p [L/T]. For a given  area A [L2] and runoff fraction R [unitless] this yields a volume of water per time q [L3/T]. q represents the surface water that runs off of each cell and is accumulated by the flow accumulator. 

q= p * A * R = [L/T] * [T] * [L2] * [-] = [L3/T]

Our field `water__unit_flux_in` thus represents p * R (and I think it might make sense for it to be called `surface_water__unit_flux_in`, but that is not a discussion that needs to happen now). 

For each node, the flow accumulator accumulates the upstream area [L2] and discharge [L3/T]. These get stored as the fields `drainage_area` and `surface__water_discharge`.

The standard FlowAccumulator takes all flow coming into a node, adds the runoff from that node's cell, and passes the flow downstream. In contrast, the LossyFlowAccumulator takes all flow coming into a node and applies the loss function. Then the runoff coming from that node's cell is added and passed downstream. 

An example of a geomorphic system where this would be necessary might be a karst landscape where surface flow "disappears" from the surface when it encounters a sinkhole or other surface entry to the karst system.

mcflugen:
@SiccarPoint Did you ever figure out your Travis problems with Python 3.6 and Mac? I was able to fix it by adding `osx_image: xcode6.4` in the `.travis.yml`. I guess there is a problem with a newer Mac compiler or something. Anyway, I've fixed it on another branch, if you want to [have a look](https://github.com/landlab/landlab/blob/mcflugen/full-flexure-coverage/.travis.yml).

SiccarPoint:
@mcflugen Thanks - all passing now by the looks of things.

SiccarPoint:
I think you have this right, in broad strokes. However, the how the loss function works is both powerful and flexible. You're describing doing this:
```
>>> def mylossfn(Qw, nodeID, dummy, grid):
...     return Qw * (1. - grid.at_node[somenodefield][nodeID])  # or whatever
```
This is simple, and means this is implicitly an upstream solution.

But it's equally valid (and arguably more useful) to do this with your function:
```
>>> mylossfn2(Qw, dummy, linkID, grid):
...     return Qw*(1. - grid.at_link[somelinkfield][linkID])
```
Now, here we are actually honouring the physics properly, since loss occurs during flow not at a point (i.e., on a link, not at a node). Also, creative use of functions like this allow loss to be dependent on both length of link (longer flow => more loss?), or to enable the route-to-multiple stuff to be more sensible, since you might want different properties depending on which way the flow goes.

We retain the nodeID option though, since it's clearly a common, simple use case. It's also possible to define an even simpler function that loses or gains in a way that isn't spatially or temporally variable. The docstring examples show all of these happening, and should guide the user through it.

Karst is a valid use case, but the target one would be true coupling to subsurface flow. The surface discharge needs to be able to lose to (and gain from) the subsurface as it goes. There's also a clear use case here for evaporative losses from the actual river during flow. This kind of evaporative loss is the reason why the component treats the accumulated flow separately from the incoming discharge at the node - we only want to model *flow* losses, not static ET balances at a point which could be handled by e.g. Green-Ampt before being passed in to the accumulator, as part of P. Best not in fact to think of sinkholes, since that implies total loss at a point, rather than partial loss during flow.

kbarnhart:
@SiccarPoint I'm writing down the things we talked about yesterday so I can remember
1) you convinced me you need a function.
2) We decided that the flow loss you calculate should get written out into a field so another function (e.g. subsurface flow) can use it. 
3) We decided that the run_one_step method should be inherited, and that both the normal FlowAccumulator and the LossyFlowAccumulator would get two private functions:

`_accumulate_flow_to_n()`
`_accumulate_flow_to_one()`

that would be called within the run-one-step

4) the run one step method (at the base, flow accumulator level) can get cleaned up a bit (most of this is detritus from the different flow receiver names that we fixed in PR #775 

5) Not duplicate all the documentation in FlowAccumulator, but point to it. 

6) Importantly, this means that this component won't have its own run_one_step method. Ideally it would be good to have the ability to stick inherited method documentation within this (and other components) documentation.  We decided that this is a good idea, but it merits a separate issue. (I'm going to make one).

SiccarPoint:
Ok, I'm working at (2). The field that stores this info really ought to be on links, as the loss is definitely happening during transit, so on them not the nodes. There's also a significant uniqueness problem in the ..._to_n cases, since nodes can both shed and receive discharge to and from more than one node.

But: If the routing is happening as D8, we have a problem - AFAIK there isn't a way to create a true d8 field, right? So this info can't get stored as a field??

SiccarPoint:
If we want to add these, there's a whole kettle of fish around mapping links to d8s... We've previously dodged this problem by largely pretending the actual links don't exist when flow routing, but here that approach is reaching its limits!

SiccarPoint:
I'm resolving this by mapping onto the nodes regardless, but may be worth revisiting at some point.

SiccarPoint:
I believe this is now good to go. @kbarnhart - you might want to skim what I've done in FlowAccumulator before I hit go, but I'm happy at this point. Looks way cleaner now! Let me know if you want to or not.

kbarnhart:
@SiccarPoint I should be able to look at this by Monday. That timeline OK?

SiccarPoint:
Absolutely! Thanks so much

SiccarPoint:
Unclear why this isn't building right. What's this linting error??

mcflugen:
@SiccarPoint It was just telling you that you had some unused imports and variables that were assigned but never used. There was also some PEP8 stuff. I fixed up these linting errors so I think the tests should pass now.

I figured it would be easier to fix this kind of stuff as we go rather than with infrequent but big cleanups so I added these checks to be run along with our usual tests on Travis. If you have *flake8* installed, you can run `make lint` to run these checks locally (or `flake8 <file-to-check>`). Sorry for the trouble.

kbarnhart:
@mcflugen can you point us all to a statement about the lint standards that are now being enforced? Ideally, this would be something in the docs.

mcflugen:
@kbarnhart I had meant to add something to the docs before I left but I didn't have time. I will though.

Currently we check for all the [*flake8* violations](https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes) and [*pycodestyle* violations](http://flake8.pycqa.org/en/latest/user/error-codes.html) except for (as defined in our `setup.cfg`)
* E203: whitespace before ‘:’
* E501: line too long (n > 88 characters)
* W503: line break before binary operator

To format files for PEP8, I like to use *isort* + *black*.

SiccarPoint:
Wait - I'm not sure about b1d4145 back there. I really don't like the look of those really long lines, especially when the whole of the rest of the code is so nicely blocked out at 80 chars. What I had was totally Pythonic, no? We don't do this with our docstrings, so why do it here??

mcflugen:
@SiccarPoint I rewrapped the strings for `_var_doc`. The issue I was dealing with was that the keys for `_var_doc` were getting lost in the wrapped strings. The way it is now, lines that have wrapped strings begin with a `+`, which make it easier to identify the keys.

What do you think?

landlab/components/flow_accum/flow_accumulator.py
```diff
@@ -987,31 +984,40 @@ def accumulate_flow(self, update_flow_director=True):
             self._grid["node"]["flow__data_structure_delta"][:] = delta[1:]
 
             if self._is_raster:
-                tempD = BAD_INDEX_VALUE * np.ones((self._grid.number_of_links * 2))
+                tempD = BAD_INDEX_VALUE * np.ones(
```
kbarnhart: @SiccarPoint FYI. This is the janky approach I had to take because we can't store information at D8s...

landlab/components/flow_accum/flow_accumulator.py
```diff
@@ -987,31 +984,40 @@ def accumulate_flow(self, update_flow_director=True):
...
+            a[:], q[:] = self._accumulate_A_Q_to_n(s, r, p)
 
-            # step 4. Accumulate (to one or to N depending on direction method. )
-            a, q = flow_accum_to_n.find_drainage_area_and_discharge_to_n(
-                s, r, p, self.node_cell_area, self._grid.at_node["water__unit_flux_in"]
-            )
-            # store drainage area and discharge.
-            self._grid["node"]["drainage_area"][:] = a
-            self._grid["node"]["surface_water__discharge"][:] = q
+        return (a, q)
```
kbarnhart: @SiccarPoint this revision of the flow accumulator run-one-step looks great!

landlab/components/flow_accum/lossy_flow_accumulator.py
```diff
@@ -0,0 +1,464 @@
...
+    cannot accumulate any discharge.
+
+    LossyFlowAccumulator stores as ModelGrid fields:
+
+        -  Node array of drainage areas: *'drainage_area'*
+        -  Node array of discharges: *'surface_water__discharge'*
+        -  Node array of discharge loss in transit (vol/sec). This is the
+            total loss across all of the downstream links:
+            *'surface_water__discharge_loss'*
+        -  Node array containing downstream-to-upstream ordered list of node
```
kbarnhart: @SiccarPoint lets have a group discussion about this field name. 

I like it, but I think its worth talking about it before we add a new standard name.

I've added it to the call agenda.

landlab/components/flow_accum/lossy_flow_accumulator.py
```diff
@@ -0,0 +1,464 @@
...
+        is to be calculated, linkID is the ID of the link down which the
+        outflow drains, and grid is a Landlab ModelGrid. Note that if a linkID
+        is needed, a nodeID must also be specified, even if only as a dummy
+        parameter; similarly, if a grid is to be passed, all of the preceding
+        parameters must be specified. Both nodeID and linkID are required to
+        permit spatially variable losses, and also losses dependent on flow
+        path geometry (e.g., flow length). The grid is passed to allow fields
+        or grid properties describing values across the grid to be accessed
+        for the loss calculation (see examples).
+        This function should take (float, [int, [int, [ModelGrid]]]), and
```
kbarnhart:
```suggestion
        This function expects (float, [int, [int, [ModelGrid]]]), and
```

landlab/components/flow_accum/lossy_flow_accumulator.py
```diff
@@ -0,0 +1,464 @@
...
+        return a single float.
```
kbarnhart:
```suggestion
        must return a single float. This behavior is verified during component instantiation. 
```

landlab/components/flow_accum/lossy_flow_accumulator.py
```diff
@@ -0,0 +1,464 @@
...
+        DepressionFinderAndRouter instance (e.g., partion_method for
+        FlowDirectorMFD). This will have no effect if an instantiated
+        component is passed using the flow_director or depression_finder
+        keywords.
+
+    Examples
+    --------
+    These examples pertain only to the LossyFlowAccumulator. See the main
+    FlowAccumulator documentation for more generic and comprehensive examples.
+
```
kbarnhart: Add a link to the FlowAccumulator documentation.

landlab/components/flow_accum/lossy_flow_accumulator.py
```diff
@@ -0,0 +1,464 @@
...
+    array([[  0. ,   0. ,   0. ,   0. ,   0. ,   0. ],
+           [  5.6,   5.6,   3.6,   2. ,   2. ,   0. ],
+           [ 10.4,  10.4,   8.4,   6.4,   4. ,   0. ],
+           [  0. ,   0. ,   0. ,   0. ,   0. ,   0. ]])
+    >>> mg.at_node['surface_water__discharge'].reshape(mg.shape)
+    array([[ 0. ,  0. ,  0. ,  0. ,  0. ,  0. ],
+           [ 4. ,  4. ,  2. ,  2. ,  2. ,  0. ],
+           [ 0. ,  8.5,  6.5,  4.5,  2.5,  0. ],
+           [ 0. ,  0. ,  0. ,  0. ,  0. ,  0. ]])
+    """
```
kbarnhart: I think its much better now that you just have the Lossy documentation!

landlab/components/flow_accum/lossy_flow_accumulator.py
```diff
@@ -0,0 +1,464 @@
...
+        super(LossyFlowAccumulator, self).__init__(
+            grid,
+            surface=surface,
+            flow_director=flow_director,
+            runoff_rate=runoff_rate,
+            depression_finder=depression_finder,
+            **kwargs
+        )
+
+        if loss_function is not None:
```
kbarnhart: @SiccarPoint and @mcflugen 

Once this PR is merged, both LossyFlowAccumulator _and_ Lithology will do function argument and return checking. Does it make sense to move the function checking capabilities for both of them to `landlab.utilities`?

landlab/components/flow_accum/lossy_flow_accumulator.py
```diff
@@ -0,0 +1,464 @@
...
+                return float(Qw)
+
+            self._lossfunc = lossfunc
+
+        # add the new loss discharge field if necessary:
+        if "surface_water__discharge_loss" not in grid.at_link:
+            self.grid.add_zeros(
+                "node", "surface_water__discharge_loss", dtype=float, noclobber=False
+            )
+
```
kbarnhart: @SiccarPoint a thought about this field -- that we should discuss on the next call. 

This field JUST records the impact of the loss function (direction of information: from function to component to field). But, lets pretend that there is someone who doesn't want/need to use a loss function, but wants to specify a spatially variable field that is "surface_water__discharge_loss" (or "surface_water__discharge_loss_proportion"?)

We should at least consider supporting this. I've made notes in the teleconference agenda.

landlab/components/flow_accum/tests/test_lossy_flow_accumulator.py
```diff
@@ -0,0 +1,1131 @@
...
+    )
+
+    LossyFlowAccumulator(mg)
+    assert_array_equal(z, mg.at_node["topographic__elevation"])
+    assert_array_equal(np.zeros(100), mg.at_node["drainage_area"])
+    assert_array_equal(np.ones(100), mg.at_node["water__unit_flux_in"])
+
+    LossyFlowAccumulator(mg, runoff_rate=2.)
+    assert_array_equal(np.full(100, 2.), mg.at_node["water__unit_flux_in"])
+
```
kbarnhart: I think you need to check that the surface_water__discharge_loss field has been created.

landlab/components/flow_accum/lossy_flow_accumulator.py
```diff
@@ -0,0 +1,464 @@
...
+        runoff rate is desired, pass this field name and update the field
+        through model run time. If both the field and argument are present at
+        the time of initialization, runoff_rate will *overwrite* the field. If
+        neither are set, defaults to spatially constant unit input.
+    depression_finder : string, class, instance of class, optional
+        A string of class name (e.g., 'DepressionFinderAndRouter'), an
+        uninstantiated DepressionFinder class, or an instance of a
+        DepressionFinder class.
+        This sets the method for depression finding.
+    loss_function : Python function, optional
```
kbarnhart: Two things to be more specific about (here and other places where you list loss_function in the docs).

1) is a positive number a loss or a gain?

2) I think that linkID also means d8ID (if you are on a raster with D8s). We should be more careful with this language. @mcflugen thoughts?

landlab/components/flow_accum/flow_accumulator.py
SiccarPoint: Thanks...

landlab/components/flow_accum/flow_accumulator.py
SiccarPoint: Sweet

landlab/components/flow_accum/lossy_flow_accumulator.py
SiccarPoint: I'll leave this until Monday...

landlab/components/flow_accum/lossy_flow_accumulator.py
SiccarPoint: Remind me how to do this from a docstring? Not sure if I ever figured that out... Is it just conventional markup?

landlab/components/flow_accum/lossy_flow_accumulator.py
SiccarPoint: Cool

landlab/components/flow_accum/lossy_flow_accumulator.py
SiccarPoint: I like this a lot, but it should certainly be a separate PR if we want to support it. I'll leave this open to remind me to make one once we've resolved the naming as above. I think it would be slightly clearer to have this as an absolute loss rather than a proportional one, though I'm not wedded to that. Basically, I think it will save a round of calculation if you're taking this as an output and doing something else with it. Clearly if an absolute number we'd have to threshold it at the amount of discharge currently at the node if used as an input.

landlab/components/flow_accum/tests/test_lossy_flow_accumulator.py
SiccarPoint: Done

landlab/components/flow_accum/lossy_flow_accumulator.py
SiccarPoint: 1. I think this is clear, due to the way the loss function works: it takes a discharge value, and returns a discharge value. This returned value must always be >0. How the user implements this inside the loss function is their business.
2. Good call.

SiccarPoint: I've also tried to make (1) more clear.

landlab/components/flow_accum/lossy_flow_accumulator.py
SiccarPoint: We're OK with this naming on the call.

landlab/components/flow_accum/lossy_flow_accumulator.py
SiccarPoint: We're suggesting on the call that the input field should be specified as a FLUX i.e. [L/T] to make it obviously distinct from the output field - also many inputs might provide this.

Will make a PR.

Issue #824, opened 2018-11-02
Two functions to create fields


kbarnhart:
It would be good to have two functions that can help create fields. 

1) `mg.create_field_from_file`

```python
def create_field_from_file(self, at, name, path, halo)
```

2) `mg.create_synthetic_field`
something like (forgive the bad formatting):
```python
def create_synthetic_field(self, at, name,
            mean=0,
            std=0,
            function=lambda x, y: x*0+y*0,
            add_initial_value_to_all_nodes=False,
            add_noise_to_all_nodes=False,
            add_function_to_all_node: False):

   # do some function checking here.


   # create field values
   field_values = np.zeros(self.size(at))

   # add initial value
   if add_initial_value_to_all_nodes:
        field_values += mean
   else:
        field_values[self.core_nodes] += mean

   # add noise
   noise = std * np.random.randn(self.size(at))

   if add_noise_to_all_nodes:
        field_values += noise
   else:
        field_values[self.core_nodes] += noise[self.core_nodes]

    func_val = function(self.x_of_node, self.y_of_node)

    if add_function_to_all_node:
         field_values += func_val
    else:
         field_values[self.core_nodes] += func_val[self.core_nodes]

   self.add_field(at, name, field_values)

```

SiccarPoint:
This sounds great. I've been wishing for (2) for ages.

PR #827, opened 2018-11-03
Synthetic fields

kbarnhart:
One half of addressing #824.

In progress


mcflugen:
@kbarnhart I was thinking this would be a module (maybe `landlab.synthetic` or maybe a subpackage?) that would just contain functions that took *x* and *y*, an optional `out` keyword, function-specific keywords, and return an array of calculated values. For example we could have functions that generate:
* a linearly dipping plane, given a point and a normal
* an array of white noise
* an array of pink noise
* an array with a step function

I'm sure there are lots more. Each function should also be compatible with one another so that they could be chained together. For instance, a user could say, "give me a step function and add some white noise to it".

What do you think?

mcflugen:
And then we would have a function that attaches these functions to model grids that operate something like `grid.ones`, `grid.full`, etc.

kbarnhart:
@mcflugen I think that is much better and makes sense. 

On thing that I think is important in the creation of the synthetic grids is the ability to add values to some, but not all nodes (e.g. core, not core). 

So would a yaml that implemented a few of these synthetic functions look something like:

```
at_node:
    topographic__elevation:
        synthetic:
           constant:
              value: 2.0
              add_to_all_nodes: False
           white_noise:
              mean: 3.0
              std: 1.0
              add_to_all_nodes: True
           plane:
              point: (0.0, 0.0)
              normal: (1.0, 2.0)
              add_to_all_nodes: False
```

mcflugen:
@kbarnhart How about using a `where` keyword (kind of like `np.maximum`) that is an array of `True` and `False`. Your example *yaml* might then look something like,
```yaml
at_node:
    topographic__elevation:
        synthetic:
           white_noise:
              mean: 3.0
              std: 1.0
              where: is_core
```
or if functions are to be added one after the other maybe,
```yaml
at_node:
    topographic__elevation:
        synthetic:
           constant:
              value: 2.0
              add_to_all_nodes: not is_core
           white_noise:
              mean: 3.0
              std: 1.0
              where: is_core
```

mcflugen:
Or how about dropping the "synthetic" section?
```yaml
at_node:
    topographic__elevation:
       constant:
          value: 2.0
          where: not core
       white_noise:
          mean: 3.0
          std: 1.0
          where: core
    air__temperature:
        filepath: airtemp.nc
```

kbarnhart:
I like the `where` option. 

I don't think we can just write "filepath: name" because we already support options like including a halo and those are important for things like watershed imports. . 

however we might want to support chaining a file path and adding noise. 

e.g. 

```yaml
at_node:
    topographic__elevation:
       from_file:
           filepath: 'file_path.ext'
           no_data: -9999
           halo: 1.0
       white_noise:
          mean: 0.0
          std: 1.0
          where: core
    air__temperature:
        filepath: airtemp.nc
```

mcflugen:
Yeah, I thought about that but I figured if you didn't want any of those options, it would be weird to write:
```yaml
at_node:
    topographic__elevation:
       from_file:
           filepath: 'file_path.ext'
```
How about if something like,
```yaml
at_node:
    topographic__elevation:
       from_file: 'file_path.ext'
```
means, use the argument as the first arg to the function, and then just default keywords?

kbarnhart:
I think the second option is good!

kbarnhart:
we have settled (tentatively) on `landlab.values`

kbarnhart:
This is ready for review. @mcflugen  and @SiccarPoint I think you'd be best suited.

kbarnhart:
ping @SiccarPoint, @mcflugen. Might one of you review this?

mcflugen:
👍 This looks good @kbarnhart. I added some more tests for `_where_to_add_values` and reduced some of the complexity of the function itself. I've taken a liking to this function. Maybe we should add it as a method to `ModelGrid` but rename it (*select*?)? Something like,
```python
grid.select(grid.BC_NODE_IS_CORE, at="node")
grid.select([grid.BC_LINK_IS_FIXED, grid.BC_LINK_IS_FIXED], at="link")
grid.select(at="patch")
```

kbarnhart:
your changes are awesome @mcflugen! Thank you. Merging...!

PR #828, opened 2018-11-05
Full coverage for Flexure component

mcflugen:
This pull request adds tests to `landlab.components.flexure` so that it is now 100% covered. It also fixes some bugs, removes old code, and tidies up some of the interface (for instance, `subside_point_load` is now more general in the arguments it can take) - all of which were found with the new tests. Hooray for tests! 🎉


kbarnhart:
@mcflugen. I dealt with the merge conflicts. I have also verified that this gets flexure to 100%. Nice work! 🎉 💯

Issue #830, opened 2018-11-05
Watershed utility could be made faster by using flow__upstream_node_order


kbarnhart:
This has come up because of time considerations @mcflugen has had with delineating watersheds in Puerto Rico (associated with @christinab's rapid).

@nathanlyons I know you are at GSA

nathanlyons:
Cool! I will review in just a bit.

@mcflugen, this is how I create a colormap for watersheds. Non-watershed cells are colored white. Perhaps this can be a simple plotting function, although generalized to create random colormaps given a number of elements (eg watersheds). 

```python
# Get an array of colors with the length of the max mask id. Change `prism` to use
# other colormaps.
vals = np.linspace(0, 1, max(masks))
colors = plt.cm.prism(vals)

# Create a cmap of the shuffled colors.
np.random.shuffle(colors)
cmap = plt.cm.colors.ListedColormap(colors)

# Grid areas not assigned an id by the watershed util are assigned -1. Color
# these cells  white.
cmap.set_under('white')

imshow_grid(mg, masks, cmap=cmap, vmin=0)
```

PR #831, opened 2018-11-05
Barnhark/make watershed faster

kbarnhart:
Addresses #830 

@mcflugen try this and see if it is faster. 

It should take about as much time as flow accumulator. 

@nathanlyons, your documentation and tests are awesome. They made modifying this (relatively) easy. Let me know if you have any comments.... I **think** the only thing I've done is modify small parts of the guts to make them faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
 os:
 - linux
 - osx
+osx_image: xcode6.4
 sudo: false
 filter_secrets: false
 before_install:
--- a/landlab/utils/__init__.py
+++ b/landlab/utils/__init__.py
@@ -10,4 +10,5 @@ from .source_tracking_algorithm import (
 from .watershed import (get_watershed_mask,
                         get_watershed_masks_with_area_threshold,
                         get_watershed_nodes,
-                        get_watershed_outlet)
+                        get_watershed_outlet,
+                        get_watershed_masks)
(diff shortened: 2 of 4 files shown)


kbarnhart:
@mcflugen try now. There were some small errors before our meeting around four. I think I've fixed them.

mcflugen:
@kbarnhart Thanks! It seems to be working great. It's identifying about 7M watersheds in just a few seconds. :tada:

The problem I'm having now is just trying to plot the watersheds. Ideally, I would want to do something like apply the four color theorem to the map. The problem is that adjacent watersheds have similar IDs and so similar colors so that they blend together and appear as one.

kbarnhart:
Making a 4 color theory-based color RGB array at the raster level sounds challenging as it would require determining adjacency of watershed edge nodes... But you could assign a random number to each watershed and us that as the basis for the cmap. That would be a reasonable first stab.

nathanlyons:
Cool! I will review in just a bit.

@mcflugen, this is how I create a colormap for watersheds. Non-watershed cells are colored white. Perhaps this can be a simple plotting function, although generalized to create random colormaps given a number of elements (eg watersheds).

```python
# Get an array of colors with the length of the max mask id. Change `prism` to use
# other colormaps.
vals = np.linspace(0, 1, max(masks))
colors = plt.cm.prism(vals)

# Create a cmap of the shuffled colors.
np.random.shuffle(colors)
cmap = plt.cm.colors.ListedColormap(colors)

# Grid areas not assigned an id by the watershed util are assigned -1. Color
# these cells  white.
cmap.set_under('white')

imshow_grid(mg, masks, cmap=cmap, vmin=0)
```

kbarnhart:
Thanks @nathanlyons ! Sorry for making those sloppy typos...

Issue #829, opened 2018-11-05
Issues with origin


kbarnhart:
1) We often state that the origin is specified by (x, y) but in fact in needs to be specified by (y, x) .
This needs to be made consistent across the whole package and its documentation (including update origin)

2) Update origin should move origin to specified values. NOT move the origin BY the amount specified. 

3) and an origin property

4) and the origin property needs a setter.

PR #833, opened 2018-11-07
Update origin methods

kbarnhart:
Addresses #829.

So far I've dealt with RasterModelGrid and HexModelGrid. 

Still to go:

- [x] RadialModelGrid

- [x] VoronoiModel Grid

However.... @mcflugen I have a question for you as I do this. 

What exactly is the relationship between the `landlab.graph` package and the `landlab.grid` package? 
It seems like much of the functionality of creating the x-y node points is in both...

### DIFF
--- a/.conda-recipe/meta.yaml
+++ b/.conda-recipe/meta.yaml
@@ -30,7 +30,7 @@ requirements:
 
 test:
   requires:
-    - pytest <3.7.2
+    - pytest
   imports:
     - landlab
   commands:


kbarnhart:
@mcflugen, couple of questions/thoughts.

1) For a raster model grid, the concept of an origin is pretty straightforward. Its the coordinates of the lower left corner. This happens to be node 0. 

2) For a radial model grid, I think the most reasonable origin is the center node (this is the present configuration). 

3) For a Voronoi and Hex, its not as straightforward.

  a) I think the most reasonable thing for a Voronoi is to use the minimum value of x and the minimum value of y. I'd be happy if you have a better idea. 

  b) for a hex, we have to wrangle with a couple of permutations: Shape (hex/rect), orientation (horizontal/vertical) and, I think, size (even-odd, even-even, odd-odd, odd-even). My initial inclination is that the origin is the (y, x) of node 0. However, a simple test I wrote indicates that (as the grids are presently constructed) this is only sometimes the case. 

What do you think makes the most sense? 

```python
from landlab import HexModelGrid

_START_ORIGIN = (10., 20.)
_MOVE_ORIGIN = (30., 45.)
for shape in shapes:
     for orientation in orientations:
         for size in sizes:
             try:
                 mg = HexModelGrid(
                     size[0], size[1], 
                     dx=2.0, origin=_START_ORIGIN, 
                     orientation=orientation, 
                     shape=shape
                 )
                 assert mg._origin == _START_ORIGIN
                 assert mg.x_of_node[0] == _START_ORIGIN[1]
                 assert mg.y_of_node[0] == _START_ORIGIN[0]
                 print(shape, orientation, size, "Passed!")
             except:
                 print(shape, orientation, size, "Failed!")

rect horizontal (8, 7) Passed!
rect horizontal (7, 8) Passed!
rect horizontal (7, 11) Passed!
rect horizontal (8, 12) Passed!
rect vertical (8, 7) Passed!
rect vertical (7, 8) Passed!
rect vertical (7, 11) Passed!
rect vertical (8, 12) Passed!
hex horizontal (8, 7) Passed!
hex horizontal (7, 8) Failed!
hex horizontal (7, 11) Failed!
hex horizontal (8, 12) Passed!
hex vertical (8, 7) Failed!
hex vertical (7, 8) Failed!
hex vertical (7, 11) Failed!
hex vertical (8, 12) Failed!
```

mcflugen:
@kbarnhart The idea of an *origin* only makes sense for certain grids, I think. At least in the way that we use the word "origin" for constructing grids, anyway.

We only use an *origin* for certain grid types that aren't constructed with explicit *x* and *y* coordinates (e.g. a `RasterModelGrid`). For a `RasterModelGrid` we need a *shape*, *spacing*, and *origin* so that we can calculate the *x* and *y* coordinates of the nodes. That is,
```python
y_of_node = np.arange(shape[0]) * spacing[0] + origin[0]
x_of_node = np.arange(shape[1]) * spacing[1] + origin[1]
```
However, for a `VoronoiModelGrid` the user provides the *x* and *y* values so we don't need an *origin* like we do with the other grids.

Once a grid is created, I don't think we need to keep the origin as a property at all. Once it's created, the *x* and *y* values for the nodes provide all the spatial information that's needed.

I think part of the problem here is that the word "origin" has several different meanings. Users might, for instance, think that the *origin* is some point from which a grid's `x_of_node` and `y_of_node` are measured relative to. However, that's not what we mean by *origin*.

If we're going to have something that, like you suggest, points to the minimum *x* and *y* of a grid, it should be the same for all grid types. So, for instance, with a `RadialModelGrid` this would be some point outside of the grid that marks the lower left corner of a bounding box around the grid. If this is what we want, we should use a different name (maybe `ModelGrid.bounding_box` which would be a tuple of (*lower_left_x*, *lower_left_y*, *width*, *height*?).

I guess this is a longwinded way of saying that I'm not all that certain what we mean by `ModelGrid.origin` and how we want to use it? One thing we may want to do with an *origin* property is be able to shift the grid. However, I think that would best done by something like `grid.x_of_node += dx`.

kbarnhart:
@gregtucker, a clarification. I think you are proposing that all grids keep two pairs of coordinates?

(y, x) of projected coordinates and (y, x) of model grid origin. These may be the same in situations when there is no coordinate system (or when near the origin of a UTM grid). But in order to have smaller numbers in the x_of_node, y_of_node, the model grid origin could be translated to (0,0)

mcflugen:
@kbarnhart @gregtucker In this discussion we're talking about a couple different things. What we now call an "origin" is grid-specific. The *origin* keyword just informs the grid class about how to construct `x_of_node` and `y_of_node`. Because voronoi grids are created by specifying *x_of_node* and *y_of_node* they don't have an *origin*.

If we want an "origin" that is a point were *x_of_node* and *y_of_node* are measured relative to, then that's a separate thing and not something we have.

Perhaps we need some new names to make things clearer? Maybe what we call "origin" for each of the grids should be renamed to better reflect what it is. For instance, for a `RasterModelGrid` it could be *lower_left*?

Grid type | New "origin" keyword
--- | ---
`RasterModelGrid` | `lower_left`
`RadialModelGrid` | `center`
`HexModelGrid` | `lower_left`
`VoronoiModelGrid` | N/A

As @kbarnhart pointed out, "lower_left" might be a bit ambiguous for a `HexModelGrid`. I think what we mean is simply, "the coordinates of the first node".

For a `RasterModelGrid` it would look like,
```python
>>> grid = RasterModelGrid((3, 4), spacing=(1., 2.), lower_left=(-1., 0))
>>> grid.lower_left
(-1., 0.)
>>> grid.origin
(0., 0.)
>>> grid = RasterModelGrid((3, 4), spacing=(1., 2.), lower_left=(-1., 0), origin=(-2., 1.))
>>> grid.lower_left
(-1., 0.)
>>> grid.origin
(1., 1.)
>>> grid.x_of_node + grid.origin[1] # The "absolute" x coordinates
array([ 1.,  3.,  5.,  7.,  1.,  3.,  5.,  7.,  1.,  3.,  5.,  7.])
```

kbarnhart:
@mcflugen I think what you've just written makes a lot of sense. 

`ModelGrid.origin` becomes the absolute offset, likely most relevant for issues related to projection, like what @gregtucker raised.

Then each grid type (except Voronoi) has a grid specific keyword that conveys the reference from which the grid is constructed. 

I propose a slight revision to what you've proposed. 

Grid type | New "origin" keyword
--- | ---
`RasterModelGrid` | `yx_lower_left`
`RadialModelGrid` | `yx_center`
`HexModelGrid` |  `yx_node_zero`
`VoronoiModelGrid` | N/A

Two reasons for my revisions: 

First I think that specifying these coordinates as (y, x) instead of (x,y) is nonintuitive and may cause some confusion. By putting `yx` in the kwarg, we emphasize the order in which the values are expected. 

 Second, since the HexModelGrid first node is not always exactly the lower left corner, use a more descriptive name.

I can work on making these changes in this PR. How concerned do you think we should be about breaking backwards?

mcflugen:
@kbarnhart I like adding the `yx` prefix. 👍 I guess if we're going to use a prefix, why not just use `xy` and have it be *(x_of_lower_left, y_of_lower_left)*? We do this elsewhere (I think we have something called `xy_of_node`).

How about if we add an `_of_` (or an `_at_`?) in there to be more consistent? So, `xy_of_lower_left`? Or how about `xy_of_lower_left_node`? Are we getting too long? Yeah, probably.

I'm not sure about *node_zero* though. It doesn't describe *where* that node is. I guess it's clear if you know our node ordering system but, otherwise, it's not, and it's inconsistent with the others. I think *lower_left* is still pretty clear. However, as you showed, we would have to change `HexModelGrid` so that the lower left node really is at the value provided.

This would mean definitely breaking some of our APIs, particularly for `HexModelGrid`. In the case of `HexModelGrid`, there are some other issues that need addressing that will also break things (I need to open another issue about this. It's the main reason we're not using the Graph classes yet in the ModelGrids.).

The new table,

Grid type | New "origin" keyword
--- | ---
`RasterModelGrid` | `xy_of_lower_left`
`RadialModelGrid` | `xy_of_center`
`HexModelGrid` |  `xy_of_lower_left`?
`VoronoiModelGrid` | N/A

What do you think?

kbarnhart:
@mcflugen I think using `xy_` is more intuitive. However, as you have pointed out, it is inconsistent with (nrow, ncol) and (dy, dx) which is the standard order. 

I think it this discussion we started with me arguing for xy, and you countering with yx. Now we've flipped and I would argue for yx and you are arguing for xy. Perhaps this is a question we should raise on our monday call. 

Regarding HexModelGrid... Based on my understanding of the geometry, sometimes there is a true lower left hex node, and other times the leftmost node in the lowest row is not the left-most node.... SO, while I think `lower_left` is better than `node_zero`, I'm not sure I'm happy with it... Not that I have a better idea yet.

mcflugen:
@kbarnhart I was thinking that if we specify `xy` right there in the name, it's clear that it's *(x, y)* and not *(y, x)*. But perhaps not. I agree that we should ask the rest of the group on Monday.

It is true that for some `HexModelGrid` configurations there is no node that is both the most lower and most left. However, there is a node that is the most left node of the most lower nodes.

```
    + - +         + - + - +       + - + - +
   / \ / \         \ / \ /       / \ / \ / \
  + - + - +         + - +       + - + - + - +
 / \ / \ / \       / \ / \       \ / \ / \ /
+ - + - + - +     + - + - +       + - + - +
 \ / \ / \ /       \ / \ /       / \ / \ / \
  + - + - +         + - +       + - + - + - +
   \ / \ /         / \ / \       \ / \ / \ /
    O - +         O - + - +       O - + - +
```
and then vertically oriented grids,
```
+   +   +        +   +
|\ /|\ /|       /|\ /|\
| + | + |      + | + | +
|/|\|/|\|      |\|/|\|/|
+ | + | +      | + | + |
|\|/|\|/|      |/|\|/|\|
| + | + |      + | + | +
|/ \|/ \|       \|/ \|/
O   +   +        O   +
```

margauxmouchene:
I am not sure I understand what 'origin' refers to in the end, but two things:
- From the user viewpoint: nothing in Landlab is called `yx` so the name should be `xy` to avoid confusion. If it is in the name, then the ordering is clear and if the user need `(y,x)` (e.g. for lat, lon)  it's up to them to do the conversion.
- on the `HexModelGrid` problem: my intuition would be to set the origin at the leftmost node of the lowermost row, as described in the last comment by @mcflugen. However, if:

> The origin keyword just informs the grid class about how to construct x_of_node and y_of_node

 that means that some nodes may have a **negative** `x_of_node`. Is that a problem?

kbarnhart:
@margauxmouchene its definitely not a problem that `x_of_node` and `y_of_node` are negative. This happens already in some `HexModelGrids`

As to the coordinates being (x, y), I agree with you that (x, y) is 10000% more intuitive. I think we should go this way unless someone feels otherwise in the call.

The argument that @mcflugen originally made to me was more based on consistency: 
* (nrows, ncols) is in y, x order
* when spacing is specified as a tuple it is supposed to be (y, x)

In fact, existing documentation (in some places but not all) currently states that the origin keyword is (x, y) when it expects (y, x). Other places (e.g. `ModelGrid.move_origin`) requests that the value of the origin be given in (x, y). 

So, right now we are neither consistent, nor correctly documenting the expectations.

gregtucker:
For hex grids, I like the idea of using leftmost of lowest row.

kbarnhart:
@mcflugen I like your idea:

Grid type | New "origin" keyword
--- | ---
`RasterModelGrid` | `xy_of_lower_left`
`RadialModelGrid` | `xy_of_center`
`HexModelGrid` |  `xy_of_lower_left`
`VoronoiModelGrid` | N/A

I've added discussing this to the call document.

SiccarPoint:
Agree with @gregtucker on the leftmost of lowest row, as this creates consistency with our node ordering too (i.e., all else equal, (0, 0) should be node 0!)

Feeling conflicted about (x, y) or (y, x). As long as it's clear, I think either is OK.

kbarnhart:
@mcflugen quick question. I **think** you are are dealing with the HexModelGrids on a different branch? And that means that I can work on finalizing the OTHER model grids on this branch. Correct?

mcflugen:
@kbarnhart I'm dealing with this as it applies to the *Graphs* on a different branch (*HexGraph*, *RadialGraph*, etc.). I'm only dealing with *HexModelGrid* (or any *ModelGrid*) in the *graph-to-grid* branch.

mcflugen:
@SiccarPoint Do you think it's clear that `xy_of_lower_left` means *xy of leftmost node of the lowest row*?

kbarnhart:
A couple of notes from our call.

We decided that raster grid should be specified as:
(row, col), (dx, dy), (x, y)

We would deprecate spacing as (dy, dx) and origin as (y, x)

We would come up with a new name for spacing in RasterModelGrid. 

We would use new (and different) names for each grid's reference point (as discussed above)

And provide a value called, origin_offset as distance from (0, 0) in projected coordinates. This final option is useful when float values are large (as in UTM).

I'm working on this.

kbarnhart:
@mcflugen a question. 

Based on my little test:
```python
from landlab import HexModelGrid

_START_REFERENCE = (10., 20.)
_MOVE_REFERENCE = (30., 45.)


shapes = ["rect", "hex"]
orientations = ["horizontal", "vertical"]
sizes = ((8, 7),
         (7, 8),
         (7, 11),
         (8, 12))
for shape in shapes:
    for orientation in orientations:
        for size in sizes:
            try:
                
                
                mg = HexModelGrid(
                    size[0],
                    size[1],
                    dx=2.0,
                    xy_of_lower_left=_START_REFERENCE,
                    orientation=orientation,
                    shape=shape
                )
                assert mg._xy_of_lower_left == _START_REFERENCE
                assert mg.x_of_node[0] == _START_REFERENCE[0]
                assert mg.y_of_node[0] == _START_REFERENCE[1]
    
                mg.xy_of_lower_left = _MOVE_REFERENCE
                assert mg._xy_of_lower_left == _MOVE_REFERENCE
                assert mg.x_of_node[0] == _MOVE_REFERENCE[0]
                assert mg.y_of_node[0] == _MOVE_REFERENCE[1]
            except ValueError:
                print("Value Error", shape, orientation, size)
                
            except AssertionError:
                print("Assertion Error", shape, orientation, size)
```
I find that the following Hex specifications fail:

```
Value Error hex horizontal (7, 8)
Value Error hex horizontal (7, 11)
Value Error hex vertical (8, 7)
Assertion Error hex vertical (7, 8)
Assertion Error hex vertical (7, 11)
Assertion Error hex vertical (8, 12)
```

I'm presently working on determining the cause of the `ValueErrors`. However, I expect that we won't ever solve the `AssertionErrors` as these are the hex grids for which 

`(x_of_node[0], y_of_node[1] != (0.0)`.

As this is not something we are planning to fix, how do you think I should handle this?


If you have any ideas before I figure it out... the `ValueError` I get is:

```python
  File "landlab/grid/cfuncs.pyx", line 80, in landlab.grid.cfuncs.create_links_at_patch (landlab/grid/cfuncs.c:3090)

ValueError: could not broadcast input array from shape (2) into shape (3)
```

kbarnhart:
Remaining tasks:

- [ ] Tests for all the deprecation warnings I've now created. These should include tests that the raster model grids with spacing specified as (dy, dx) works correctly. 
- [ ] Add the ability to shift a voronoi around (because the other grids can be moved around). 
- [ ] Add documentation about the `origin_offset`.
- [ ] Make sure all the documentation is consistent.

kbarnhart:
@mcflugen (and others!)

In the grid below, which point should be at (0, 0)? A or B? 

Right now we set A to the point, but if we stick with the left most node in the first row, then it needs to be B. 

Does that seem OK?

test hexvertical 9 9

kbarnhart:
@mcflugen and @gregtucker agree with me that it is OK for B to be at (0,0)

mcflugen:
@kbarnhart Instead of `reference_point_coordinates` how about `xy_of_reference`? Having the `xy_` prefix matches other names and is an indicator to the user that it's *(x, y)* not *(y, x)*. I also don't think we need to *"point"* suffix since the *xy_* indicates it's a point (like with `xy_of_lower_left`).

kbarnhart:
sounds good to me. I made that change. This is now passing except for the errors associated with #856

mcflugen:
👍 Looks good. I added some tests and parametrized some others to make it a little cleaner. The only real change I made outside of the tests was to allow `xy_of_lower_left` etc. to be numpy arrays and ensure that the getters return tuples.

kbarnhart:
@mcflugen do you have any ideas as to how to fix the remaining errors? There are five deprecation warnings that are not emitted on py27/windows...?

PR #834, opened 2018-11-09
Remove lint

mcflugen:
Over the years we've gotten a little lax about housecleaning here at the landlab. This pull request is mostly a superficial cleanup of things like trailing whitespace, mixtures of tabs and spaces, unused variables etc. However, it also fixes bugs like, for instance, places where variables were used but never defined (yikes!). It touches **a lot** of files but none of the changes should change any functionality (and all the tests pass, obvs).

There are a couple more things I would like to do to get landlab looking spick-and-span but maybe they should wait for another pull request (like formatting for PEP8) - unless folks think we should just get it done all at once.


mcflugen:
@gregtucker It depends on the method/function. I know I removed one of the methods in the grids because it was duplicated, not in a subclass or anything, but in the very same class (just several thousand lines lower down in the file)! Another one that I removed was completely broken. If it were ever called it would have failed. I think it was just really old code.

mcflugen:
@gregtucker Incidentally, ideally we should probably remove these sorts of `assert` statements all together. `assert` should really just be used for debugging. If someone decides to run landlab with `python -O` (to make it run faster) the above code will not work.
```
$ python --help | grep -A1 -- -O
-O     : remove assert and __debug__-dependent statements; add .opt-1 before
         .pyc extension; also PYTHONOPTIMIZE=x
--
-OO    : do -O changes and also discard docstrings; add .opt-2 before
         .pyc extension
```
This would be a pretty big overhaul, as we use `assert` like this all over landlab. Not many people run python like this, but they could (and maybe they should? I'm not sure what the speedup would be like).

kbarnhart:
Looks like a good idea to me. Any reason not to run black on all of landlab at the same time?

mcflugen:
@kbarnhart I was holding off until I heard back from you to see what you thought. I wasn't sure if it would be too disruptive to people. Too many changes or something. I guess we might as well. That way I can get everything fixed in this one pull request and be done with it.

gregtucker:
@mcflugen the proposed change to celllab_cts.py looks reasonable to me. I'm think @SiccarPoint might have written that bit---does it look familiar, @SiccarPoint ?

gregtucker:
Incidentally, I agree about using assert statements only for cases that would represent bugs (as opposed to inappropriate user input)

mcflugen:
@mcflugen Thanks. I'll wait for the rest of the tests to complete and then merge.

If you have a moment, could you have a look at #828? There's not much to it. I mostly just added a bunch of new tests to flexure to get it to 💯.

kbarnhart:
@mcflugen I'll look at #828 tonight.

PR #839, opened 2018-11-21
Add Python 3.7 support

mcflugen:
This pull request makes *landlab* work with Python 3.7 and adds 3.7 builds to Travis (and drops 3.5).


visr:
Would probably also be good to update the classifiers in `setup.py`, they only indicate support for 2.6 and 2.7.
https://github.com/landlab/landlab/blob/6b8e71b8180ce581b4fc7cf5cd81006c197243ad/setup.py#L172-L173

PR #844, opened 2018-11-29
Fix #843; remove as_id_array cruft

### LINKED ISSUE #843: bug in as_id_array
There's an issue with as_id_array. The function is followed (`core.utils` lns 182-197) by an unindented code block which is running as a meaningless script. I assume this is supposed to be inside the function?? Don't feel confident correcting this myself as I don't know the full intended functionality of the script - sorry to assign you Katy! (Also, it's untested?) Hopefully a very quick fix.

mcflugen:
This pull request removes `as_id_array` cruft from `landlab/utils.py` as described in #843. I've also changed `as_id_array` to return the input array when possible instead of a view. I think that should be ok. There's another doctest, which uses `np.where`, too.

### DIFF
--- a/landlab/core/utils.py
+++ b/landlab/core/utils.py
@@ -179,24 +179,6 @@ def as_id_array(array):
         return np.asarray(array, dtype=np.int)
 
 
-if np.dtype(np.intp) == np.int:
-
-    def _as_id_array(array):
-        if array.dtype == np.intp or array.dtype == np.int:
-            return array.view(np.int)
-        else:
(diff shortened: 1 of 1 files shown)


SiccarPoint:
I'm seeing a ton of weird Travis fails at the moment. Is that one happening in the current build one of them? This looks great to go code-wise once tests clear - feel free to merge it yourself if you get to it before me.

kbarnhart:
I just re-started the incomplete travis runs. Assuming they passes, I think this is OK to merge

Issue #842, opened 2018-11-28
Drainage density issue


jdgemperline:
Hello, I created my own mask of channels to feed into the drainage density component. After running the drainage density component with the mask, it appears that the component re-writes the original mask and deletes small parts of the channels that are originally identified. I have attached two screenshots to demonstrate what this looks like.

I looked at this the other day with @cmshobe, and it looks like the problem area in the component is on line 229 & 230:
        self.channel_network = (
            grid.at_node['channel__mask'].view(dtype=np.uint8))

 @kbarnhart, I think Charlie said you helped write that bit so I tagged you as well.

Thanks!
John Gemperline

Before:
![ih_channels_before](https://user-images.githubusercontent.com/35937621/49175513-f96f3100-f305-11e8-9e23-ca14743ed13f.png)

After:
![ih_channels_after](https://user-images.githubusercontent.com/35937621/49175522-fc6a2180-f305-11e8-94d1-de6b7d49d1c2.png)

kbarnhart:
@jdgemperline thanks for making this issue. 

Could you please provide a minimally complete working example for me to use while addressing this? Here is a stack overflow [link](https://stackoverflow.com/help/mcve) giving a bit more background on what this is. 

Let me know if you have any questions. 

Here is a [link](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code-and-syntax-highlighting) indicating how to do python syntax highlighting in markdown (which is how these comments get formatted).

jdgemperline:
Sure thing:

`import numpy as np
from landlab import RasterModelGrid
from landlab.components.flow_routing import FlowRouter
from landlab.components import FastscapeEroder, DrainageDensity
from landlab.plot import imshow_grid
import matplotlib.pyplot as plt, numpy as np

mg = RasterModelGrid((50, 50), 1.0)
_ = mg.add_zeros('node', 'topographic__elevation')
np.random.seed(50)
noise = np.random.rand(50*50)
mg.at_node['topographic__elevation'] += noise
mg.at_node['topographic__elevation'] # doctest: +NORMALIZE_WHITESPACE
fr = FlowRouter(mg)
fsc = FastscapeEroder(mg, K_sp=.01, m_sp=.5, n_sp=1)
for x in range(100):
    fr.run_one_step()
    fsc.run_one_step(dt = 10.0)
    mg.at_node['topographic__elevation'][mg.core_nodes] += .01

[...]

PR #847, opened 2018-11-30
Addresses DrainageDensity error

kbarnhart:
This PR addresses the issues in DrainageDensity outlined in #842.

Importantly, it now requires that the channel mask is a uint8 dtype

It also
- makes the component fully tested
- expands the documentation
- changes the node processing order in the cython

I'm requesting @mcflugen to review. 

@jdgemperline please check this branch out and verify that it solves the problem you presented in Issue #842.


kbarnhart:
@jdgemperline a thing that you can do that would be really helpful

We have a new test to make sure DrainageDensity doesn't have the behavior you reported in #842 .

At its core is running the following model
```python
mg = RasterModelGrid((80, 80), 1.0)
_ = mg.add_zeros('node', 'topographic__elevation')
np.random.seed(50)
noise = np.random.rand(mg.size('node'))
mg.at_node['topographic__elevation'] += noise
mg.at_node['topographic__elevation'] # doctest: +NORMALIZE_WHITESPACE
fr = FlowAccumulator(mg, flow_director='D8')
fsc = FastscapeEroder(mg, K_sp=.01, m_sp=.5, n_sp=1)
for x in range(100):
    fr.run_one_step()
    fsc.run_one_step(dt = 10.0)
    mg.at_node['topographic__elevation'][mg.core_nodes] += .01
```

It would be ideal if the grid size (currently 80x80) and the number of time steps (100) were as small as possible (while still resulting in the error).

Can you please play around with different grid sizes, numbers of time steps (and possibly random seeds) to identify the smallest possible grid and time step size we can use to reproduce this error

kbarnhart:
Also. I've turned on building these docs so we can see how they look (and to make improvements, and check math formatting) . 

https://landlab.readthedocs.io/en/barnhark-fix_drainage_density_bug/landlab.components.drainage_density.html

jdgemperline:
Just to clarify, the test should check and see if channel nodes are changed only if they're not along an open boundary, correct?

kbarnhart:
@jdgemperline you should use the current version of landlab `master` (that is, not this branch). And you should try and find the smallest grid (and number of time steps) for which some of the nodes ID'd as channel in your original mask are then set from 1 (channel) to 0 (not channel). Does that make sense?

jdgemperline:
Got it! Here's the whole MWE that I used to replicate the issue. I tweaked the grid size, K for the fastscape eroder, time steps, and identifying channels. I didn't get any grids smaller than 10x10 to replicate the issue, though it may be possible if you want to try and make the test even quicker.


'''
import numpy as np
from landlab.components.flow_routing import FlowRouter
from landlab.components import FastscapeEroder, DrainageDensity
from landlab.plot import imshow_grid
import matplotlib.pyplot as plt, numpy as np

size = 10 # Changes grid size and length of noise array
mg = RasterModelGrid((size, size), 1.0) #instantiate a grid
_ = mg.add_zeros('node', 'topographic__elevation')
np.random.seed(3542) # This is the random seed I used to replicate the problem I was having
noise = np.random.rand(size*size)
mg.at_node['topographic__elevation'] += noise #add noise to grid
mg.at_node['topographic__elevation'] # doctest: +NORMALIZE_WHITESPACE
fr = FlowRouter(mg) # Old example used FlowRouter, should work fine with the accumulator
fsc = FastscapeEroder(mg, K_sp=.5, m_sp=.5, n_sp=1) #tweaked K
for x in range(2):
    fr.run_one_step()
    fsc.run_one_step(dt = 10)
    mg.at_node['topographic__elevation'][mg.core_nodes] += .5 # add uplift after timesteps

mask = np.zeros(len(mg.at_node['topographic__elevation'])) # create a mask for drainage density
mask[np.where(mg.at_node['drainage_area'] > 0)] = 1 # setting all nodes as channels makes it more likely the problem appears

#%matplotlib inline
#imshow_grid(mg, mask)
#plt.show()

dd = DrainageDensity(mg, channel__mask=mask)

#imshow_grid(mg, mask)
#plt.show()

mean_drainage_density = dd.calc_drainage_density() #One node towards the bottom left should be altered after
# running this line of code

#imshow_grid(mg, mask)
#plt.show()
'''

kbarnhart:
thanks @jdgemperline !

PR #848, opened 2018-11-30
Create Network Model Grid

kbarnhart:
This PR creates a NetworkModelGrid

It addresses Issue #672 and replaces PR #627.

Currently in progress


kbarnhart:
@mcflugen I think this is ready for you to review. 

A couple of things to note. 

1) I don't know how to make a class have conditional attributes (in fact, this seems antithetical to the intention of a class). So we now have `NetworkGraph` and `Graph`. The latter inherits from the former. I think I've got all the de-duping done, but you might have a way to improve the inits. 

2) We have a way to read from shape file but no way to write to shapefile. I've made a note in the docs that if someone wants this they should make an issue. Since we write to netcdf, I don't think it makes sense to make a method that (to the best of my knowledge) no one is planning to use. 

3) [Docs page is here](https://landlab.readthedocs.io/en/barnhark-nmg_third_attempt/landlab.grid.network.html)

mcflugen:
@kbarnhart Looks good. I made only one tiny change. I was going to do more but then decided to do it in another pull request. I think we should rename some of the graph classes. For instance, what you call a *NetworkGraph* really is just a "[graph](https://en.wikipedia.org/wiki/Graph_(discrete_mathematics))" (or maybe a "directed graph" or "tree graph"). What we had been calling a *Graph* (that is, nodes, links, *and* patches) is something more.

Anyway, this is good to go. We can talk about the renaming in another pull request.

kbarnhart:
I just thought of at least one thing needs to change... writing it down so
I don’t forget.

The shapefile import should allow a grid to be specified like the other io
functions.

I’ll do that when my computer is on next.

Issue #838, opened 2018-11-20
Reading DEM data directly from OpenTopography


gregtucker:
This would be a useful functionality to have. Their REST API is:
https://opentopography.org/developers
and TopoToolbox allows this:
https://blogs.mathworks.com/loren/2018/10/16/introducing-the-topotoolbox/

PR #850, opened 2018-12-06
Create field from OpenTopography

kbarnhart:
This PR is in progress. It will address #838.


SiccarPoint:
IIRC I've tried Python-specific gdal modules before, but they all seem to be wrappers for existing OS gdal installs that the user does separately - and which are a right pain to get set up correctly. Is that no longer the case? That would be awesome.

Issue #861, opened 2018-12-22
BMI can't wrap Lithology


kbarnhart:
```
landlab/bmi/components.py:12
  /Users/barnhark/projects/landlab/landlab/landlab/bmi/components.py:12: UserWarning: unable to wrap class Lithology
    warnings.warn("unable to wrap class {name}".format(name=cls.__name__))
  /Users/barnhark/projects/landlab/landlab/landlab/bmi/components.py:12: UserWarning: unable to wrap class LithoLayers
    warnings.warn("unable to wrap class {name}".format(name=cls.__name__))
```

@mcflugen will you help me fix this?

PR #862, opened 2018-12-23
Fix issue 861

mcflugen:
This pull request fixes issue #861. The `Lithology` (and so `Litholayers`) component could not be wrapped with a BMI because it did not inherit from `landlab.Component`. This pull request adds:
1. `Component` as a parent to `Lithology`, and
1.  a couple tests to ensure that both `Lithology` and `Litholayers` can be wrapped with a BMI using `wrap_as_bmi`.


kbarnhart:
thanks @mcflugen!

Issue #867, opened 2018-12-28
set_watershed_boundary_condition methods should be able to take field name


kbarnhart:
If they can take a field name, they will be compatible with the forthcoming magical constructor of #865.

In the PR that addresses this LLCATS info should get added to the HexModelGrid watershed methods so they show up in the documentation correctly.

PR #870, opened 2018-12-28
Allow watershed methods to take a field name

kbarnhart:
Addresses #867 

@mcflugen I did this by adding a new method to `ModelDataFieldsMixIn`.

My thinking was that I want something like the `use_field_name_or_array` decorator that can work for model grids. 

I didn't want to make an `if isinstance(field, stringtypes` block in all six watershed methods (and then to test them all). 

Note, I did not use `np.array(field).flatten()` because that returns a copy of the array.


kbarnhart:
ping @mcflugen. Might you review this?

Issue #866, opened 2018-12-27
read_netcdf can't add values to an existing grid


kbarnhart:
It would be good to be able to provide a grid (as in read_esri_ascii) rather than always have a grid created. 

It would also be good for read_netcdf to be able to add a halo of no-data values.

Issue #849, opened 2018-12-04
Best practice for no-data values


kbarnhart:
@mcflugen , I'm working on a PR to address #838 and I've run across an issue I'd like to discuss.

When I pull the DEM data from OpenTopography, I get a no-data value. I don't want to immediately set this no-data value to nan. I want to attach it to the grid, but I might expect that different at-node fields have different no-data values. This raises the question, what is the correct place to store this information. 

My recommendation: store such a no-data value as an xarray variable attribute. I think this would be best because it would travel with the variable when it is written to netcdf. 

What do you think about this?

methods that use no-data values (e.g. set_no_data_to_closed) could check for this attribute and use it.

kbarnhart:
@mcflugen and @gregtucker agree that this should be stored as an xarray variable attribute.

PR #873, opened 2019-01-11
Add halo and add-to-grid for netcdf

kbarnhart:
This PR addresses #866

### DIFF
--- a/landlab/io/esri_ascii.py
+++ b/landlab/io/esri_ascii.py
@@ -434,25 +434,8 @@ def read_esri_ascii(asc_file, grid=None, reshape=False, name=None, halo=0):
 
     data = np.flipud(data)
 
-    # REMEMBER, shape contains the size with halo in place
-    # header contains the shape of the original data
-    # Add halo below
     if halo > 0:
-        helper_row = np.ones(shape[1]) * nodata_value
-        # for the first halo row(s), add num cols worth of nodata vals to data
-        for i in range(0, halo):
-            data = np.insert(data, 0, helper_row)
-        # then for header['nrows'] add halo number nodata vals, header['ncols']
-        # of data, then halo number of nodata vals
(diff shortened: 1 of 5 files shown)


mcflugen:
👍 Looks good @kbarnhart. I added some tests for `add_halo` and also changed it's signature. I hope you're ok with it.

* I removed the `shape` argument. I think the user should just reshape the input array to what it should be (and now the only return value is the new array with the correct shape),
* `halo` is now a keyword with the default being `1`,
* changed `nodata_value` to `halo_value` as I thought that was more descriptive, I also made it a keyword,
* the default for `halo_value` is `None`, which means don't initialize the halo values.

What do you think?

kbarnhart:
@mcflugen I have a couple of thoughts/questions:

1) I think the default halo should be of size 0, not 1. This is to keep consistency with `read_esrii_ascii` (and current behavior). 

2) I think it should stay as `nodata_value` (for consistency with the `read_esrii_ascii` function). Also, because eventually I think we'll handle the `nodata_values` better (see Issue #849). 

What do you think?

mcflugen:
1. Having the default be `0` is fine, although for a function called `add_halo` I think it might be a bit surprising to have `add_halo(data)` not actually add a halo.

2. I guess I'm thinking about `add_halo` a little more generally than what it's being used for in this pull request. I don't think of the halo values as being *nodata*. As an example of them truly not being *nodata*, in the future, we could add additional ways of setting the halo values (for instance the fill values could be chosen to maintain some gradient, or wrap-around boundary conditions).

Right now, `add_halo` is only being used in `read_esri_ascii` and `read_netcdf` but I think it could be used much more generally.

kbarnhart:
@mcflugen I totally misunderstood what you meant in the first comments... 🤦‍♀️ 

I like how you have modified `add_halo`! Assuming it passes, I think its good!

Issue #888, opened 2019-01-29
Unnecessary(?) NetCDF import tests


kbarnhart:
We have a bunch of code that tests whether NetCDF was imported correctly. This exists in case a user doesn't have the NetCDF4 library installed.

The way we handle this masks errors in terrainbento. I've made a PR that removes these import tests.

PR #889, opened 2019-01-29
removing netcdf exceptions

kbarnhart:
Address #888


kbarnhart:
@mcflugen feel free to use this in your work or to indicate that you've dealt with these exceptions and we can close this PR without merging.

Issue #892, opened 2019-02-04
Improvements based on profiling FlowAccumulator


kbarnhart:
I'm doing a project that requires running Landlab on a 200x200 grid with Stream Power. I was surprised by how long time steps took, so I profiled the code:

```
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    49099    0.663    0.000 7882.546    0.161 /Users/barnhark/projects/landlab/landlab/landlab/components/flow_accum/flow_accumulator.py:1201(run_one_step)
    49099    7.419    0.000 7881.882    0.161 /Users/barnhark/projects/landlab/landlab/landlab/components/flow_accum/flow_accumulator.py:1101(accumulate_flow)
    49099   13.208    0.000 5271.519    0.107 /Users/barnhark/projects/landlab/landlab/landlab/components/flow_accum/flow_accum_bw.py:213(make_ordered_node_array)
    98198 4685.405    0.048 4687.355    0.048 /Users/barnhark/projects/landlab/landlab/landlab/components/flow_accum/flow_accum_bw.py:164(_make_array_of_donors)
 44370995   17.367    0.000 2897.285    0.000 /Users/barnhark/projects/landlab/landlab/landlab/components/flow_accum/flow_accum_bw.py:56(add_to_stack)
 44370995 2879.918    0.000 2879.918    0.000 {built-in method landlab.components.flow_accum.cfuncs._add_to_stack}

```

The function `_make_array_of_donors` took a very large proportion of time (and happens to not yet be cythonized). 

Forthcoming is a PR that cynhonizes it.

kbarnhart:
FYI @nicgaspar @nathanlyons

kbarnhart:
Re profiling with changes in PR #893 gives

```
 ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    61502    0.929    0.000 4839.123    0.079 /Users/barnhark/projects/landlab/landlab/landlab/components/flow_accum/flow_accumulator.py:1201(run_one_step)
    61502   10.896    0.000 4838.194    0.079 /Users/barnhark/projects/landlab/landlab/landlab/components/flow_accum/flow_accumulator.py:1101(accumulate_flow)
    61502   18.626    0.000 4410.193    0.072 /Users/barnhark/projects/landlab/landlab/landlab/components/flow_accum/flow_accum_bw.py:180(make_ordered_node_array)
[...]

PR #893, opened 2019-02-04
Cythonize parts of flow routing

kbarnhart:
This PR addresses #892.

FYI @nicgaspar @nathanlyons

### DIFF
--- a/landlab/components/flow_accum/cfuncs.pyx
+++ b/landlab/components/flow_accum/cfuncs.pyx
@@ -100,3 +100,23 @@ cpdef _make_donors(DTYPE_INT_t np,
         ri = r[i]
         D[delta[ri] + w[ri]] = i
         w[ri] += 1
+
+
+@cython.boundscheck(False)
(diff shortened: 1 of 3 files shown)


kbarnhart:
@SiccarPoint I believe the reason the coverage dropped is that I reduced the number of lines of code in `landlab/components/flow_accum/flow_accum_to_n.py` but didn't change the _true_ coverage of the functionality. Because I reduced lines, the proportion of non covered lines increased. Looking at [`flow_accum_to_n.py`](https://coveralls.io/builds/21439542/source?filename=landlab/components/flow_accum/flow_accum_to_n.py) on coveralls shows that lines 665 and 666 are the two uncovered lines (they have to do with passing boundary nodes in and setting them to have drainage area to zero). 

I would expected coverage to drop in `flow_accum_bw.py` as well because a similar thing happened in this file (and it has the same two uncovered lines). 

Assuming this passes now that I've updated this branch still OK to merge?

PS. Thank you for looking at these PRs. 🎉

SiccarPoint:
Sounds good.

PR #902, opened 2019-03-19
Gt/clean celllab code

gregtucker:
Cleans up celllab_cts.py and related, removing unused Python functions that were translated to Cython, and some older functions that were superseded.


gregtucker:
Improved the coverage of the core CA files, so hopefully the coverage test will now pass too.

landlab/components/flow_accum/flow_accumulator.py
```diff
@@ -1136,7 +1136,7 @@ def accumulate_flow(self, update_flow_director=True):
                 self.depression_finder.map_depressions()
 
                 # if FlowDirectorSteepest is used, update the link directions
-                if self.flow_director._name is "FlowDirectorSteepest":
+                if self.flow_director._name == "FlowDirectorSteepest":
                     self.flow_director._determine_link_directions()
```
kbarnhart: @gregtucker I think all these comparison changes improve the code, but perhaps they belong in another PR. @mcflugen what do you think?

gregtucker: Travis tests were failing because of flake8 linter issues, including the above, so in the interest of expediency I went ahead and fixed them.

PR #903, opened 2019-03-22
Gt/fix load grid

gregtucker:
This PR modifies the __setitem__ and related functions in ScalarDataFields to handle the case of missing _size and _units variables. The issue cropped up when reading in a native Landlab grid containing a HexModelGrid. For reasons I haven't discovered, both of these attributes were missing---suggesting that the field object hadn't been initialized properly. The solution is something of a hack: it wraps a try-except around attempts to access _size and _units in __setitem__, and creates _size or _units when missing. It assumes the exception arises BECAUSE the variable is missing.


kbarnhart:
@gregtucker, I have noticed you've requested a review, but I'm going to wait until everything passes first.

gregtucker:
Tests now passing.

kbarnhart:
@gregtucker I think @mcflugen should take a look at this. 

As best as I can tell, the issue is that `ScalarDataFields` is not being initialized properly in `HexModelGrid` as the `ScalarDataFields.__init__` is where `_units` and `_size` get defined. 

I tried to determine why this is happening, but the nature of the use/inheritance of all the types of fields was really challenging to follow. `ScalarDataFields` are used by `ModelDataFields` (but are not inherited), and then `ModelDataFieldsMixIn` inherits from `ModelDataField`. `ModelGrid` inherits from `ModelDataFieldsMixIn`. `Both RasterMG` and `VoronoiMG` inherit from `ModelGrid`, and `HexMG` inherits from `VoronoiMG`.

I couldn't find where in that chain the issue is occurring...

I suspect @mcflugen will have the right idea, though. 

If a hack is necessary, I think that your hack is great!

gregtucker:
I actually consulted with @mcflugen and he pointed me to editing __setitem__. There may well be a less superficial solution, but the current fix appears sufficient to allow reading in of native landlab format hex grids that were generated from a 4-day-long series of model runs.

Issue #912, opened 2019-04-17
No utility to convert from map degrees to math radians


kbarnhart:
A common issue is the conversion from compass degrees (0-360, clockwise from north) to math radians (0-2pi, counter clockwise from east). 

landlab/core/utils.py file has  a function to convert from math radians to compass degrees. It does not have the reverse function. 

A need for such a function is required by work being done by @hjgray10. 

A forthcoming PR will address.

PR #913, opened 2019-04-17
add degree to radian function

kbarnhart:
address #912


kbarnhart:
Could someone please review this tiny PR.

kbarnhart:
np.radians doesn't change the reference frame (0 at north vs east), or the direction (CC vs CCW). merging.

Issue #864, opened 2018-12-27
Standardizing HexModelGrid and RasterModelGrid init


kbarnhart:
Right now the way that the two types of grids that use a "shape" are init is different.
This should be standardized in Landlab 2.0.

RasterModelGrid can take either rows and columns as individual arguments, or a shape tuple (rows, cols) for the first argument. For example:

```python
mg = RasterModelGrid(3,4)
```
vs 
```python
mg = RasterModelGrid((3,4))
```
HexModel grid, doesn't permit a shape tuple. 
```python
mg = HexModelGrid(3,4)
```
I don't have a strong feeling regarding which of these is better than the other. However, I do think that in Landlab 2.0 only one option should be supported and it should be consistent. If I had to choose, I'd lean towards providing a shape tuple. 

Tagging all, so that it can be discussed (and eventually I will make a PR deprecating the correct things).

@gregtucker @mcflugen @SiccarPoint @nicgaspar @nathanlyons @margauxmouchene @erkanistan @saisiddu

gregtucker:
We definitely should be consistent among grid types (at least where it makes sense). I too lean toward having a shape tuple, simply because it is more consistent with Numpy arrays.

mcflugen:
I also agree. The length of the tuple should be that of the number of grid dimensions. So, two for now but, when we introduce 1D (which we kind of already have) and 3D grids, the size of the tuple will range from one to three. This will make things easier as we won't have to add a new argument to the grid constructors.

I think the same should be done for radial grids. The *shape* tuple should be something that defines the grid in the *r* and *theta* dimensions.

gregtucker:
Steps toward this:
- accept a tuple as 1st positional argument (and in that case handle second and third appropriately)
- accept node_layout as alternative keyword arg for ('hex' or 'rect')
- deprecate use of "shape" when it refers to a string instead of tuple

kbarnhart:
Agree with @gregtucker 's steps. I'm turning the bullets into checks so we can keep track of what needs to be done.

PR #914, opened 2019-04-19
Gt/hex grid args

gregtucker:
This PR adds a keyword argument "node_layout" to HexModelGrid. This new keyword replaces "shape", which can now be repurposed as a 2-element tuple with (rows, cols), consistent with RasterModelGrid and Numpy usage. This change implements the following behavior that maintains backward compatibility as follows:

1) If user sends shape as a str ('rect' or 'hex'), it is interpreted as node_layout, and a deprecation warning is given.
2) If user sends shape as a type, it is interpreted as (base_num_rows, base_num_cols) (and overrides the corresponding positional arguments.

Note that this PR does NOT handle the case of a user sending a tuple as the first positional argument. In other words, if they want to pass tuple, they have to use the (now modified) shape keyword.

This PR addresses most of issue #864

### DIFF
--- a/landlab/grid/hex.py
+++ b/landlab/grid/hex.py
@@ -42,9 +42,11 @@ class HexModelGrid(VoronoiDelaunayGrid):
     orientation : string, optional
         One of the 3 cardinal directions in the grid, either 'horizontal'
         (default) or 'vertical'
-    shape : string, optional
+    node_layout : string, optional
         Controls the shape of the bounding hull, i.e., are the nodes arranged
         in a hexagon, or a rectangle? Either 'hex' (default) or 'rect'.
+    shape : tuple of 2 int
+        Alternative way to specify (base_num_rows, base_num_cols)
 
     Returns
     -------
@@ -57,7 +59,7 @@ class HexModelGrid(VoronoiDelaunayGrid):
     have 2 nodes, and the second nodes.
 
     >>> from landlab import HexModelGrid
-    >>> hmg = HexModelGrid(3, 2, 1.0)
+    >>> hmg = HexModelGrid(shape=(3, 2), dx=1.0)
     >>> hmg.number_of_nodes
     7
     """
@@ -250,8 +252,9 @@ class HexModelGrid(VoronoiDelaunayGrid):
         # (note: in LL2.0, shape should refer ONLY to rows and columns)
         if type(shape) is str:  # "old" (LL < 2.0) usage
             node_layout = shape
-            warn(message="Use node_layout to specify 'rect' or 'hex' layout",
-                 category=DeprecationWarning)
+            message = ("Use node_layout to specify 'rect' or 'hex' layout "
+                       + "(usage will be enforced in Landlab 2.0+).")
+            warn(message=message, category=DeprecationWarning)
         elif type(shape) is tuple:
             (base_num_rows, base_num_cols) = shape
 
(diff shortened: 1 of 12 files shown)


gregtucker:
(note typo above: "type" = "tuple")

kbarnhart:
@gregtucker awesome will take a closer look!

@gregtucker & @mcflugen since this changes the meaning of `shape` I think that this means this is the first PR that needs to go into our soon-to-be-created LL2.0 branch?

If so, then we need another branch that JUST makes the name change and yields a deprecation warning... (but doesn't repurpose shape).

gregtucker:
Thanks @kbarnhart. Actually this PR is for the current LL, and should be fully backward compatible. 

LL 2.0 will need to add a shift in the positional args, from (NR, NC, dx, ...) to ((NR, NC), dx, ...)

gregtucker:
To be clear, you can still do 

shape='rect' (or 'hex')

and it will be interpreted correctly (but send a deprecation warning). You can alternative do

shape=(nrows, ncols)

and that too will be interpreted correctly, as will

node_layout='rect' (or 'hex')

landlab/grid/hex.py
```diff
@@ -118,6 +119,14 @@ def __init__(
...
         7
         """
+
+        # Pre-LL2.0: handle shape and node_layout keyword args
+        # (note: in LL2.0, shape should refer ONLY to rows and columns)
+        if type(shape) is str:  # "old" (LL < 2.0) usage
+            node_layout = shape
+        elif type(shape) is tuple:
+            (base_num_rows, base_num_cols) = shape
+
```
kbarnhart: @gregtucker I think if we want this to be in current LL (as opposed to LL2.0) we need to not have lines 127 and 128... Doesn't that permit someone to pass shape as tuple and then have it used as the nr and nc?

kbarnhart:
@gregtucker yes, the functionality you pointed out is what I think needs to change. 

It is backward compatible...but also forward compatible. Since it is the same keyword argument this seems not-correct. 

But perhaps you or @mcflugen  have better founded opinions on how to implement this sort of backward-forward compatibility...

landlab/grid/hex.py
gregtucker: Yes, but I don't think that's a problem. It supports both old usage, when shape is a str, and new usage, when shape is a tuple. In other words, it should still be fully compatible with old usage, i.e., is still backward compatible.

landlab/grid/hex.py
```diff
@@ -235,24 +246,33 @@ def _initialize(
...
 
+        # Pre-LL2.0: handle shape and node_layout keyword args
+        # (note: in LL2.0, shape should refer ONLY to rows and columns)
+        if type(shape) is str:  # "old" (LL < 2.0) usage
+            node_layout = shape
+            warn(message="Use node_layout to specify 'rect' or 'hex' layout",
+                 category=DeprecationWarning)
+        elif type(shape) is tuple:
+            (base_num_rows, base_num_cols) = shape
+
```
kbarnhart: @gregtucker my only comments are to:

- [ ] Add a statement to the deprecation warning that indicates that this change will be enforced in LL2.0, and that shape will have a new meaning in LL2.0
- [ ] Add a test that the deprecation warning is raised correctly.
- [ ] Add a test that if shape is passed a tuple, it behaves as expected. 

I couldn't find tests in the PR, and I double checked that [coveralls indicates](https://coveralls.io/builds/22923077/source?filename=landlab/grid/hex.py) that those lines are still untested.

gregtucker: Good ideas, thanks---now added.

Issue #911, opened 2019-04-16
requirements.txt for a developer


SiccarPoint:
We don't enforce enough requirements.txt for a developer install. Both `pytest-cov` and (apparently; see #910) `pyqt` are essential to run the tests locally, so should be in the file. Could consider a separate file for developers, but don't really see why we wouldn't just do this for everybody.

I note the `pyqt` thing is tricksy, as it raises cryptic c++ library errors. This makes it doubly important.

kbarnhart:
I month or so ago I was discussing a related topic with @mcflugen. Namely, that requirements are listed in multiple places (the requirements.txt file, the conda recipe, etc). 

I got the impression from @mcflugen that it was more appropriate to just put the dependencies for _running_ rather than _building_ and _running_ and _testing_ in the requirements.txt file (please correct me if I misremember)... 

Another type of package to consider is something like `flake8`... it is functionally necessary to test if the code is lint-compatible...

`isort` and `black` are necessary if you want to take advantage of `make pretty`... 

The conda recipe (meta.yaml) breaks some of these categories out (which I find helpful). 

I think it would be great if we could simultaneously:
a) clean up/add to the requirements.txt file
b) remove unnecessary bits of recipes we are not really using anymore (due to conda forge)
c) be clear about what is needed to build/run/test...

kbarnhart:
as @mcflugen addressed this in #915 I'm closing this. 

I also updated some of the information on Updating and Installing (dev version) on the wiki indicating what to do if setup.py or requirements-dev change.

Issue #910, opened 2019-04-15
Terrible cython issues, redux


SiccarPoint:
OK, so I've just attempted to clean up Python on my machine, and something has gone seriously awry with landlab + gcc + cython + conda.

Workflow:
* Full Anaconda 2 uninstall (`anaconda-clean` + hand removing necessary packages)
* GUI install of current version Anaconda 3
* `conda install --file=requirements.txt`

Then
```
$ python setup.py develop
running develop
running egg_info
writing landlab.egg-info/PKG-INFO
writing dependency_links to landlab.egg-info/dependency_links.txt
writing entry points to landlab.egg-info/entry_points.txt
writing top-level names to landlab.egg-info/top_level.txt
reading manifest file 'landlab.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no files found matching 'ez_setup.py'
warning: no files found matching 'README.md'
writing manifest file 'landlab.egg-info/SOURCES.txt'
running build_ext
skipping 'landlab/layers/ext/eventlayers.c' Cython extension (up-to-date)
skipping 'landlab/ca/cfuncs.c' Cython extension (up-to-date)
skipping 'landlab/graph/cfuncs.c' Cython extension (up-to-date)
skipping 'landlab/graph/hex/ext/hex.c' Cython extension (up-to-date)
skipping 'landlab/graph/sort/ext/spoke_sort.c' Cython extension (up-to-date)
building 'landlab.graph.sort.ext.spoke_sort' extension
gcc -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I/Users/danhobley/anaconda3/include -arch x86_64 -I/Users/danhobley/anaconda3/include -arch x86_64 -I/Users/danhobley/anaconda3/lib/python3.7/site-packages/numpy/core/include -I/Users/danhobley/anaconda3/include/python3.7m -I/Users/danhobley/anaconda3/include/python3.7m -c landlab/graph/sort/ext/spoke_sort.c -o build/temp.macosx-10.7-x86_64-3.7/landlab/graph/sort/ext/spoke_sort.o
In file included from landlab/graph/sort/ext/spoke_sort.c:266:
In file included from /Users/danhobley/anaconda3/lib/python3.7/site-packages/numpy/core/include/numpy/arrayobject.h:4:
[...]

PR #915, opened 2019-04-22
Tidy up CI on Travis

mcflugen:
This pull request mainly cleans up the landlab CI on Travis. In doing so, it may also address some issues raised #910, and #911 (i.e. ensuring when we build with Python 3.7, we really build with 3.7 and not downgrade to 3.6).

Some highlights:
* Break up the build on Travis into stages: test, lint, docs, deploy (to PyPI)
* Add requirements to *setup.py* to make landlab pip-installable
* For testing, install Python with conda but install landlab with pip. We hadn't been testing pip installs before. Our conda installs will still be tested with conda-forge
* On success, deploy to PyPI. We had only been doing this on Windows. Somewhere along the way we stopped deploying to PyPI from Travis

I still need to clean up the requirements files. I think `requirements.txt` maybe should have packages not strictly needed by landlab (i.e. those in setup.py) but packages required for developers (*pytest*, *flake8*, etc.). There should also be a conda environment file for landlab developers that want to use conda as a package manager instead of pip.


kbarnhart:
🎉

kbarnhart:
@mcflugen a couple of things:

As we discussed earlier today, this PR should get rid of the Win32  bit builds. 
- [ ] Remove Win32 bit builds from AppVeyor

It also doesn't seem to fix the #911 issue. Instead it extends `setup.py`  to add additional requirements, adds an `environment.yaml` and doesn't add any of the things  @SiccarPoint wanted to add to the `requirements.txt` file (e.g. pytest, pytest-cov, flake8). 

I don't have a problem with adding an `environment.yaml` file, but I think the following should be done:

- [ ] All the conda build related stuff should get removed in this PR. It should only live in the Conda-forge repo  where we build that recipe.
- [ ] Both `environment.yaml` and `requirements.txt` should include EVERYTHING necessary for a developer to build, test, evaluate lint, and calculate coverage.

@SiccarPoint do  you have any additional comments.

kbarnhart:
@mcflugen this looks great to me! 

My only question is: Do you think it makes sense to get rid of the `requirements.txt` file? 

My vote is yes. In an `environment.yaml` we are providing the minimally sufficient conditions to run/test/dev landlab. If someone wants to run landlab/master outside of this conda environment, they have a list of all the packages they need. By providing the requirements in ONLY `environment.yaml` we are making it (a) super easy to create a functional conda environment and (b) putting the requirements only in one spot. 

As it stands, this PR would need to be accompanied by:
 - [ ] changes to the wiki developer install instructions (including how to tell if the conda env needs to be updated...) (check box sorta silly as these changes only should occur after merge to master).

Perhaps other landlabbers (@gregtucker @SiccarPoint @saisiddu @nicgaspar @nathanlyons @erkanistan) want to weigh in on changes to `requirements.txt` as well.

mcflugen:
@kbarnhart I removed the Windows 32 builds and cleaned up the build process on AppVeyor in the process. It now uses *pip* to build landlab, which meant that I was able to remove the conda build stuff, as you suggested. Unfortunately, *netcdf4* seems to not be *pip* installable on Windows so I have to install it with conda for the AppVeyor builds.

I added a *environment-dev.yml* file for a complete installation environment for developing with landlab. I removed *requirements.txt* to avoid repeating the requirements, which are also in *environment-dev.yml* and also in *setup.py*. I added a new requirements file with just the extras needed to develop with landlab (*pytest*, *black*, etc.). Since the install requirements for landlab are now in *setup.py*, I didn't think we also needed them in *requirements.txt*.

To install a development version of landlab you could do one of the following,
```bash
$ pip install -e . -r requirements-dev.txt
```
or
```bash
$ conda env create --file=environment-dev.yml
```

kbarnhart:
@mcflugen 🎉 I think this looks great! I really like how its been cleaned up and the splitting of requirements between `setup.py` and `requirements-dev.txt`. It makes a lot of sense to me and seems much cleaner. 

Am I crazy to think that the following is what is needed to do a full conda install?

```bash
$ conda env create --file=environment-dev.yml
$ conda activate landlab_dev
$ python setup.py develop
```

I also just looked through the Travis builds and I noticed that the coveralls command doesn't seem to be working. 
- [ ] fix coveralls command

kbarnhart:
Looks like the coveralls issue is fixed. For example: https://coveralls.io/jobs/47945505

kbarnhart:
@mcflugen Should I make a new issue and assign you related to updating the install and update instructions on the wiki?
Or is mentioning it here sufficient?

PR #916, opened 2019-04-26
make pretty

kbarnhart:
@mcflugen recently make pretty started to change more than just the files I have been working on (and thus have formatting errors). 

So this PR brings landlab master up to date with current make pretty standards.


kbarnhart:
@mcflugen it looks like this passes. What do you think?

PR #919, opened 2019-04-29
Update BMI wrapping using bmipy package

mcflugen:
This pull request updates the *BMI* wrapping of landlab components to use the latest *BMI* specification as provided by the *bmipy* package. Most of the changes are fairly minor. The biggest one is to a component's input file (that is, the file passed to the BMI initialize function). With this pull request, a component's parameters are passed in a section whose name is that of the component (converted to snake case). So, for example,
```yaml
flexure:
  eet: 15.e+3
clock:
  start: 0
  stop: 100.
  step: 2.
grid:
  type: raster
  shape: [20, 40]
  spacing: [1000., 2000.]
```
Previously, in the above example, *eet* would have been at the same level as *clock* and *grid*. Having a component's parameters in a separate section allows us to use a single initialization file to initialize multiple components that want to share a clock and a grid.


kbarnhart:
@mcflugen I think the Win32 changes should go over in #915 (or in there own  PR), but not here. 

Does it make sense to add support for multiple  grid types  and/or the `create_grid` function now? 

Also, (perhaps I just can't find this) but where should I be able to see the expectations for the input file format?) I think a little more than the examples in initialize and in the docstring would be good. For example, does the name "flexure: ..." need to be the lower case version of the component, or would "Flexure" work?

kbarnhart:
@mcflugen I was thinking more about this earlier today. 

I think it makes sense to switch now to multiple grid types and the use of `create_grid`. 

Something like the following should work with the PyMT Jinja templates and with landlab. 

```yaml
grid:
  grid: {{ grid | to_pretty_yaml }}
  fields:
    at_node:
      topographic__elevation: {{ topographic__elevation | to_pretty_yaml }}
      additional_field_name : {{ additional_field_name | to_pretty_yaml }}
  boundary_conditions: {{ boundary_conditions  | to_pretty_yaml }}
```

Within the code, replacing the lines

```python
grid_params = params.pop("grid")
gtype = grid_params.pop("type")
if gtype == "raster":
    cls = RasterModelGrid
else:
    raise ValueError("unrecognized grid type {gtype}".format(gtype=gtype))
grid = cls.from_dict(grid_params)
```

with
 
```python
grid_params = params.pop("grid")
grid = create_grid(grid_params)
```

will accomplish the task of switching from the current implementation to using `create_grid`.

The only additional thing that would need doing is 
- [ ] adding grid type-testing and various implemented/not implemented errors to the grid shape/connectivity type BMI attributes. 

what do you think?

mcflugen:
@kbarnhart I agree, it probably does make sense to make use of `create_grid`. How entrenched are we (you) in the current format for the serialization of grids that `create_grid` understands? I think there may be some tweaks that I would like to make.

kbarnhart:
@mcflugen I think you should make the tweaks. Technically the terrainbento 1.x versions need `create_grid` in its current form. But I can indicate that the 1.x versions use a particular set of landlab versions and make the PyMT compatible version of terrainbento the 2.0 release.

mcflugen:
@kbarnhart Yes, please have a look. I think I'm pretty much done with major changes now. I'm mainly trying to get things working with Python 2.7 now (I'll be so happy when we leave Python 2 behind !).

landlab/bmi/bmi_bridge.py
```diff
@@ -260,7 +263,7 @@ def initialize(self, fname):
             clock_params = params.pop("clock")
             self._clock = TimeStepper(**clock_params)
```
kbarnhart: @mcflugen I'm confused. Isn't ~L261 the place where we need to use the new `create_grid`?

landlab/bmi/bmi_bridge.py
```diff
@@ -310,42 +313,88 @@ def get_value_ref(self, name):
...
+            return self._base.grid.number_of_nodes
+
+        def get_grid_nodes_per_face(self, grid, nodes_per_face):
+            raise NotImplementedError("get_grid_nodes_per_face")
+
+        def get_grid_size(self, grid):
+            return self._base.grid.number_of_nodes
+
+        def get_grid_x(self, grid, x):
+            return self._base.grid.x_of_node
```
kbarnhart: A general BMI question. Since uniform rectilinear grids don't _need_ `get_grid_x`.... Is it a problem that it is implemented? Or is it OK?

Ultimately this is useful for making the BMI grid functions cross-grid compatible... That is, if it is OK to have `get_grid_x`, etc for RasterModelGrids, then we can just raise NotImplemented errors after testing for Hex/Radial/Voronoi in the appropriate functions (e.g., get grid shape).

landlab/grid/create.py
```diff
@@ -264,106 +413,77 @@ def create_grid(file_like):
...
 
-    for grid_type in grid_dict:
-        if grid_type in _MODEL_GRIDS:
-            grid_class = _MODEL_GRIDS[grid_type]
-        else:
-            msg = "create_grid: provided grid type not supported."
-            raise ValueError
+    new_grids = []
+    for grid_type, grid_desc in as_list_of_tuples(grids):
+        grid_desc = norm_grid_description(grid_desc)
```
kbarnhart: Should there be a way that a user can indicate at this point which grids get which grid IDs?

Related: 
Also, since this supports having multiple, grids, where are the getting stored. 

`self._grids`??? with a critical plural.

I guess there also needs to be a way for the input file to indicate which grid goes with which component.

landlab/grid/raster.py
```diff
@@ -349,6 +358,53 @@ class RasterModelGrid(DiagonalsMixIn, ModelGrid, RasterModelGridPlotter):
     or set it up such that one can create a zero-node grid.
     """
 
+    @staticmethod
+    def _parse_parameters(args, kwds):
+        """Parse __init__ parameters into new-style parameters."""
```
kbarnhart: this is great 🎉 :

landlab/bmi/bmi_bridge.py
mcflugen: Right. Good point. I spent all that time refactoring `create_grid` but don't actually use it. 🤦‍♂ .

landlab/bmi/bmi_bridge.py
mcflugen: No, I don't think it's a problem that `get_grid_x` is defined for a uniform rectilinear grid. It's probably just redundant and should never get called. For other grid types we'll use it.

landlab/grid/create.py
mcflugen: The newly created grids are returned in a list in the order they are read from the *yaml* file. Grids don't have an ID apart from their index in the list, which depends on how they are listed in the *yaml* file.

Ultimately, I'm not sure how useful being able to describe multiple grids in a single file really is. If one wants to describe multiple components, each with their own grid, they probably should be broken up into separate files where each file has just one grid.

landlab/grid/raster.py
mcflugen: I thought you would like this part!

landlab/grid/create.py
kbarnhart: Does that imply that create_grid should only be able to create one grid?

mcflugen: That's what I'm wondering. For now, I think it's good to have something that can read multiple grids but perhaps we should think about if we really need that functionality or not.

kbarnhart: Perhaps a better way to think about is that `create_grid` is something that can create multiple grids from a single file, but that the `bmi_bridge` checks that only one grid is provided?

I think the general functionality of `create_grid` makes sense.

kbarnhart:
@mcflugen ping me again when I should look next.

I also feel strongly that you should spend as little time as possible making this work with Py27. Even if that means we have some tests that just verify that if someone is running 27 they get an error message for BMI-related functions.

mcflugen:
@kbarnhart This is ready for you to have another look at. Thank you!

landlab/bmi/bmi_bridge.py
```diff
@@ -310,42 +308,88 @@ def get_value_ref(self, name):
...
+
+        def get_grid_y(self, grid, y):
+            return self._base.grid.y_of_node
+
+        def get_grid_z(self, grid, z):
+            raise NotImplementedError("get_grid_z")
+
+        def get_value_at_indices(self, name, dest, inds):
+            raise NotImplementedError("get_value_at_indices")
+
```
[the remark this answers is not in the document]
mcflugen: @kbarnhart I suppose they should since it would be so easy to do so. However, thus far, we've seen this as an extra that doesn't need to be implemented.

To be complete, I think it really should be implemented. In the future we'll want this to be part of the standard BMI and not as an extra.

landlab/bmi/bmi_bridge.py
```diff
@@ -310,42 +308,88 @@ def get_value_ref(self, name):
...
+        def get_grid_x(self, grid, x):
+            return self._base.grid.x_of_node
+
+        def get_grid_y(self, grid, y):
+            return self._base.grid.y_of_node
+
+        def get_grid_z(self, grid, z):
+            raise NotImplementedError("get_grid_z")
+
```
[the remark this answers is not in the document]
mcflugen: Never say never.

landlab/bmi/bmi_bridge.py
```diff
@@ -310,42 +308,88 @@ def get_value_ref(self, name):
...
+            return self._base.grid.number_of_patches
+
+        def get_grid_face_nodes(self, grid, face_nodes):
+            return self._base.grid.nodes_at_patch
+
+        def get_grid_node_count(self, grid):
+            return self._base.grid.number_of_nodes
+
+        def get_grid_nodes_per_face(self, grid, nodes_per_face):
+            raise NotImplementedError("get_grid_nodes_per_face")
```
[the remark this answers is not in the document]
mcflugen: This could definitely be implemented, I would think. I'll try to implement a few more methods to provide a more complete BMI.

landlab/bmi/bmi_bridge.py
```diff
@@ -126,27 +126,27 @@ def wrap_as_bmi(cls):
...
+    ...     method: flexure
     ... clock:
     ...     start: 0.
     ...     stop: 10.
     ...     step: 2.
     ... grid:
-    ...     type: raster
-    ...     shape: [20, 40]
-    ...     spacing: [1000., 2000.]
+    ...     raster_model_grid:
```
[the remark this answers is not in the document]
mcflugen: This is mostly an aesthetic thing. I find `ICantReadThis` to be ugly but, of course, that's just my opinion.

* within the yaml file I think it reads better if everything is lower case with underscores separating words
* one less rule - everything's snake case. Not, everything's snake case except...
* when we use `raster_model_grid` in this context we're using it as a key to a mapping, which would normally be snake case, and can be thought of as referring to an instance

However... I see where you're coming from and is why I left in the option for both.

Another option for describing grids, which I think I like, could be,
```yaml
grids:
  - _type: RasterModelGrid
    args: [4, 5]
    xy_spacing: 2.0
  - _type: RasterModelGrid
    args: [40, 50]
    xy_spacing: 0.2
```
In this case *grids* is a list of mappings where the type of each mapping is given by the `_type` key.

If it's ok with you, I would like to merge this with the current format and then possibly home in on a better format in a separate pull request. This way I can get a bunch of components into *pymt* before the meeting.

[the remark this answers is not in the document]
kbarnhart: Update from LL call: We will just use `RasterModelGrid` (or other grid class names).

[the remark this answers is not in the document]
kbarnhart: @mcflugen I'm making this change right now.

kbarnhart:
@mcflugen I think this is really close. 

- [x] I made the changes we talked about during the LL call. 
- [ ] The `create_grid` docstring needs to be revised to reflect the new changes.
- [ ] We need to make an issue to keep track of needing to add remaining tests and BMI functions at a later time (I've made comments to keep track of these. 

I'd approve this now, but it won't count for merging unless no further changes are made.

landlab/bmi/bmi_bridge.py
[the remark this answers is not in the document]
mcflugen: Sounds good but I think we may still want to change this format in the future. Somehow there gets to be too many levels of indentation where the additional levels seem to sometimes be redundant.

I agree we should try to stick to having as few ways of specifying things as is reasonable. There are still a few things that we can improve in this regard.

mcflugen:
@kbarnhart Thank you!!! 🎉

mcflugen:
@kbarnhart I've updated the *create_grid* docstring to reflect the new format for the *yaml* file. Please let me know if you need anything else.

landlab/grid/create.py
```diff
@@ -198,17 +343,23 @@ def create_grid(file_like):
...
+        grid:
+          RasterModelGrid:
+            - [4, 5]
+            - xy_spacing: [3, 4]
+            - boundary_conditions:
+              - set_closed_boundaries_at_grid_edges:
+                - True
+                - True
+                - True
+                - True
```
kbarnhart: looks good!

Issue #925, opened 2019-05-13
Remove very old deprecated functions


kbarnhart:
There are a few pre 1.0 functions still kicking around. Forthcoming PR.

kbarnhart:
@mcflugen Looks like I already made an issue about this. 

Once #997 is in, we can work on removing/discussing the remaining functions.

PR #926, opened 2019-05-13
Deprecation (and remove old docs folder)

kbarnhart:
This PR does the following:
1. Starts to address #925 
2. Removes the `2014-docs` folder. I have zero idea why this folder is still in the repo... anyone know?

Still to do:
@mcflugen are those things deprecated with `version="1.0beta"` things that should be removed at the time of LLv1.0, or things that should stay until just before 2.0?

What about things deprecated with `version="1.0"`?


kbarnhart:
@mcflugen it looks like the the `version=1.0beta` deprecations were made in May 2016 (e.g. [calc_grad_at_link](https://github.com/landlab/landlab/commit/5808cc49ad7f1630b139c0b730270a851cee67fa)).
and the `version=1.0` ones (e.g. [nodes_at_corners_of_grid](https://github.com/landlab/landlab/commit/486109e6423b9fdf7d28e33948d34d5cba2e421c)) were also in May 2016.
 
As all of this was before the 1.0 release in Sept 2016, I'm going to remove them in this PR.

kbarnhart:
Alrighty... I've now removed the remaining deprecated function. We have some failed tests because there are still components and tests that use functions that have been long-deprecated (like in 2016). I've fixed some of the low hanging 🥝 🍓 but request help on the following. @saisiddu @jadams15 @mcflugen please commit to this branch...

@saisiddu : remove use of the following deprecated functions from radiation:
- [ ] `calculate_slope_aspect_at_nodes_burrough`
- [ ] `face_connecting_cell_pair`
- [ ] `calculate_diff_at_links`

@jadams15 : remove use of the following from overland flow
- [ ] `_active_links_at_node`

@mcflugen either you or I can deal with the following in grid:
- [ ] `calculate_diff_at_links`
- [ ] `face_connecting_cell_pair`
- [ ] `_active_links_at_node`

jadams15:
hi Katy, I'm pretty swamped for the next few weeks so I don't know when I can make time to address this - because this is going to be a complicated fix. Essentially, I don't know *how* I would replace this in OverlandFlow - I need only active link IDs before I have values on the grid - having those IDs before calculation is essential to set up vertical and horizontal active neighbor arrays. Those are required to do any calculations in the component. Advice appreciated.

kbarnhart:
@jadams15 all of the deprecated functions that have replacements (`_active_links_at_node`) had hints in the deprecation warnings indicating how to address with non-deprecated functions.

For this function it was 
```python
@deprecated(use="vals[links_at_node]*active_link_dirs_at_node", version=1.0)
```

jadams15:
I did see the deprecation warning. But because the deprecated function is used before there are values on the grid, it's not as simple as swapping them out (I think). It's a fundamental part of the set-up of the OF component.

saisiddu:
@kbarnhart I will not be able to get to this until the first week of June. I tried replacing the method 'calculate_slope_aspect_at_nodes_burrough' with the suggested function long time back, and it didn't work out (calculated slopes were not the same). I will give it another try, but as I said, I can get to it in first week of June.

kbarnhart:
@saisiddu @jadams15 If these functions cannot be replaced by the recommended replacements, do you think that is an argument that they should never have been deprecated?

kbarnhart:
Also, I don't think that there is a big hurry on this... But I'm trying to do a bit of cleaning house in advance of LL2.0.

saisiddu:
@kbarnhart: I would like to compare outputs of 'calculate_slope_aspect_at_nodes_burrough' and the current (latest) method to comment on whether this should not have been deprecated. It is possible that I might have messed it up. I would definitely want to get to this before the Ecohydrology paper is out.

jadams15:
I am not sure why the method in my case was deprecated. 

My hesitation about replacing it is that the suggested replacement does not accomplish what I need. The "replacement" method finds the VALUES stored on active links connected to a node. I needed to query the IDs of the active links connected to a node.

Because the leading underscore indicates it is not a user-facing method, and because the new method does not replicate what I am using it for, I would argue it should not be deprecated.

kbarnhart:
I've removed the deprecations that are irritating to deal with and. put them in another PR. 
@mcflugen if this passes, consider reviewing?

mcflugen:
@kbarnhart Yes, I will give it a review. I had a quick look and saw a lot of red, which makes me happy.

mcflugen:
👍 @kbarnhart I think this looks good.

Issue #928, opened 2019-05-16
Tons of deprecated warnings from new grid construction


kbarnhart:
#919 is great... but it didn't address all the new deprecation warnings it caused. Forthcoming PR fixes them.

PR #929, opened 2019-05-16
Fix grid warnings

kbarnhart:
Addresses #928


kbarnhart:
@mcflugen can you approve?

PR #930, opened 2019-05-16
Implement BMI methods

mcflugen:
This pull request makes some necessary changes to get our BMI wrapping working. The main thing to notice is that all of the BMI methods are now implemented. I've also had to make some minor tweaks to a few components - typos in the metadata, `update` method shouldn't have any required arguments.

I've also made a change to the `synthetic.constant` function. Because not all fields will be *float*, I've added a *dtype* keyword so that a user can set the data type of the created array. If *dtype* is not provided, it is determined from the type of the *value* keyword (i.e. 1.0 means *float*, 1 means *int*).

@kbarnhart Would you please have a look at this? It should be quick.

### DIFF
--- a/landlab/bmi/bmi_bridge.py
+++ b/landlab/bmi/bmi_bridge.py
@@ -20,6 +20,26 @@ from ..core import load_params
 from ..core.model_component import Component
 from ..framework.decorators import snake_case
 from ..grid.create import create_grid
+from ..grid import RasterModelGrid, HexModelGrid
+
+
+BMI_LOCATION = {
+    "node": "node",
+    "link": "edge",
+    "patch": "face",
+    "corner": "node",
+    "face": "edge",
+    "cell": "face",
+}
+
+BMI_GRID = {
+    "node": 0,
+    "link": 0,
+    "patch": 0,
+    "corner": 1,
+    "face": 1,
+    "cell": 1,
+}
 
 
 class TimeStepper(object):
@@ -213,7 +233,8 @@ def wrap_as_bmi(cls):
 
         def get_time_units(self):
             """Time units used by the component."""
-            raise NotImplementedError("get_time_units not implemented")
+            return "s"
+            # raise NotImplementedError("get_time_units not implemented")
             # Should be implemented.
 
         def initialize(self, config_file):
@@ -287,19 +308,25 @@ def wrap_as_bmi(cls):
 
         def get_var_grid(self, name):
             """Get the grid id for a variable."""
-            return 0
+            at = self._base._var_mapping[name]
+            return BMI_GRID[at]
 
         def get_var_itemsize(self, name):
             """Get the size of elements of a variable."""
-            return np.dtype("float").itemsize
+            at = self._base._var_mapping[name]
+            return self._base.grid[at][name].itemsize
 
         def get_var_nbytes(self, name):
             """Get the total number of bytes used by a variable."""
-            return self.get_itemsize(name) * self._base.grid.number_of_nodes
+            at = self._base._var_mapping[name]
+            return self._base.grid[at][name].nbytes
+            # return self.get_var_itemsize(name) * self._base.grid.number_of_nodes
 
         def get_var_type(self, name):
             """Get the data type for a variable."""
-            return str(np.dtype("float"))
+            at = self._base._var_mapping[name]
+            return str(self._base.grid[at][name].dtype)
+            # return str(np.dtype("float"))
 
         def get_var_units(self, name):
             """Get the unit used by a variable."""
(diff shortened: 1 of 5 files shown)


landlab/bmi/bmi_bridge.py
```diff
@@ -19,7 +19,27 @@
...
+
+
+BMI_LOCATION = {
+    "node": "node",
+    "link": "edge",
+    "patch": "face",
+    "corner": "node",
+    "face": "edge",
+    "cell": "face",
+}
```
kbarnhart: YES! glad to know we have such a mapping in the source code.

landlab/bmi/bmi_bridge.py
```diff
@@ -288,46 +308,59 @@ def finalize(self):
...
-            return np.dtype("float").itemsize
+            at = self._base._var_mapping[name]
+            return self._base.grid[at][name].itemsize
 
         def get_var_nbytes(self, name):
             """Get the total number of bytes used by a variable."""
-            return self.get_itemsize(name) * self._base.grid.number_of_nodes
+            at = self._base._var_mapping[name]
+            return self._base.grid[at][name].nbytes
+            # return self.get_var_itemsize(name) * self._base.grid.number_of_nodes
```
kbarnhart: here and elsewhere, do we need to get rid of the comment lines?

landlab/bmi/bmi_bridge.py
```diff
@@ -288,46 +308,59 @@ def finalize(self):
...
             """Set the values of a variable."""
             if name in self.get_input_var_names():
-                if name in self._base.grid.at_node:
-                    self._base.grid.at_node[name][:] = values.flat
-                else:
-                    self._base.grid.at_node[name] = values
+                at = self._base._var_mapping[name]
+                self._base.grid[at][name][:] = values.flat
+                # if name in self._base.grid.at_node:
+                #     self._base.grid.at_node[name][:] = values.flat
```
kbarnhart: For example, these comments

landlab/bmi/bmi_bridge.py
```diff
@@ -288,46 +308,59 @@ def finalize(self):
...
+                # else:
+                #     self._base.grid.at_node[name] = values
             else:
                 raise KeyError("{name} is not an input item".format(name=name))
 
         def get_grid_origin(self, grid, origin):
             """Get the origin for a structured grid."""
-            origin[:] = (self._base.grid.node_y[0], self._base.grid.node_x[0])
+            if grid == 0:
+                origin[:] = (self._base.grid.node_y[0], self._base.grid.node_x[0])
```
kbarnhart: I think I understand what you are doing... basically you have set this up so that both parts of the Landlab dual graph are accessible, with the node-link-patch as the grid with id 0 and the corner-face-cell having an id of 1... 

I think this **may** make sense... but it needs documentation. 

Should one of them be default?

landlab/values/synthetic.py
```diff
@@ -333,7 +333,7 @@ def _get_x_and_y(grid, at):
     return x, y
 
 
-def constant(grid, name, at="node", where=None, value=0.0):
+def constant(grid, name, at="node", where=None, value=0.0, dtype=None):
```
kbarnhart: I think this is great.

landlab/bmi/bmi_bridge.py
mcflugen: Correct. BMI doesn't have a concept of a dual graph so, in BMI speak, landlab components have 2 grids: one with *nodes*, *links*, and *patches* and the other with *corners*, *faces*, and *cells*.

I don't think there can be a default. BMI requires these methods to accept a grid id as input.

kbarnhart: Sounds good... However, as most landlab grids are node-focused, it is almost certainly the case that users will want to use grid=0...

As long as there is documentation (I think in the class docstring is sufficient, rather than in every function), it is OK.

mcflugen: @kbarnhart I've added documentation to *wrap_as_bmi* that outlines some of the differences between landlab components and BMI components.

landlab/bmi/bmi_bridge.py
mcflugen: Removed!

mcflugen:
@kbarnhart I've updated the documentation and removed the unnecessary comments. What do you think?

mcflugen:
@kbarnhart Thank you!

landlab/bmi/bmi_bridge.py
```diff
@@ -108,6 +128,46 @@ def advance(self):
...
+        *  *cell*: *face*
+         
+    3.  In BMI, the *initialize* method requires an input file that is
+        used to create and setup the model for time-stepping. landlab
+        components generally do not have anything like this; instead
+        this task is usually done programmatically. Thus, the 
+        input file that is used by the BMI *initialize* method is
+        a standard landlab input file as used by the landlab *create_grid*
+        function.
+
```
mcflugen: Love it! 🎉

landlab/values/synthetic.py
mcflugen: Cool. Note that `constant(grid, "x", value=10)` will create a field of integers - not floats. Since this follows the same behavior as numpy functions like `arange` I figured it was fine.

mcflugen:
@kbarnhart I removed some more trailing whitespace and, in doing so, dismissed your review. Would you please approve again? Thank you.

PR #931, opened 2019-05-19
Add boundary conditions to BMI

mcflugen:
This pull request adds boundary conditions (*status_at_node*) to the BMI wrapping of landlab components. When a landlab component is wrapped to expose a BMI, a new input/output var is added to the BMI component called "boundary_condition_flag", which is an interface to the landlab component's *status_at_node* field. Since BMI doesn't have any special way of representing boundary conditions, it must be done through BMI variables.

@kbarnhart , I made one other small change. In wrapping the *FlowAccumulator*, I noticed that the *flow__nodes_not_in_stack* field was never used. Because the field was never created, this caused problems for the wrapping. Anyway, I removed it from the class metadata (*_output_var_names* etc.). What do you think?

### DIFF
--- a/landlab/bmi/bmi_bridge.py
+++ b/landlab/bmi/bmi_bridge.py
@@ -12,6 +12,8 @@ The `wrap_as_bmi` function wraps a landlab component class so that it
 exposes a Basic Modelling Interface.
 
 """
+import inspect
+
 import numpy as np
 
 from bmipy import Bmi
@@ -243,6 +245,16 @@ def wrap_as_bmi(cls):
             self._clock = None
             super(BmiWrapper, self).__init__()
 
+            self._cls._input_var_names = tuple(
+                set(self._cls._input_var_names) | {"boundary_condition_flag"}
+            )
+            self._cls._output_var_names = tuple(
+                set(self._cls._output_var_names) | {"boundary_condition_flag"}
+            )
+            self._cls._var_mapping["boundary_condition_flag"] = "node"
+            self._cls._var_units["boundary_condition_flag"] = ""
+            self._cls._var_doc["boundary_condition_flag"] = "boundary condition flag of grid nodes"
+
         def get_component_name(self):
             """Name of the component."""
             return self._cls.name
@@ -319,11 +331,20 @@ def wrap_as_bmi(cls):
             self._clock = TimeStepper(**clock_params)
 
             self._base = self._cls(grid, **params.pop(snake_case(cls.__name__), {}))
+            self._base.grid.at_node["boundary_condition_flag"] = self._base.grid.status_at_node
 
         def update(self):
             """Update the component one time step."""
             if hasattr(self._base, "update"):
                 self._base.update()
+            elif hasattr(self._base, "run_one_step"):
+                nargs = len(inspect.signature(self._base.run_one_step).parameters)
+
+                if nargs == 0:
+                    self._base.run_one_step()
+                else:
+                    self._base.run_one_step(self._clock.step)
+
             self._clock.advance()
 
         def update_frac(self, frac):
@@ -382,8 +403,11 @@ def wrap_as_bmi(cls):
         def set_value(self, name, values):
             """Set the values of a variable."""
             if name in self.get_input_var_names():
-                at = self._base._var_mapping[name]
-                self._base.grid[at][name][:] = values.flat
+                if name == "boundary_condition_flag":
+                    self._base.grid.status_at_node = values
+                else:
+                    at = self._base._var_mapping[name]
+                    self._base.grid[at][name][:] = values.flat
             else:
                 raise KeyError("{name} is not an input item".format(name=name))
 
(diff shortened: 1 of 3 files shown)


mcflugen:
In running these tests, I've discovered that landlab doesn't work with the latest version of *scipy* (1.3), which must have just recently been released. The problem comes from the *statsmodels* package, which is used by the *LandslideProbability* component. It appears that *statsmodels* uses a previously deprecated function that is now no longer available in `scipy.misc` (it was moved to `scipy.special`).

Anyway, until *statsmodels* is fixed, we are stuck with *scipy* less than version 1.3.

@saisiddu, @RondaStrauch We only use *statsmodels* for *ECDF* in *LandslideProbability*. Do you know of another implementation of this we can use?

setup.py
```diff
@@ -81,7 +81,7 @@ def run(self):
         "netcdf4",
         "pyyaml",
         "pyshp",
-        "scipy",
+        "scipy<1.3",
         "six",
```
kbarnhart: @mcflugen : Can you turn your comment to @saisiddu and @RondaStrauch into an issue so we can keep track of this beyond closing this PR?

landlab/components/flow_accum/flow_accumulator.py
```diff
@@ -633,7 +633,7 @@ class DepressionFinderAndRouter to the parameter `depression_finder`.
         "drainage_area",
         "surface_water__discharge",
         "flow__upstream_node_order",
-        "flow__nodes_not_in_stack",
+        # "flow__nodes_not_in_stack",
         "flow__data_structure_delta",
```
kbarnhart: I think removing this is great. I'd get rid of the commented lines.

landlab/bmi/bmi_bridge.py
```diff
@@ -319,11 +352,20 @@ def initialize(self, config_file):
...
             if hasattr(self._base, "update"):
                 self._base.update()
+            elif hasattr(self._base, "run_one_step"):
+                nargs = len(inspect.signature(self._base.run_one_step).parameters)
+
+                if nargs == 0:
+                    self._base.run_one_step()
+                else:
+                    self._base.run_one_step(self._clock.step)
+
```
kbarnhart: Does it make sense to make an issue about this (and tag LL2.0) requesting that all R1S methods take dt as the first (and only) argument...?

landlab/bmi/bmi_bridge.py
```diff
@@ -243,17 +253,40 @@ def __init__(self):
...
+            #     set(self._cls._output_var_names) | {"boundary_condition_flag"}
+            # )
+            self._var_mapping = self._cls._var_mapping.copy()
+            self._var_units = self._cls._var_units.copy()
+            self._var_doc = self._cls._var_doc.copy()
+
+            self._var_mapping["boundary_condition_flag"] = "node"
+            self._var_units["boundary_condition_flag"] = ""
+            self._var_doc["boundary_condition_flag"] = "boundary condition flag of grid nodes"
+            # self._cls._var_mapping["boundary_condition_flag"] = "node"
```
kbarnhart: - Recommend remove comments
- Are our boundary condition flags at all universal? (my assumption is no...) is there a more universal format that makes sense to use? I imagine this won't change between now and (tomorrow/the asap time when this is merged), but is it worth making an issue to keep track of doing this somewhat correctly?

landlab/bmi/bmi_bridge.py
```diff
@@ -243,17 +253,40 @@ def __init__(self):
...
+            )
+            # self._cls._output_var_names = tuple(
+            #     set(self._cls._output_var_names) | {"boundary_condition_flag"}
+            # )
+            self._var_mapping = self._cls._var_mapping.copy()
+            self._var_units = self._cls._var_units.copy()
+            self._var_doc = self._cls._var_doc.copy()
+
+            self._var_mapping["boundary_condition_flag"] = "node"
+            self._var_units["boundary_condition_flag"] = ""
```
kbarnhart: is "-" more appropriate? I think we've used that in the past. Or is "" the correct _no units_ value for UDUNITS?

kbarnhart:
@mcflugen other than the comments made I think this looks good. Rather than asking you to really address some of the issues in this, I outlined Issues that I think we should make to keep track of them. That way this can get into `master` on the timeline you need and we can keep track of problematic things that this PR exposes. Sounds good?

I'll watch to approve when changes are made so that this can keep moving... Though I won't be super on top of this later because of 🎲 🐉 👑 !!!

setup.py
mcflugen: Definitely. I intended to do just that.

landlab/bmi/bmi_bridge.py
mcflugen: "-" is an invalid unit for *cfunits*, so I've been trying to use the empty string to indicate "no units".

mcflugen:
@kbarnhart Thank you! I've removed the commented out code.

landlab/bmi/bmi_bridge.py
mcflugen: I don't think our boundary condition flags are at all universal but I'm not sure that anyone's are.

landlab/bmi/bmi_bridge.py
mcflugen: Done.

setup.py
mcflugen: Done.

PR #943, opened 2019-06-04
add return_array_at_link

gregtucker:
Add a return_array_at_link function, as a complement to the existing return_array_at_node. Also put both in __init__ so they're visible to landlab.utils, and catch an exception in one of the decorators.


kbarnhart:
@gregtucker just checking that you don't have anything local before I merge this?

Issue #909, opened 2019-04-13
Subsetting a DataRecord doesn't return a true DataRecord


kbarnhart:
Selecting a subset of a `DataRecord` (as shown by the MWE below) results in the
return of a `DataRecord` instance --- but one that does not have important attributes of DataRecord (for example the attribute `self._grid`). This means that it is not possible to run `self.calc_aggregate_value` which requires the grid to determine the size of the aggregate value array to create. 

This issue arose as @pfeiffea had asked me how to do a cleaner job indexing some
of the parcels stored in the in-progress NetworkSedimentTransporter (related to Issue #901 ).
I realized that part of the issue was that she really only wanted to deal with 
the current time step's parcels, so I tried to select just those and use them.

I then found this bug...

It would be great if indexing and selecting portions of the data record supported by Xarray (e.g. [these shown here](http://xarray.pydata.org/en/stable/indexing.html#indexing-and-selecting-data)) were supported by DataRecord. 

Tagging @nathanlyons and @margauxmouchene

```python

import numpy as np
from landlab.data_record import DataRecord
from landlab import RasterModelGrid
grid = RasterModelGrid((3,3))


element_id = np.array(
    [0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 1], dtype=int
)
element_id = np.expand_dims(element_id, axis=1)
time = np.array([0.]) 

np.random.seed(0)

age = np.random.rand(np.size(element_id), np.size(time))  

items = {"grid_element": "node", "element_id": element_id}

variables = {"age": (["item_id", "time"], age)}

parcels = DataRecord(grid, items=items, time=time, data_vars=variables)

sel_time_parcels = parcels.sel(time=0)

sel_time_parcels
<xarray.DataRecord>
Dimensions:       (item_id: 15)
Coordinates:
    time          float64 0.0
  * item_id       (item_id) int64 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14
Data variables:
    grid_element  (item_id) object 'node' 'node' 'node' ... 'node' 'node' 'node'
    element_id    (item_id) int64 0 0 1 1 2 2 3 3 4 4 5 5 6 6 1
    age           (item_id) float64 0.5488 0.7152 0.6028 ... 0.9256 0.07104

s = sel_time_parcels.calc_aggregate_value(np.sum, data_variable='ages')

Traceback (most recent call last):

[...]

PR #948, opened 2019-06-18
Barnhark/fix909 datarecord

kbarnhart:
@nathanlyons @margauxmouchene

here is a proposed fix that approaches fixing #909. it removes DataRecord inheriting from xarray.Dataset and puts the dataset at self.dataset. 

It gets rid of inplace=True (and thus all those future warnings).

It also creates a feature needed by @pfeiffea in which items can live on a "phantom link". Later today/tomorrow I'll create a function "select_where" that wraps xarray select and where such and returns a DataRecord instance with a self.dataset that has been appropriately subsetted.

### DIFF
--- a/landlab/data_record/data_record.py
+++ b/landlab/data_record/data_record.py
@@ -47,6 +47,7 @@ class DataRecord(object):
     def __init__(
         self,
         grid,
+        dummy_elements=None,
         time=None,
         items=None,
         data_vars=None,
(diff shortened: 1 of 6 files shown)


kbarnhart:
@margauxmouchene and @nathanlyons I'm going to close this PR to replace it with two different PRs. One which removes the inheritance (and currently passes), and one that adds the dummy grid location functionality that @pfeiffea and I need for the under-construction network sediment transporter.

PR #956, opened 2019-07-07
First steps to make Components LL2.0 compatible 

kbarnhart:
This PR is in progress. So far the goals of the PR are to

- [x] Removes use filename or kwds
- [x] Removes tests and example that use the model parameter dictionary
- [x] Remove all deprecated elements
- [x] Removes all use of **kwds
- [x] Clean up creating of fields within components
- [x] Re-organize the docstring header to make a single dictionary, appropriate for use as a netcdf variable attribute. 
- [x] Modify run-one-step methods so that they take EITHER no arguments (e.g. Flow Accumulation, which doesn't need dt) or take ONLY ONE argument, dt.
- [x] Add default values where appropriate
~- [ ] Clean up Parameter blocks for auto-parsing.~ (@mcflugen and I have just discussed this and decided that it would be nice but is not necessary for a 2.0 release).
- [x] No more undocumented public attributes (Part 1 remove all)
- [x] No more undocumented public attributes (Part 2 docstrings for new properties)
- [x] Ensure that all components make the fields they say the do
- [x] Ensure components check for fields they say they need
- [x] Make sure all components include all component metadata (e.g., no missing var_doc or var_units)
- [x] Make sure all components make and get fields with the correct dtype.


landlab/components/hack_calculator/hack_calculator.py
```diff
@@ -159,40 +159,55 @@ class HackCalculator(Component):
...
-        "flow__upstream_node_order": "node",
-    }
-    _var_doc = {
-        "topographic__elevation": "Land surface topographic elevation",
-        "flow__receiver_node": "Node array of receivers (node that receives flow from current node)",
-        "drainage_area": "Upstream accumulated surface area contributing to the node's discharge",
-        "flow__link_to_receiver_node": "Node array containing ID of link that leads from each node to its receiver, or BAD_INDEX_VALUE if no link",
-        "distance_to_divide": "Distance from drainage divide.",
-        "flow__upstream_node_order": "node order such that nodes must appear in the list after all nodes downstream of them",
+    __info = {
```
mcflugen: `_info`, not `__info`

landlab/components/landslides/landslide_probability.py
```diff
@@ -184,117 +183,127 @@ class property.
...
-        ),
-    }
-
-    # Run Component
-    @use_file_name_or_kwds
     def __init__(
         self,
         grid,
         number_of_iterations=250,
+        g=None,
```
mcflugen: `g=scipy.constants.g`

landlab/components/lithology/lithology.py
```diff
@@ -260,9 +279,60 @@ def __init__(self, grid, thicknesses, ids, attrs, layer_type="MaterialLayers"):
...
+
+        Returns
+        -------
+        current rate of vertical advection
+        """
+        return return_array_at_node(self._grid, self._dz_advection)
+
+    @dz_advection.setter
+    def dz_advection(self, dz_advection):
+        _ = return_array_at_node(
```
mcflugen: `_ = ...`? Does this not cause the linter complain about an unused variable?

landlab/components/overland_flow/generate_overland_flow_Bates.py
```diff
@@ -101,61 +99,78 @@ def __init__(
         mannings_n=0.03,
         g=9.81,
```
mcflugen: `g=scipy.constants.g` just for consistency?

landlab/components/plant_competition_ca/plant_competition_ca.py
```diff
@@ -67,38 +65,43 @@ class VegCA(Component):
...
-        "plant__live_index": "None",
-        "plant__age": "Years",
-    }
-
-    _var_mapping = {
-        "vegetation__cumulative_water_stress": "cell",
-        "vegetation__plant_functional_type": "cell",
-        "plant__live_index": "cell",
-        "plant__age": "cell",
+    _name = "VegCA"
```
mcflugen: Is there a reason to change the name of the component to the name of the class? I kind of thought of the `_name` attribute as being a more descriptive name.

landlab/components/stream_power/sed_flux_dep_incision.py
```diff
@@ -646,46 +669,33 @@ def sed_flux_fn_gen(rel_sed_flux_in):
...
+        grid = self._grid
         node_z = grid.at_node["topographic__elevation"]
         node_A = grid.at_node["drainage_area"]
         flow_receiver = grid.at_node["flow__receiver_node"]
         s_in = grid.at_node["flow__upstream_node_order"]
         node_S = grid.at_node["topographic__steepest_slope"]
 
-        if type(flooded_depths) is str:
-            flooded_depths = grid.at_node[flooded_depths]
+        if type(self._flooded_depths) is str:
```
mcflugen: These should all be `if isinstance(...)`. I notice now that the `if type(...)` pattern is fairly widely used throughout landlab. How about we fix this in another pull request?

landlab/components/uniform_precip/generate_uniform_precip.py
```diff
@@ -158,26 +155,43 @@ def __init__(
...
+            self._grid.add_field("grid", "rainfall__flux", 0.0)
             self._gridupdate = True
         else:
             self._gridupdate = False
 
         self._intensity = self.get_storm_intensity()
 
+    @property
+    def storm_duration(self):
+        """ TODO
```
mcflugen: Do these *TODO*s need to be taken care of before merging this pull request?

landlab/components/vegetation_dynamics/vegetation_dynamics.py
```diff
@@ -275,21 +312,47 @@ def __init__(
...
+    @property
+    def Tb(self):
+        """Storm duration (hours)."""
+        return self._Tb
 
-        self._Blive_ini = self._Blive_init * np.ones(self.grid.number_of_cells)
-        self._Bdead_ini = self._Bdead_init * np.ones(self.grid.number_of_cells)
+    @Tb.setter
+    def Tb(self, Tb):
+        assert Tb >= 0
```
mcflugen:
```python
if Tb < 0.0:
    raise ValueError("Tb must be non-negative")
```

landlab/components/hack_calculator/hack_calculator.py
kbarnhart: Thanks for noticing that. Fixed.

landlab/components/landslides/landslide_probability.py
kbarnhart: I've fixed this here and elsewhere.

landlab/components/overland_flow/generate_overland_flow_Bates.py
kbarnhart: Fixed (here and elsewhere). For some tests this now means specifying that g=9.81

landlab/components/plant_competition_ca/plant_competition_ca.py
kbarnhart: You are right... I will change the places where I changed this back to their original state.

landlab/components/stream_power/sed_flux_dep_incision.py
kbarnhart: Fixing all the `type()` to `isinstance()` was a good bus task. 🚌 ! done (well at least within components)

landlab/components/vegetation_dynamics/vegetation_dynamics.py
kbarnhart: changed.

landlab/components/uniform_precip/generate_uniform_precip.py
kbarnhart: Yes! Now taken care of.

landlab/components/lithology/lithology.py
kbarnhart: Nope. The linter knows that `_` is fake. But I removed it anyway.

landlab/core/model_component.py
```diff
@@ -316,30 +376,43 @@ def var_loc(cls, name):
         str
             The location ('node', 'link', etc.) where a variable is defined.
         """
-        return cls._var_mapping[name]
+        return cls._info[name]["mapping"]
 
-    def initialize_output_fields(self):
+    def initialize_output_fields(self, dim=None):
```
kbarnhart: @mcflugen can you comment on whether you can think of a better name for `dim`. 

I made this to do some additional clean up to the route-to-multiple flow directions which use fields that are (n-grid-elements, max-number-of-receivers) in shape. 

I thought about `axis`, `second_dimension`... and don't like these, or any of the other ideas I came up with.

mcflugen: @kbarnhart When I've done things like this in the past I've used something like *values_per_element*. For example, when data are stored in fields in an *xarray* *Dataset*, the dimensions of the array you describe would be something like *(node, receiver_nodes_per_node)*. What do you think?

kbarnhart: I like it. Have made the change.

kbarnhart:
@landlab/core This is now ready for your review.

tests/components/test_components.py
```diff
@@ -0,0 +1,98 @@
...
+
+_VALID_LOCS = ("grid", "node", "link", "patch", "corner", "face", "cell")
+
+
+_REQ_ATTRS = ["doc", "mapping", "dtype", "intent", "optional", "units"]
+
+
+@pytest.mark.parametrize("Comp", COMPONENTS)
+def test_component_metadata(Comp):
+    if Comp.name not in (
```
gregtucker: This is beautiful!

landlab/components/depression_finder/lake_mapper.py
```diff
@@ -313,24 +291,61 @@ def updated_boundary_conditions(self):
...
+
+    @property
+    def depression_depth(self):
+        """At node array of depression depths."""
+        return self._depression_depth
+
+    @property
+    def depression_outlet_map(self):
+        """At node array indicating the node-id of the depression outlet."""
+        return self._depression_outlet_map
```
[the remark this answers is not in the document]
kbarnhart: I'll provide a small bit of background to this and similar changes. 

Previously this component had a public, undocumented attribute `self._depression_outlet_map` and it was featured in a docstring.

Thus in treating this (and similar) instances, I had two choices:

1) Make the attribute private and change the docstring OR
2) Make the attribute private, interpret the use in a docstring as intention by the component developer that such an attribute should exist as a public attribute and thus create a @property for it. 

I went with 2 because it seemed to honor the original developer's intention AND not change the user-facing example. 

I'm not certain how pythonic it is to have two ways to access the same array (through the grid or through the component), but I can say that many components provide access to key fields (and highlight this access in the docstrings) in this way.

landlab/components/depth_dependent_taylor_soil_creep/hillslope_depth_dependent_taylor_flux.py
```diff
@@ -126,7 +126,8 @@ class DepthDependentTaylorDiffuser(Component):
     (if_unstable = 'warn') and a boolean flag that turns on dynamic timestepping
     (dynamic_dt = False).
 
-    >>> DDdiff.soilflux(2., if_unstable='warn')
+    >>> DDdiff = DepthDependentTaylorDiffuser(mg,  if_unstable='warn')
```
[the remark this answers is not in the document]
kbarnhart: fixed. thanks for noticing.

landlab/components/detachment_ltd_erosion/generate_detachment_ltd_erosion.py
```diff
@@ -161,26 +160,24 @@ def __init__(
...
 
-        self.I = self._grid.zeros(at="node")  # noqa: E741
-        self.uplift_rate = uplift_rate
-        self.entrainment_threshold = entrainment_threshold
+        self._K = K_sp
+        self._m = m_sp
+        self._n = n_sp
 
-        self.dzdt = self._grid.zeros(at="node")
+        self._I = self._grid.zeros(at="node")  # noqa: E741
```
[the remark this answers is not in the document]
kbarnhart: This component previously used a public attribute called `self.I`. I changed it to `self._I`.

landlab/components/detachment_ltd_erosion/generate_erosion_by_depth_slope.py
```diff
@@ -113,44 +117,45 @@ class DepthSlopeProductErosion(Component):
...
     }
 
     def __init__(
         self,
         grid,
-        k_e,
+        k_e=0.001,
         fluid_density=1000.0,
-        g=9.81,
+        g=scipy.constants.g,
```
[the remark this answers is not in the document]
kbarnhart: I'm going to let @mcflugen respond to this one as I made the `scipy.constants.g` changes in response to his recommendation.

landlab/components/flow_director/flow_director_mfd.py
```diff
@@ -327,7 +380,7 @@ def __init__(self, grid, surface="topographic__elevation", **kwargs):
             A grid.
         surface : field name at node or array of length node, optional
             The surface to direct flow across, default is field at node:
-            topographic__self.surface_valuesation.
+            topographic__self._surface_valuesation.
```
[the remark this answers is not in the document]
kbarnhart: yup. fixed.

landlab/components/lithology/lithology.py
```diff
@@ -87,6 +99,12 @@ def __init__(self, grid, thicknesses, ids, attrs, layer_type="MaterialLayers"):
             used, then erosion removes material and creates layers of thickness
             zero. Thus, EventLayers may be appropriate if the user is interested
             in chronostratigraphy.
+        dz_advection : float, `(n_nodes, ) shape array, or at-node field array optional
```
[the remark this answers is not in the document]
kbarnhart: Its a length. I've added text to state the units explicitly.

tests/components/test_components.py
kbarnhart: Glad you like it!

landlab/components/detachment_ltd_erosion/generate_erosion_by_depth_slope.py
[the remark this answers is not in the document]
mcflugen: Previously, components that used *g* defined it as different values. Some used `scipy.constants.g`, some used 9.81, some used 10 (ok, not really). Just to be consistent, it seems like using the more accurate `scipy.constants.g` (defined as 9.80665) is the way to go. Having it here alerts the user that they are able to change it if they like, although, it is up to the component if it can be spatially variable. Not having it as a keyword argument would really be hiding it. I'm not sure if any components do that any more or not.

A user looking at the documentation for the component will see `g=9.80665` rather than `g=scipy.constants.g`, so they'll know exactly what the default value being used is.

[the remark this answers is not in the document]
gregtucker: That makes sense, given that the actual numerical value will be readily apparent to users.
(by the way, I meant "decimal" not "dimension" ;)

landlab/components/detachment_ltd_erosion/generate_detachment_ltd_erosion.py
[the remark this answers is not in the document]
gregtucker: Sorry, I wasn't clear, but I looked up what "noqa" means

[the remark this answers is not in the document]
mcflugen: @gregtucker the `# noqa: E741` was to disable the linter from warning about using a variable named *I* (because it can easily be mistaken with a lowercase *l* in many text editors). Because of the underscore, we may no longer strictly need the pragma but, even still, ideally we should come up with a better name. It's just asking for a typo in the future!

landlab/components/depression_finder/lake_mapper.py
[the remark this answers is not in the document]
gregtucker: Ok, that makes sense. I understand that the more general pattern this implies is "leave it to the component author to decide whether or not to offer public attribute variables that are also fields," which seems reasonable to me.

landlab/components/detachment_ltd_erosion/generate_detachment_ltd_erosion.py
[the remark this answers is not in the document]
mcflugen: @gregtucker @kbarnhart Oh, sorry, I see what your question was now. How is this
```python
self.dzdt = self._grid.zeros(at="node")
```
equivalent to this
```python
self._I = self._grid.zeros(at="node")
```
Why not, instead,
```python
self._dzdt = self._grid.zeros(at="node")
```
Was that what your *?* was about?

landlab/components/depression_finder/lake_mapper.py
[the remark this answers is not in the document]
mcflugen: @gregtucker @kbarnhart I agree, I think it's fine to leave it up to the component authors. However, I'm not sure we should encourage it. I like the idea of having just a single way to get values from the grid. I could definitely be persuaded though.

🤔 Actually, maybe something along these lines would be a good thing. Not necessarily having fields be properties of components, but properties of grids (e.g. `grid.at_node.depression_outlet_map`. This might reduce the chance of some pesky, hard-to-find errors where one component changes a reference to a field but doesn't tell other components.

landlab/components/detachment_ltd_erosion/generate_detachment_ltd_erosion.py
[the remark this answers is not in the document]
kbarnhart: Screen Shot 2019-09-27 at 11 25 35 AM

@mcflugen see screenshot... I think your question has more to do with GitHub's like comparison than any change I made.

landlab/components/depression_finder/lake_mapper.py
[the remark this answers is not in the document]
kbarnhart: There are a couple of reason's I would advocate making no changes but advocating to component authors that they access the grid fields through the grid (a recommendation in our Rules and Recommendations)

My reasons:

1) These component attributes have been user facing and we highlight them in our docstrings. Thus making changes may create more problems for users than we'd like. 

2) I think @mcflugen 's idea is nice, with but there is a major barrier to it being functional. The component attribute names and the field names are not always the same. For example, `DepressionFinder.depression_outlet_map` points to `grid.at_node["depression__outlet_node"]`.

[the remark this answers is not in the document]
mcflugen: @kbarnhart Sorry, I was changing the topic. I meant something slightly different. All I meant was using
```python
>>> grid.at_node.depression__outlet_node
```
instead of
```python
>>> grid.at_node["depression__outlet_node"]
```
*depression__outlet_node* would be a property of *grid.at_node*, which would mean a user couldn't easily change it's reference. Components themselves could still create properties with whatever names they like.

landlab/components/detachment_ltd_erosion/generate_detachment_ltd_erosion.py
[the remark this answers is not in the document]
gregtucker: Sorry for confusion, I just hadn't known about the # noqa directive

kbarnhart:
Thanks to @mcflugen and @gregtucker for taking a look at this. 

@landlab/core please comment by end of Thursday so I can merge this by the end of the week.

kbarnhart:
@mcflugen I'm going to merge this. OK?

PR #965, opened 2019-07-11
Deploy wheels to PyPI

mcflugen:
This pull request fixes the deployment from *Travis* to *PyPI* of tagged releases. The *Travis* deployment section wasn't working for us so I changed it to do it manually through *twine*. In addition, we now deploy binary distributions for both *linux* and *macos*. Previously, before it broke anyway, we only provided wheels for *macos*, not *linux*. Building wheels for was *linux* more complicated than I anticipated but it seems to be working now (you can see the releases for *v1.9.0b5*). We now build *manylinux* distributions following the example provided by the [Python Packaging Authority](https://github.com/pypa/python-manylinux-demo).


kbarnhart:
@mcflugen this looks like it was a total pain. Thanks for making it work!

PR #966, opened 2019-07-11
make fracture grid a component (FractureGridGenerator)

gregtucker:
Convert the old fracture_grid utility to a proper component.

### DIFF
--- a/landlab/components/fracture_grid/fracture_grid.py
+++ b/landlab/components/fracture_grid/fracture_grid.py
@@ -158,7 +158,11 @@ class FractureGridGenerator(Component):
     --------
     >>> from landlab import RasterModelGrid
     >>> grid = RasterModelGrid((5, 5))
-
+    >>> fg = FractureGridGenerator(grid=grid, frac_spacing=3)
+    >>> fg.run_one_step()
+    >>> grid.at_node['fracture_at_node']
+    array([1, 0, 0, 0, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+           0, 0], dtype=int8)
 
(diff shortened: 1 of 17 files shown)


landlab/components/fracture_grid/fracture_grid.py
```diff
@@ -158,55 +138,105 @@ def trace_fracture_through_grid(m, start_yx, spacing):
...
-    m : Numpy array
-        Array containing fracture grid, represented as 0's (matrix) and 1's
-        (fractures). If model_grid parameter is given, returns a 1D array
-        corresponding to a node-based array in the model grid. Otherwise,
-        returns a 2D array with dimensions given by numrows, numcols.
+    Examples
+    --------
+    >>> from landlab import RasterModelGrid
+    >>> grid = RasterModelGrid((5, 5))
+
```
kbarnhart: Should there be an actual example here? Even a really simple one?

landlab/components/fracture_grid/fracture_grid.py
```diff
@@ -158,55 +138,105 @@ def trace_fracture_through_grid(m, start_yx, spacing):
...
+
+    Notes
+    -----
+    Potential improvements:
+    - Add doctests
```
kbarnhart: I think once the example exists this improvement can be removed.

gregtucker:
Oops, forgot to finish doctest. Now done. The .rst files already exist; changed name to component name in one of them. @kbarnhart can you take another look?

kbarnhart:
@gregtucker looks great! When you squash-merge, try using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/). I'm still learning how to use them, but I think this would be classified as a `feat`.

PR #980, opened 2019-07-22
feat: add initial version of GroundwaterDupuitPercolator

gregtucker:
Add a component that models shallow groundwater flow

### DIFF
--- a/landlab/components/groundwater/dupuit_percolator.py
+++ b/landlab/components/groundwater/dupuit_percolator.py
@@ -21,17 +21,6 @@ class GroundwaterDupuitPercolator(Component):
     The GroundwaterDupuitPercolator uses the Dupuit approximation that the
     hydraulic gradient is equal to the slope of the water table.
 
-    Parameters
-    ----------
-    grid: ModelGrid
-            Landlab ModelGrid object
-    hydraulic_conductivity: float, field name, or array of float
-            saturated hydraulic conductivity, m/s
-            Default = 0.01 m/s
-    recharge_rate: float, field name, or array of float
-            Rate of recharge, m/s
-            Default = 1.0e-8 m/s
-
     Examples
     --------
     >>> from landlab import RasterModelGrid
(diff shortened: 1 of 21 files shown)


landlab/components/groundwater/dupuit_percolator.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+    Simulate groundwater flow in a shallow unconfined aquifer.
+
+    The GroundwaterDupuitPercolator uses the Dupuit approximation that the
+    hydraulic gradient is equal to the slope of the water table.
+
+    Parameters
+    ----------
+    grid: ModelGrid
+            Landlab ModelGrid object
```
kbarnhart: I'd put the parameters only in one place. Just after the init would be my preference (e.g., not here).

landlab/components/groundwater/dupuit_percolator.py
```diff
@@ -0,0 +1,197 @@
...
+            Rate of recharge, m/s
+            Default = 1.0e-8 m/s
+
+    Examples
+    --------
+    >>> from landlab import RasterModelGrid
+    >>> mg = RasterModelGrid((3, 3))
+    >>> z = mg.add_zeros('node', 'topographic__elevation')
+    >>> gdp = GroundwaterDupuitPercolator(mg)
+
```
kbarnhart: I might expand this example to show how this component is used a bit more.

landlab/components/groundwater/dupuit_percolator.py
```diff
@@ -0,0 +1,197 @@
...
+    using the average depth at the nodes.
+
+    Note that the current version does NOT handle surface seepage.
+    """
+
+    _name = "GroundwaterDupuitPercolator"
+
+    _input_var_names = set(("topographic__elevation",
+                            "aquifer_base__elevation"))
+
```
kbarnhart: I think that our current conception of an input is that it must already exist before the model is initialized. Thus, I think your handling of these two inputs in the init doesn't make sense. 

They should exist (perhaps it must be asserted that the aquifer elevation is always below or equal to the topography), and if they don't a Field error gets raised.

landlab/components/groundwater/dupuit_percolator.py
```diff
@@ -0,0 +1,197 @@
...
+    def __init__(self, grid, hydraulic_conductivity=0.01,
+                 recharge_rate=1.0e-8):
+        """Initialize the GroundwaterDupuitPercolator.
+
+        Parameters
+        ----------
+        grid: ModelGrid
+            Landlab ModelGrid object
+        hydraulic_conductivity: float, field name, or array of float
+                saturated hydraulic conductivity, m/s
```
kbarnhart: Not clear from this that hydraulic conductivity and recharge are on link and not on node.

landlab/components/groundwater/dupuit_percolator.py
```diff
@@ -0,0 +1,197 @@
...
+        recharge_rate: float, field name, or array of float
+                Rate of recharge, m/s
+                Default = 1.0e-8 m/s
+        """
+        # Store grid
+        self._grid = grid
+
+        # Shorthand
+        self.cores = grid.core_nodes
+        self.inactive_links = np.where(grid.status_at_link != ACTIVE_LINK)[0]
```
kbarnhart: Make life easier for future you and make all these private attributes?

landlab/components/groundwater/tests/test_dupuit_percolator.py
```diff
@@ -0,0 +1,48 @@
...
+                  'bottom': 'closed'}
+    rg = RasterModelGrid((3, 3), bc=boundaries)
+    gdp = GroundwaterDupuitPercolator(rg, recharge_rate=1.0e-8,
+                                      hydraulic_conductivity=0.01)
+    for i in range(12):
+        gdp.run_one_step(5.0e4)
+
+    assert_equal(np.round(gdp.thickness[4], 5), 0.00141)
+
+    # Re-instantiate to test the case when the necessary fields already exist
```
kbarnhart: this test is great!!! But I'm not sure what's happening with these last two lines.

landlab/components/groundwater/dupuit_percolator.py
```diff
@@ -0,0 +1,197 @@
...
+    recharge_rate: float, field name, or array of float
+            Rate of recharge, m/s
+            Default = 1.0e-8 m/s
+
+    Examples
+    --------
+    >>> from landlab import RasterModelGrid
+    >>> mg = RasterModelGrid((3, 3))
+    >>> z = mg.add_zeros('node', 'topographic__elevation')
+    >>> gdp = GroundwaterDupuitPercolator(mg)
```
kbarnhart: I think you need to show the component import from `landlab.components`. Not that this won't work at the moment without new `__init__.py` files being created/existing ones modified.

landlab/components/groundwater/dupuit_percolator.py
gregtucker: We need a standard... after __init__ is fine with me.

landlab/components/groundwater/dupuit_percolator.py
gregtucker: There's a tutorial notebook (or will be soon) for that purpose

landlab/components/groundwater/dupuit_percolator.py
gregtucker: There's a larger issue: what do we call fields that can be created a priori, but if not the component provides a default? Or creates via create_grid? I'm going to leave this as is for now until we come up with a general resolution/recommendation.

landlab/components/groundwater/tests/test_dupuit_percolator.py
gregtucker: Tests an "if" branch

landlab/components/groundwater/dupuit_percolator.py
gregtucker: good idea

PR #982, opened 2019-07-26
Gt/add line to coords method

gregtucker:
Adds a function to raster_funcs to do scan-line conversion, i.e., take two endpoints of a line segment in grid coordinates, and return an array containing (column, row) coordinates of the grid nodes that represent that line segment.

### DIFF
--- a/landlab/bmi/bmi_bridge.py
+++ b/landlab/bmi/bmi_bridge.py
@@ -15,7 +15,6 @@ exposes a Basic Modelling Interface.
 import inspect
 
 import numpy as np
-
 from bmipy import Bmi
 
 from ..core import load_params
--- a/landlab/ca/hex_cts.py
+++ b/landlab/ca/hex_cts.py
@@ -60,7 +60,7 @@ class HexCTS(CellLabCTSModel):
         initial_node_states,
         prop_data=None,
         prop_reset_value=None,
-        seed=0
+        seed=0,
     ):
         """
         HexCTS constructor: sets number of orientations to 1 and calls
@@ -103,5 +103,5 @@ class HexCTS(CellLabCTSModel):
             initial_node_states,
             prop_data,
             prop_reset_value,
-            seed
+            seed,
         )
--- a/landlab/ca/tests/test_celllab_cts.py
+++ b/landlab/ca/tests/test_celllab_cts.py
@@ -127,10 +127,7 @@ def test_raster_cts():
     nsg[6] = 0
     pd = mg.add_zeros("node", "property_data", dtype=int)
     pd[5] = 50
-    ca = RasterCTS(
-        mg, ns_dict, xn_list, nsg, prop_data=pd,
-        prop_reset_value=0, seed=1
-    )
+    ca = RasterCTS(mg, ns_dict, xn_list, nsg, prop_data=pd, prop_reset_value=0, seed=1)
     prior_first_event_time = event_time
     (event_time, index, event_link) = ca.priority_queue.pop()
     assert event_time != prior_first_event_time, "event times should differ"
--- a/landlab/components/chi_index/channel_chi.py
+++ b/landlab/components/chi_index/channel_chi.py
@@ -148,6 +148,7 @@ class ChiFinder(Component):
         min_drainage_area=1.0e6,
         reference_area=1.0,
         use_true_dx=False,
+        noclobber=True,
         **kwds
     ):
         """
@@ -168,6 +169,9 @@ class ChiFinder(Component):
             spacing along the channel (which can lead to a quantization effect,
             and is not preferred by Taylor & Royden). If False, the mean value of
             node spacing along the all channels is assumed everywhere.
+        noclobber : bool (default True)
+            Raise an exception if adding an already existing field.
+
         """
         if grid.at_node["flow__receiver_node"].size != grid.size("node"):
             msg = (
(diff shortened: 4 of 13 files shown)


gregtucker:
Note: I've asked for a review from any of @kbarnhart , @mcflugen , @nathanlyons , @SiccarPoint --- whoever happens to be available. This should be simple. It is intended for use in an application involving simulation of a cinder cone, in which it is necessary to map a ballistic trajectory (bounded by the launch point, and an end point representing an arbitrary upper limit to horizontal distance traveled) onto a RasterModelGrid.

kbarnhart:
@gregtucker this is super cool. I'm not really able to review at the moment, but I had a thought:
I think it makes sense to use the [scikit-draw implementation](https://scikit-image.org/docs/dev/api/skimage.draw.html#line) since it is optimized and the vision is to use this many times over.

gregtucker:
By all means, I invite anyone who wishes to go ahead and replace my implementation with scikit-draw. But in the meantime, I would get this merged so we can start using it. I think this implementation should be fairly efficient, as it uses numpy arrays.
The rounding error (which caused test fails that I'd missed) should now be fixed.

gregtucker:
Ok, the latest test fail is in a file I haven't touched (at least not recently):

  File "/home/travis/build/landlab/landlab/landlab/grid/tests/test_hex_grid/test_shape_dep_warning.py", line 20, in test_shape_dep_warning
    assert len(w) == 1
AssertionError: assert 3 == 1
  -3
  +1

Anyone recognize this?

kbarnhart:
@gregtucker as we discussed yesterday the pro/con between this and scikit line is pro: no dependency, con, maybe speed. 

So I tested the speed. 

```python
import time
import numpy as np
import matplotlib.pylab as plt
from skimage.draw import line
from landlab.grid.raster_funcs import line_to_grid_coords

nitter = 10000
size = 2**np.arange(2, 17)

landlab_mean = []
scikit_mean = []

for s in size:
    i = 0

    landlab = []
    skit = []
    while i < nitter:
        #choose a random starting point
        x0 = np.random.randint(0, 100)
        y0 = np.random.randint(0, 100)
        
        theta = np.random.uniform(0, 2*np.pi)
        
        x1 = int(x0 + s*np.cos(theta))
        y1 = int(y0 + s*np.sin(theta))
        
        t0 = time.time()
        out = line_to_grid_coords(x0, y0, x1, y1)
        llt = time.time()-t0

        t0 = time.time()
        out = line(x0, y0, x1, y1)
        skt = time.time()-t0
        
        i += 1 
        landlab.append(llt)
        skit.append(skt)
    landlab_mean.append(np.mean(landlab))
    scikit_mean.append(np.mean(skit))
    
plt.loglog(size, landlab_mean, label="landlab")
plt.loglog(size, scikit_mean, label="scikit")
plt.xlabel("length of line")
plt.ylabel("duration of evaluation")
plt.legend()
plt.show()
```
exampleduration


Summary is that both your and scikit-image scale, and that it is 2x faster, but that both are really fast.

gregtucker:
Cool, thanks for testing this!

gregtucker:
@mcflugen I changed a test that I think you wrote, could you take at look at this PR when you get a chance

kbarnhart:
@gregtucker I think this looks good (with the changes recommended in the comments below) so long as @mcflugen is OK with the changes you made to his test (they make sense to me). 

Documentation is nice, tests are nice, and performance is only 2x the equivalent in scikit-image. If landlab starts needing more things from scikit-image, I think we should consider removing this and adding the dependency, but I agree that we don't need to add the dependency for this function. 

When-if that happens, we should anticipate some issues related to what seem to be slight rounding changes between this and scikit-image's implementation of line. 

I think it would be a good idea to make the following changes to the interface and documentation (inspired by the scikit-image implementation, which I think has some nice features). These should be very minor changes to make to the interface, but I think they are well worth it. (Checkboxes to keep track of what is/is not done. 

 - [ ] I think that the inputs must int and that we should consider enforcing this. 
- [ ] I would recommend calling them row and column coordinates and using r and c in the function definition and documentation (e.g.,  (r0, c0, r1, c1) instead of x and y) to ensure people are not confused that they are actual x and y values. 
- [ ] I would recommend pointing the documentation to [np.unravel_index](https://docs.scipy.org/doc/numpy/reference/generated/numpy.unravel_index.html) which would permit someone to convert the node id to the (r0, c0, r1, c1) values.
- [ ] I would recommend changing the output to not be a number of line elements x 2 array but instead  to be the same output as scikit-image's line function:
    rr, cc(N,) ndarray of int
    Indices of pixels that belong to the line. May be used to directly index into an array, e.g. `img[rr, cc] = 1`.

Importantly this last one is not just something that keeps our interface in line with scikit-image, but would make the output directly usable by [np.ravel_multi_index](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ravel_multi_index.html#numpy.ravel_multi_index) which is what a user would want to use to get the landlab node ids associated with a line.

gregtucker:
Thanks @kbarnhart , good suggestions, changes made. @mcflugen please take a look at this when you get a chance, and approve if the test fix in hex looks ok to you.

kbarnhart:
looks good @gregtucker thanks for making those changes!

kbarnhart:
@mcflugen just approved this via teleconference.

PR #918, opened 2019-04-27
Barnhark/hacks law

kbarnhart:
Addresses #917 

@nicgaspar @nathanlyons @ amforte 

I was going to have to do this anyway to calculate Hacks (I don't have code to do it). 

Let me know what you think.

It relies on the new Channel Profiler (#878 )

I ** think** I"m calculating distance to the divide correctly.

PR #984, opened 2019-08-01
feat: add distance to divide

kbarnhart:
In #918 I added a hack calculator and a distance to the divide function to `v2_dev`.

For a variety of reasons, it would be nice to have the distance to the divide function in `master`. This PR does that.

### DIFF
--- a/landlab/utils/distance_to_divide.py
+++ b/landlab/utils/distance_to_divide.py
@@ -1,0 +1,234 @@
+#! /usr/bin/env python
+"""Functions to calculate flow distance from divide."""
+import numpy as np
+
+from landlab import BAD_INDEX_VALUE, FieldError, RasterModelGrid
+
+
+def calculate_distance_to_divide(
+    grid, longest_path=True, add_to_grid=False, noclobber=True
+):
+    """
+    Calculate the along flow distance from drainage divide to point.
+
+    This utility calculates the along flow distance based on the results of
+    running flow accumulation on the grid. It will use the connectivity
+    used by the FlowAccumulator (e.g. D4, D8, Dinf).
+
+    Parameters
+    ----------
+    grid : ModelGrid
+    longest_path : bool, optional
+        Take the longest (or shortest) path to a drainage divide. Default is
+        true.
+    add_to_grid : boolean, optional
+        Flag to indicate if the stream length field should be added to the
+        grid. Default is False. The field name used is ``distance_to_divide``.
+    noclobber : boolean, optional
+        Flag to indicate if adding the field to the grid should not clobber an
+        existing field with the same name. Default is True.
+
+    Returns
+    -------
+    distance_to_divide : float ndarray
+        The distance that has to be covered from an imaginary flow, located in
+        each node of the grid, to reach the watershed's outlet.
+
+    Examples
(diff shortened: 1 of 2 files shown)


landlab/utils/distance_to_divide.py
```diff
@@ -0,0 +1,234 @@
...
+import numpy as np
+
+from landlab import BAD_INDEX_VALUE, FieldError, RasterModelGrid
+
+
+def calculate_distance_to_divide(
+    grid, longest_path=True, add_to_grid=False, noclobber=True
+):
+    """
+    Calculate the along flow distance from drainage divide to point.
```
gregtucker: along-flow

landlab/utils/distance_to_divide.py
```diff
@@ -0,0 +1,234 @@
...
+    >>> from landlab.utils.distance_to_divide import (
+    ...     calculate_distance_to_divide)
+    >>> dx = 1
+    >>> hmg = HexModelGrid(5,3, dx)
+    >>> _ = hmg.add_field('topographic__elevation',
+    ...                   hmg.node_x + np.round(hmg.node_y),
+    ...                   at = 'node')
+    >>> hmg.status_at_node[hmg.boundary_nodes] = CLOSED_BOUNDARY
+    >>> hmg.status_at_node[0] = FIXED_VALUE_BOUNDARY
+    >>> fr = FlowAccumulator(hmg, flow_director = 'D4')
```
gregtucker: It's a bit odd that the name 'D4' applies here, though I understand why. For LL 2.0 we should change the name to, say, Dadjacent or something (or keep D4 for purposes of raster grids, but add Dadjacent as a synonym.)

landlab/utils/distance_to_divide.py
```diff
@@ -0,0 +1,234 @@
...
+    ...     noclobber=False)
+    >>> hmg.at_node['distance_to_divide']
+    array([ 3.,  0.,  0.,
+         0.,  2.,  1.,  0.,
+       0.,  1.,  1.,  0.,  0.,
+         0.,   0.,  0.,  0.,
+            0.,  0.,  0.])
+    """
+    # check that flow__receiver nodes exists
+    if "flow__receiver_node" not in grid.at_node:
```
gregtucker: This again points out the need for LL 2.0 to handle optional input fields. In this case, it's not optional, but in some cases, lack of a field prompts the component to create one and fill it with default values.

landlab/utils/distance_to_divide.py
```diff
@@ -0,0 +1,234 @@
...
+        to_one = True
+
+    flow__receiver_node = grid.at_node["flow__receiver_node"]
+    drainage_area = grid.at_node["drainage_area"]
+
+    # get the upstream node order
+    flow__upstream_node_order = grid.at_node["flow__upstream_node_order"]
+
+    # get downstream flow link lengths, result depends on type of grid.
+    if isinstance(grid, RasterModelGrid):
```
gregtucker: Ultimately we need to expand the menu of rectilinear grid types to include cases with nonuniform dx, dy, and when that day comes, this will need to reference whatever base class applies...

PR #989, opened 2019-08-29
fix: change datarecord to not inherit from Xarray Dataset

kbarnhart:
As part of the discussion regarding #909, and other needs of @pfeiffea and I with regard to the network sediment transporter, I'm replacing the now closed #948 with two PRs. This is the first of them. 

It changes DataRecord to not inherit from an Xarray Dataset. 

@margauxmouchene what do you think? 

I believe this will minimally impact ClastTracker, but please correct me if I am mistaken.

### DIFF
--- a/landlab/components/chi_index/channel_chi.py
+++ b/landlab/components/chi_index/channel_chi.py
@@ -195,7 +195,9 @@ class ChiFinder(Component):
         self._set_up_reference_area(reference_area)
 
         self.use_true_dx = use_true_dx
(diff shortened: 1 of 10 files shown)


kbarnhart:
@margauxmouchene thanks for taking a look. 

you are right that those two changes are un-related to DataRecord. I just added a few other minor formatting changes as well. Now when I run `make pretty` I get no changes off this branch. 

Assuming things pass, I will merge.

PR #995, opened 2019-09-07
Component to create profiles using endpoints

nathanlyons:
This component inherits from BaseProfiler to create profiles set up by endpoints selected by the user. This component replaces #608, which will be/is abandoned. For Profiler, I generally mimicked ChannelProfiler where it made sense so these profilers can work and read similarly.

Questions for reviewers.
- Is `Profiler` not clear? Is `EndpointProfiler` better?
- Is the attribute name `network_structure` fitting for profiles in general? Or is this a legacy of channel_profiler's creation as the first profiler? Sure, the profile segments of Profiler make a network, but I wonder if "network" along with network model grid (I  assume most often used for stream networks) will lead users to be confused by "network" for the data structure. One (e.g. I) could name this attribute something else because it just return `_net_struct`, although its nice when things work the same way, when possible. Alternatively, the data structure could be named something else (eg 'profile_structure' or just 'structure'). I also think users will figure it out if others think 'network_structure' works.
- An option of BaseProfile.plot_profiles_in_map_view is `endpoints_only`, which works great for Profiler (thanks @kbarnhart). To create a map view plot with Profiler this flag must be set to `True`, not its default using: `profiler.plot_profiles_in_map_view(endpoints_only=True)`. Is it worthwhile for Profiler to override BaseProfile.plot_profiles_in_map_view to set this default (just by a wrapper in Profiler)?
- Lets consider multi-segment Profiler profiles. The final sample of each segment is repeated at the first sample of the following segment, for example (note the ids):
```python
>>> profiler.network_structure[0]['ids']
[10, 11, 12, 13, 14, 15, 16]
>>> profiler.network_structure[1]['ids']
[16, 24, 32, 40, 48, 56, 64]
```
The repeat is necessary for BaseProfiler plotting. This could be problematic if one wishes to, for example, get elevation of the profile. One profile node will be extracted twice. Do you foresee issues if the  `network_structure` property returned a modified `_net_struct` with the duplicates removed? One issue could be if users wish to create figures themself (not those created by BaseProfiler).


kbarnhart:
thanks @nathanlyons. I expect I'll be able to get to this by end of Tuesday. Excited to see it in LL!!!

kbarnhart:
@nathanlyons some comments to your questions as I start my review:

- Is Profiler not clear? Is EndpointProfiler better?
    - I have no strong opinion on this. This is the most basic type of profiler I can think of, so I think it is reasonable that it get the name `Profiler`. 
- Is the attribute name network_structure fitting for profiles in general? Or is this a legacy of channel_profiler's creation as the first profiler? Sure, the profile segments of Profiler make a network, but I wonder if "network" along with network model grid (I assume most often used for stream networks) will lead users to be confused by "network" for the data structure. One (e.g. I) could name this attribute something else because it just return _net_struct, although its nice when things work the same way, when possible. Alternatively, the data structure could be named something else (eg 'profile_structure' or just 'structure'). I also think users will figure it out if others think 'network_structure' works.
    - I think you are correct in thinking that this name is a holdover and not appropriate. I am 100% fine with having _net_struct changed to something like _datastructure (more description than _struct). If you could make this change across all three components that would be AWESOME. 
- An option of BaseProfile.plot_profiles_in_map_view is endpoints_only, which works great for Profiler (thanks @kbarnhart). To create a map view plot with Profiler this flag must be set to True, not its default using: profiler.plot_profiles_in_map_view(endpoints_only=True). Is it worthwhile for Profiler to override BaseProfile.plot_profiles_in_map_view to set this default (just by a wrapper in Profiler)?
   - I have no strong opinion on this. Which probably means that I lean "override". 
- Lets consider multi-segment Profiler profiles. The final sample of each segment is repeated at the first sample of the following segment, for example (note the ids):
```
>>> profiler.network_structure[0]['ids']
[10, 11, 12, 13, 14, 15, 16]
>>> profiler.network_structure[1]['ids']
[16, 24, 32, 40, 48, 56, 64]
```
The repeat is necessary for BaseProfiler plotting. This could be problematic if one wishes to, for example, get elevation of the profile. One profile node will be extracted twice. Do you foresee issues if the network_structure property returned a modified _net_struct with the duplicates removed? One issue could be if users wish to create figures themself (not those created by BaseProfiler).
    - Channel profiler has this repeat and I think it should stay in the datastructure. I've run into it as a thing to pay attention to when calculating statistics is done (e.g. [line 302](https://github.com/landlab/landlab/blob/07db6133811bbc14d69f039efdc2f3e911505244/landlab/components/hack_calculator/hack_calculator.py#L302) in the HackCalculator). It can be dealt with by running `np.unique` on the nodes in question (as they should be unique, but the elevations are not necessarily).

landlab/components/profiler/profiler.py
```diff
@@ -0,0 +1,346 @@
...
+
+        X X X X X X X X X
+        X o X X X X X X X
+        X X * X X X X X X
+        X X X * X X X X X
+        X X X X * X X X X
+        X X X X X * X X X
+        X X X X X X * X X
+        X o . . . . . o X
+        X X X X X X X X X
```
kbarnhart: Love the diagram!

landlab/components/profiler/profiler.py
```diff
@@ -0,0 +1,346 @@
...
+         1: {
+            'ids': [16, 24, 32, 40, 48, 56, 64],
+            'distances': [6, 7.41, 8.83, 10.24, 11.66, 13.07, 14.49]
+            'color': (0.13,  0.57,  0.55,  1)
+            }
+        }
+
+    Examples
+    --------
+    >>> from landlab import RasterModelGrid
```
kbarnhart: I would explicitly state that this is the example from above.

landlab/components/profiler/profiler.py
```diff
@@ -0,0 +1,346 @@
...
+    node of the first segment.
+    >>> profiler.network_structure[1]['ids']
+    array([16, 26, 35, 45, 54, 64])
+
+    Alternative to nodes, profiles can be instantiated with coordinates.
+    >>> profiler = Profiler(mg, [(10, 10), (70, 10), (10, 70)])
+
+    Endpoints can also be set with a combination of coordinates and nodes.
+    >>> profiler = Profiler(mg, [(10, 10), 16, (10, 70)])
+    """
```
kbarnhart: These last two examples are great.

landlab/components/profiler/profiler.py
```diff
@@ -0,0 +1,346 @@
...
+            'distances': [0, 1, 2, 3, 4, 5, 6]
+            'color': (0.27,  0,  0.33,  1)
+            },
+         1: {
+            'ids': [16, 24, 32, 40, 48, 56, 64],
+            'distances': [6, 7.41, 8.83, 10.24, 11.66, 13.07, 14.49]
+            'color': (0.13,  0.57,  0.55,  1)
+            }
+        }
+
```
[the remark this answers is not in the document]
nathanlyons: I added 'no interpolation' to the first paragraph of component docstring. 

Interpolation and node sampling options can be future enhancements.

nathanlyons:
@kbarnhart 

Two checks are failing that you may know more about. 

The first is a docstring in DataRecord unrelated to Profiler. Should the dict be an OrderedDict?
```
3127________ [doctest] landlab.data_record.data_record.DataRecord.__init__ ________
3128149         >>> dr1.dataset.time.values
3129150         array([ 0.])
3130151 
3131152         >>> dr1.variable_names
3132153         ['mean_elevation']
3133154 
3134155         >>> dr1.dataset['mean_elevation'].values
3135156         array([100])
3136157 
3137158         >>> dr1.dataset.attrs
3138Expected:
3139    OrderedDict([('time_units', 'y')])
3140Got:
3141    {'time_units': 'y'}
3142
3143c:\projects\landlab\landlab\data_record\data_record.py:158: DocTestFailure
```
The second is because Profiler's required init parameters are a grid and an 'endpoints' parameter. Should Profiler be included in the list of components to exclude?
```
3144______________________ test_component_metadata[Profiler] ______________________
3145Traceback (most recent call last):
3146  File "C:\projects\landlab\tests\components\test_components.py", line 37, in test_component_metadata
3147    _ = Comp(grid)
3148TypeError: __init__() missing 1 required positional argument: 'endpoints'
3149
```

kbarnhart:
@nathanlyons, I can answer the second one (yes, needs to be excluded). 
regarding the first one @mcflugen fixed it on another branch (I think #1019)

nathanlyons:
@mcflugen Any ideas why this build is failing? Two attempts return this error:
```
Worker information
6
7
8No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
9Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
10
11The build has been terminated
12
```

kbarnhart:
@nathanlyons I've had that error a bunch over the weekend. I think there is some sort of server errors somewhere in the CI pipeline. I just restarted your builds...

nathanlyons:
@kbarnhart Out of curiosity, is it clear why your build restart worked? Before you restarted, I effectively restarted a build with a push and got the same error. Just timing relative to the existence of a server error?

I had to change the "net_struct" var name to "data_struct" in the hack calculator. Do you see anything else?

kbarnhart:
@nathanlyons I'm pretty sure its just bad timing.

kbarnhart:
Also, usually when the build fails before landlab is installed (as was the case for these errors), the fault is in the CI plumbing, not in our specification.

nathanlyons:
I see. Thanks for the explanation.

Issue #932, opened 2019-05-19
Importing landlab fails with ImportError: cannot import name 'factorial'


mcflugen:
With the release of *scipy* 1.3, importing landlab gives an import error from *statsmodels*,
```python
>>> import landlab
Traceback (most recent call last):
...
ImportError: cannot import name 'factorial' from 'scipy.misc'
```
This appears to be a problem with the *statsmodels* package and has been identified in statsmodels/statsmodels#5747. It looks like an easy fix but, until then, users will have to downgrade *scipy* to something less than 1.3. Thus far, *conda* only sees version 1.2 but 1.3 is now available on PyPI.

PR #998, opened 2019-09-12
fix: relax scipy version requirement

kbarnhart:
#932 identified that there was an issue between statsmodels and scipy. In response to that @mcflugen added the requirement [on this commit](https://github.com/landlab/landlab/commit/28ad18bd624608b905b0419e04136c4da10ff237). 

If this passes, CI then we can relax this requirement.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,7 +81,7 @@ setup(
         "netcdf4",
         "pyyaml",
         "pyshp",
-        "scipy<1.3",
+        "scipy",
         "six",
         "statsmodels",
         "xarray",


setup.py
```diff
@@ -81,7 +81,7 @@ def run(self):
         "netcdf4",
         "pyyaml",
         "pyshp",
-        "scipy<1.3",
```
TR160qg, line 84: Could you tell where to do it? I met this problem when i use Jupyternote book?

mcflugen: @TR160qg *landlab* does not require a particular version of *scipy*. What you are looking at here is an old pull request that removes the version requirement.

If you've encountered a problem using *landlab*, please submit an [issue](https://github.com/landlab/landlab/issues) describing your problem. My guess, if *landlab* is requiring an older version of *scipy*, is that you are using an older version of *landlab* and you could try updating it.

PR #1003, opened 2019-09-21
fix: 2.0 update for docs

kbarnhart:
This PR (when ready) will do the following:

- [x] Simple port of all wiki to rst and integrate with the rest of the docs
- [x] Ensure that all parts of Landlab are actually accessible in the docs
- [x] Be verified to makes sure the links work (since many are back to the wiki...)
    - [x] Remove links to wiki
    - [x] Replace hyperlinks to read the docs to sphinx page references. 
- [x] Fix broken images that existed on the wiki (e.g., examples)
- [x] Have closer to zero errors.
- [x] Move all links from nbviewer to binder (!!!)
- [x] Have zero errors
- [x] Enforce zero build errors AND zero broken links through CI

Things for the future
- [x] Update the standard name list (which gives inputs and outputs). 
- [x] ~Put binder links to tutorials in the component docstrings.~ I made Issue #1015 to handle this. 
- [x] After component PR is merged (#956) and grid PR (#997) is merged  update codebase docstrings to remove additional errors). 
 - [x] we can check for broken hyperlinks with `make linkcheck`. There are MANY broken links. @kbarnhart ~is committed to  fixing those that are NOT associated with the publications/presentations page. Those can be left for the contributors to fix on their own time.~ They are now all fixed.  
- [x] Make an issue to keep track of having someone copy edit the docs (@kbarnhart would be the 100% worst person for this job, @SiccarPoint would be good). This can be done in another PR. 

See the docs [here](https://landlab.readthedocs.io/en/barnhark-bring_wiki_to_docs/).

PR #1004, opened 2019-09-21
Move tests into a separate tests folder

mcflugen:
This pull request moves all unit tests from inside the landlab package into a separate folder outside of the package (e.g. the tests inside `landlab/grid/tests` are now in `tests/grid`). This will clean up our package and reduce the amount of stuff that is distributed with landlab.

Still remaining to do:
* [ ] move tests for the components
* [ ] update docs (?)


kbarnhart:
@mcflugen 🎉 

Few thoughts.

Docs happening over at #1003 (in progress, but they build!!!) I've done my best to keep all those changes INSIDE of the landlab/docs folder. Let's talk Monday about sequence of merging (as I assume some things will need to change in the docs once #997 is merged. There are also a number of doc warnings that I'd like to fix that occur because of bad formatting in component docstrings. I'm ignoring them for now since I wan't to make those change _after_ #956 is merged. 

I think that component test moving should happen in #956 (or in a subsequent PR after this and that PR are merged) as I've had to modify some tests as part of my changes.

mcflugen:
@kbarnhart For the most part, I'm just moving the tests, not making changes to the test files, so it shouldn't be a problem to merge this before #956. However, I'm happy to wait on this and merge it after you've merged #956. I don't think this needs to be included in #956.

kbarnhart:
@mcflugen I think you are right and this should merge first. Looks good!!

Makefile
```diff
@@ -51,11 +51,11 @@ clean-test: ## remove test and coverage artifacts
 	rm -fr .pytest_cache
 
 lint: ## check style with flake8
-	flake8 landlab --exclude=examples
+	flake8 landlab tests --exclude=examples
```
kbarnhart: I *think* we can remove the `--exclude=examples` as I removed the examples folder. 

I should have done that on #996.

kbarnhart:
A note, coverage decreases b/c the test files (all at 100% were counted towards the total).

PR #1005, opened 2019-09-22
test: require netcdf4 to be installed

mcflugen:
This pull request removes the checks to see if *netcdf4* is installed. A long time ago *netcdf4* was a pain to install and some users didn't have it installed so we didn't strictly require it to be present to install *landlab*. However, now *netcdf* is easy to install and we even list it as a dependency in *setup.py* and our *conda-forge* recipe. As such, the checks just added unnecessary complexity so I think it makes sense to remove them.


kbarnhart:
@mcflugen good idea! Looks good. 

Also looks like all the tests pass (even though one AppVeyor tests shows up as failing).

PR #1007, opened 2019-09-26
Add seepage to GroundwaterDupuitPercolator 

DavidLitwin:
Allows water table to intersect with the surface and create seepage. The definition of the hydraulic gradient has also been changed to allow for non-uniform slope of the aquifer impermeable base, in accordance with Troch et al 2003 (Hillslope Boussinesq equations).


gregtucker:
@DavidLitwin it looks as if there are a couple of test fails to be sorted out before reviewing this PR. To run the tests locally, use:

pytest --doctest-modules

(you may need to install pytest)

DavidLitwin:
@gregtucker Looking at the outputs of the failed tests it seems they all reference a name error for 'unsaturated_zone' in the line: `self._unsat = unsaturated_zone`

I don't see this line in the version of the code I sent in this pull request. Is it looking at commits before the most recent one?

DavidLitwin:
Ooops, shouldn't open my mouth before I do a real check. Latest commit got rid of it.

Issue #1010, opened 2019-10-03
Docstring for ExponentialWeatherer


DavidLitwin:
I was looking of the documentation for the ExponentialWeatherer on readthedocs and noticed a few possible errors in the equation. I also mentioned this to @gregtucker. As it is right now, the equation reads:

<a href="https://www.codecogs.com/eqnedit.php?latex=\dot{w}&space;=&space;w_0&space;\exp(d/w^*)" target="_blank"><img src="https://latex.codecogs.com/svg.latex?\dot{w}&space;=&space;w_0&space;\exp(d/w^*)" title="\dot{w} = w_0 \exp(d/w^*)" /></a>

The sign on the argument of the exponential should be negative, and some of the other notation seems unclear to me. There's a dot on the weathering rate, but not on the constant w0, and w is also used for the characteristic depth. Can I suggest something more like this? 

<a href="https://www.codecogs.com/eqnedit.php?latex=w&space;=&space;w_0&space;\exp(-d/d^*)" target="_blank"><img src="https://latex.codecogs.com/svg.latex?w&space;=&space;w_0&space;\exp(-d/d^*)" title="w = w_0 \exp(-d/d^*)" /></a>

kbarnhart:
@DavidLitwin thanks for making this comment. Best thing to do would be to make a pull request proposing these changes! Start by making a brach off of the `master` branch. Let me know if you have any questions.

PR #1012, opened 2019-10-04
Clarify ExponentialWeatherer docstring

DavidLitwin:
Clarifying the docstring of the ExponentialWeatherer, as per issue #1010.


kbarnhart:
Thanks for doing this @DavidLitwin ! 🎉 Noticing things like this and pointing out how they can be improved is a really valuable contribution to Landlab!

DavidLitwin:
You're welcome! It seemed like an easy fix to make it more readable. 👍

PR #1019, opened 2019-10-15
Remove deprecated functions

mcflugen:
This pull request addresses #925. I haven't yet removed all of the deprecated functions - but just about! A remaining one that's a bit tricky is,

`RasterModelGrid.calculate_slope_aspect_at_nodes_burrough`

(used in the radiation component). The alternatives don't quite seem to work. I probably just need to dig into it a little deeper.


kbarnhart:
I'm going to tag @saisiddu on this one since the remaining function impacts a component he wrote.

kbarnhart:
@mcflugen I think we need to deal with

`RasterModelGrid.set_fixed_link_boundaries_at_grid_edges` and 
`RasterModelGrid._calculate_gradients_at_d8_active_links`

These are two remaining instances I can find of "deprec*" in the codebase. These are not formally deprecated, but have deprecation mentioned in the comments.

mcflugen:
I've now removed all of the functions/methods that were marked as `@deprecated`.

I de-deprecated `RasterModelGrid.calculate_slope_aspect_at_nodes_burrough` as, it appears anyway, to work with a raster grid while the suggested alternatives do not (at least not straight away). Now that I think of it, perhaps it should be renamed to match the suggested alternatives. That way, users just have to call `calc_slope_at_nodes`/`calc_aspect_at_nodes`.

mcflugen:
> @mcflugen I think we need to deal with
> 
> `RasterModelGrid.set_fixed_link_boundaries_at_grid_edges` and
> `RasterModelGrid._calculate_gradients_at_d8_active_links`
> 
> These are two remaining instances I can find of "deprec*" in the codebase. These are not formally deprecated, but have deprecation mentioned in the comments.

Yes! I was just about to say the same thing.

mcflugen:
@kbarnhart The only remaining methods now are

* [ ] `RasterModelGrid._calculate_gradients_at_d8_active_links`
* [ ] `RasterModelGrid._calculate_gradients_at_d8_links`

which are only used in the flow directors. As you're most familiar with the flow routers, do you have any thoughts on this? I think, at the very least, they should be renamed to match our conventions (i.e. `calc_grad_at_d8`) and moved into `raster_gradients.py`. I guess we also no longer have methods like `calc_grad_at_active_d8`.

kbarnhart:
@mcflugen I don't have strong feelings about these names...

kbarnhart:
@mcflugen I've started down the path of fixing these... I've made progress, but need your assistance (perhaps we can talk after the call this afternoon). 

My vision: 

Both functions above removed and the addition of:


`calc_diff_at_diagonal
`calc_diff_at_d8`
`calc_grad_at_diagonal`
`calc_grad_at_d8`

Once these exist, then the flow direction functions should not be terrible to fix (I've half done it...)

I've made some progress on these, but I'm having a hard time figuring out how to deal with diagonal link directions correctly. 

Perhaps you can help by getting these four functions right, then I'll fix the use in flow direction?

mcflugen:
@kbarnhart I’ve actually already done this but just haven’t pushed the changes to github yet. Will do so later today.

kbarnhart:
cool. you can take or leave my commits (e.g., feel free to clobber them)

mcflugen:
@kbarnhart Could you please have a look at the *flow_director* stuff? Something's still not quite right there.

Something to note is that, despite their names, `_calculate_gradients_at_d8_active_links` returned gradients on diagonals and links (as you would expect), while `_calculate_gradients_at_d8_links` returned gradients on just diagonals.

kbarnhart:
@mcflugen I'm going to let you handle the merge conflicts.

PR #1020, opened 2019-10-18
fix: ordered dict

kbarnhart:
we just started to get some errors about ordered dicts on `v2_dev`. @mcflugen fixed this earlier, but it would be nice to separate from the deprecation PR.


kbarnhart:
FYI @nathanlyons

PR #1022, opened 2019-10-18
Add tutorials stage to travis build

kbarnhart:
As discussed in the call this afternoon, we presently don't verify that code that will go into a release version of landlab works against the release version of the tutorials repo. 

This PR (which aims to pull into `release` intentionally) is to fix that. 

The travis script will probably need some fixing... but its probably pretty close to right. 

As written it makes a tutorials stage on the **continuous-integration/travis-ci/pr** CI builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,10 +21,11 @@ jobs:
     - conda activate landlab_dev
     - pip install -e .
     script:
+    - pushd ../..
     - git clone https://github.com/landlab/tutorials.git landlab/tutorials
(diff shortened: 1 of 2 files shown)


kbarnhart:
@mcflugen this is ready for you to look at. In many ways it goes along with #1024

kbarnhart:
@mcflugen what do you think of this and the guidelines in #1024

kbarnhart:
Thanks @mcflugen.

PR #1024, opened 2019-10-19
update release notes

kbarnhart:
@mcflugen I've tried to update the release notes. They won't be quite right until #1022 is merged... but any edits are appreciated.


mcflugen:
@kbarnhart When creating a *landlab* release, can we not create the *landlab* tag on *master* and then merge *master* into *release*? It feels to me like merging *master* into *release* and then *release* back into *master* can somehow be avoided.

mcflugen:
@kbarnhart 🤔 I can see that my suggestion will have similar problems as what we have now. Maybe we could have a different set of tags along the *master* branch? Maybe with a *-dev* suffix like *v1.10.0-dev* or something like that? We would have to make sure that we don't deploy if the version doesn't have the *dev* suffix.

kbarnhart:
@mcflugen I'm was also a little stumped in thinking about this... so I went back to [the gitflow article](https://nvie.com/posts/a-successful-git-branching-model/#release-branches)  to reexamine some core ideas...

My (perhaps) radical conclusion: I think that we should get rid of `release`. Why do I say this:

1) It doesn't really give us anything (e.g., we just want `release`, to be the most recently tagged version), we can build the conda-forge pypi releases off of tagged versions of master. 

2) Wanting to keep the tags straight highlights that it is more trouble than its worth.

So I asked myself, what benefit does `release` actually give us... And the only example I could come up with was that it makes it really easy to checkout the most recently tagged version. That is, 

```bash
git checkout release
```

really just is shorthand for 

```bash
git checkout v1.x.x
```

And I think most users won't ever need to do this. Those that do will also be developing on development branches and shouldn't have a hard time checking out a tag.

So this is what I propose:

1) In landlab/landlab, we have no more `release` branch. We just have `master`, and make tags directly on master. This keeps our tags and branches cleaner, and ensures that people who are developing off of master can use `landlab.__version__` to get a dirty tag that is based off of the most recent version.

2) In landlab/tutorias we keep `release` and `next`, and put tags only on `release`. And we don't worry about tags on `next` (tags on this branch are less important b/c its not a package and doesn't use versioner).

If there is some benefit that you think the `release` branch gives us that I'm missing, please LMK.

mcflugen:
@kbarnhart one advantage of having a *release* branch is for bug fixes. Consider the following scenario,
* we release *landlab* *X.Y.Z* on *master*
* lots of stuff is added to *master* and it enters a state where it's not ready for a new release
* a bug is discovered in *vX.Y.Z* and we want to deploy a patch that fixes the bug

If we have a *release* branch, we just create a branch off of *release* (a [hotfix branch](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow)), fix the bug, and then merge back into *release*, tag, and redeploy. If we don't have a *release* branch, I'm not sure what we do. Perhaps we create a branch at the *X.Y.X* tag, fix the bug on there and then, what? Can we merge back into *master* at a point immediately after *X.Y.Z* and then update everything downstream of that? I don't think so, but I'm not sure.

One point of confusion when talking about *gitflow* is that, for historical reasons, we decided to have *master* be our development branch and to create a new branch, *release*, for releases. Thus, what we call *release* is *master* in *gitflow* and what we call *master* is *develop* in *gitflow*. We did this because we had already been using *master* for development for a long time and thought it would be confusing if we changed things up.

kbarnhart:
@mcflugen thanks for pointing out that what we call `master` is what gitflow calls `develop`, and that what we call `release` is what gitflow calls `master`...

That helps.

Under revised understanding I think we should prioritize the following. 

1. Forget about getting tags to make sense on `master`. I find this useful when debugging with people, but we can live without it. That is what git log is for...

2. These revisions should on creating a workflow that will solve a known problem: **ensuring that when we make a version release on landlab/landlab we end up with something that works with the tutorials repo FOR CERTAIN**. 

I revised some of the text to reflect this. What do you think?

The major change to the workflow is that rather than doing the merge into `release` locally, we would now do with with GitHub to ensure CI occurs.

mcflugen:
@kbarnhart Ok, I think that's good. I like the idea of creating pull requests when merging into *release*. I think this workflow should fix our problem with breaking the tutorials on the *release* branch.

For step 1 (merging *next* into *release* for the tutorials), should we include a `[skip ci]` in the commit so the tests aren't run yet?

kbarnhart:
spiffy. Good call on the `[skip ci]`.

PR #1025, opened 2019-10-20
Change noclobber keyword to clobber

### LINKED ISSUE #979: Change noclobber to clobber
noclobber is somewhat confusing... especially given the double negative of the option noclobber=False indicating that you want a field to be overwritten.  

@mcflugen and I propose that this keyword is changed to clobber=True in LL2.0

Thoughts?

mcflugen:
This pull request fixes #979 by changing the `noclobber` keyword to `clobber` for methods that add fields to a grid. The `noclobber` keyword was confusing when using the double negative,`noclobber=False`.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -42,6 +42,7 @@ build: false
 
 test_script:
   - "%WITH_COMPILER% pip install numpy"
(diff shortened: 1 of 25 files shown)


kbarnhart:
@mcflugen you missed a few instances of `noclobber` in the docs, but I got them.

PR #1026, opened 2019-10-22
Fix bug in GroundwaterDupuitPercolator, add symmetry test

DavidLitwin:
Fixes a missing sign in the gw velocity equations, adds support for locations where regolith thickness goes to zero, and ensures that the model produces a radially symmetric water table when aquifer base is radially symmetric and recharge is uniform over the domain.

### DIFF
--- a/landlab/components/groundwater/tests/test_dupuit_percolator.py
+++ b/landlab/components/groundwater/tests/test_dupuit_percolator.py
@@ -3,13 +3,13 @@
 """
 Created on Tue Jun  4 16:26:31 2019
 
-@author: gtucker
+@author: G Tucker, D Litwin
 """
 
 import numpy as np
 from numpy.testing import assert_almost_equal, assert_equal
 
-from landlab import RasterModelGrid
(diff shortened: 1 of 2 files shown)


landlab/components/groundwater/tests/test_dupuit_percolator.py
```diff
@@ -139,3 +138,34 @@ def test_conservation_of_mass_adaptive_dt():
     assert_almost_equal(
         (gw_flux + sw_flux + storage - storage_0) / recharge_flux, 1.0, decimal=3
     )
+
+def test_symmetry_of_solution():
+    """ test that water table is symmetric under constant recharge
```
kbarnhart: @DavidLitwin this is a great test! Nice work using the hexagonal grid!

kbarnhart:
@DavidLitwin Because this branch was named master. When I just delinted it I accidentally pushed to master... This showed up as merged... BUT then I reset master (since really we want to bring this into the master branch as a PR. So I"m going open a new PR for this issue.

PR #1029, opened 2019-10-24
Use xarray for netcdf io

mcflugen:
This pull request (which is still a work in progress) changes the *netcdf* io functions to use *xarray*. It will also:

* [ ] write operations:
    * [x] write fields at any grid location (not just node and cell)
    * [x] write field values by name and/or location
    * [ ] write fields at specified ids
    * [x] write grid-type (i.e. *RasterModelGrid*, *HexModelGrid*) specific data so that the file can be read back in and construct a grid of that type
    * [x] write boundary conditions

* [ ] read operations
    * [x] read fields at any grid location (not just node and cell)
    * [x] read field values by name and/or location
    * [ ] read fields at specified ids
    * [x] read grid-type specific data to construct a grid
    * [x] read boundary conditions

Because we will be able to read/write fields at multiple grid elements, and because fields at different elements can have the same name, they need to be saved with a name that differentiates them from fields with matching names on other elements. They way I've done this is by prefixing the name of each field with it's location. For example, if a field named *topographic__elevation* is defined at both *nodes* and *cells*, the resulting *netcdf* file will contain the following variables,
```
    at_cell:topographic__elevation  (cell) float64 ...
    at_node:topographic__elevation  (node) float64 ...
```


kbarnhart:
@mcflugen I have a question: 

Should a user expect that a netcdf written by landlab io can be read in by landlab io to recreate the same grid? I think the answer is yes... This has some implications for naming, as well as reading... Before I think more on the implications, however, I want to hear your answer to the first part of the question.

mcflugen:
@kbarnhart Yes, I would say definitely (assuming the user has chosen to write all the fields).

Something I forgot to add to my checklist was reading/writing boundary conditions. I'll add that now.

kbarnhart:
@mcflugen I like the naming scheme (and think it makes sense for output written by landlab). My only concern is  think one issue might be that a user reading in other data (e.g., z or precipitation or something, on a node, might find it irritating of we build in an expectation (to a reader) that at node fields are named `at_node:name_of_field`. 

what do you think. 

I also added an (empty) section to your checklist about reading.

mcflugen:
@kbarnhart Are you concerned about reading in other data, like *z* and *precipitation*, from random netcdf files? If so, I don't think there is any expectation that the variable names have the *at_* prefix.

Some other options:
1. different *at* fields are written to different *netcdf* files whose name indicates the grid elements where the fields are located (e.g. `my_grid_at_node.nc`)
2. different *at* fields are written to different *netcdf* [groups](http://xarray.pydata.org/en/stable/io.html#io-netcdf-groups) within the same file (e.g. `/at/node`, `/at/link`). I'm not sure this is supported by all *netcdf* versions (it may only be a *netcdf4* thing) or how viewers, like *paraview*, deal with it.
3. for fields that are only defined at one grid element, leave off the prefix. Only fields defined on multiple elements will have the *at_* prefix

kbarnhart:
If we make a reader that can read the landlab created outputs (e.g., expects that topographic elevation at node is stored as a variable called `at_node:topographic__elevation`, and that reader may also expect that dimensions have grid-element type names... ) ... which I think we want...

then I am concerned that such a reader will have a hard time with a really basic netcdf (e.g., if I have just an at-node field of precipitation (more likely than an at-node field of z, though I have used the before). 

ideally a reader based on xarray could deal with both cases.

mcflugen:
@kbarnhart I wasn't envisioning our *netcdf* grid readers and writers to read/write anything other than a *landlab* grid *netcdf* file. If a user would like to read data from some other *netcdf* file and add it to a grid, they would have to do it themselves (probably using *xarray*). Maybe something like the following would work,
```python
>>> grid = from_netcdf("grid.nc")
>>> data = xr.open_dataset("precip.nc")
>>> grid.at_node["precipitation"] = data["precip"]
```

We could also create functions that read a *netcdf* of some other format (maybe just some data on a rectilinear grid) and then try to create a new grid (probably a *RasterModelGrid*) and add the data to it.

kbarnhart:
@mcflugen I'm just thinking of how to maintain something like the current capability of `read_netcdf`.

mcflugen:
@kbarnhart I'm not sure how to handle the case you describe.

I see two cases:
1. A user has already created a grid and wants to add fields to it from a netcdf file. 
2. A user wants to read data from a *netcdf* file and, in addition, create a new grid and add the data to that grid.

In the first case, because *xarray* has made it so easy to read netcdf data, perhaps we should just leave it up to the user to read the data and add it to the appropriate fields.

In the second case, we need to support multiple ways of representing grids in (non-landlab style) *netcdf* files. I suspect this is mostly different ways of representing a `RasterModelGrid` (supporting cf conventions would probably be the most useful). For files that don't follow a particular convention, it should probably be up to the user to create the appropriate grid and then add the data.

For now, I would like this pull request to just concentrate on reading/writing landlab-style *netcdf*. What do you think?

kbarnhart:
@mcflugen I think your point about xarray making it so easy to read netcdfs is the key. I think your plan to focus on "Landlab netcdf" makes sense. 

I've updated the "one to two" guide outline on this branch with a note about this.

kbarnhart:
A quick note based on a conversation with @gregtucker and @mcflugen earlier today. 

1. We've decided that it is very important that we maintain interface/file format compatibility for read_netcdf and write_netcdf between LLv1 and LLv2 (while transitioning to use xarray under the hood). 

2. The rest of these things are additions of features, so its not the end of the world if they happen after the 2.0 release.

mcflugen:
@kbarnhart, @gregtucker I think this is pretty much (totally?) ready for a merge.

I thought more about this today and I think we should retain *read_netcdf* and *write_netcdf* not just for backward compatibility but also because they do slightly different things than the new *to_netcdf* and *from_netcdf* functions.

The *to_netcdf* function saves a *landlab* *ModelGrid* to a netcdf file so that it can be read back in and reconstructed with *from_netcdf*. As such, these netcdf files contain *landlab*-specific metadata (such as, for example, the *landlab* grid type (*RasterModelGrid*, *HexModelGrid*, etc.), *xy_spacing*, *xy_of_center*, etc.) so that *from_netcdf* can recreate a complete *landlab* grid (along with fields) using *from_netcdf*.

The *read_netcdf* function is intended to read data from a netcdf file that may have been created by a third party and either tries to construct a new grid from it, or read field values from it and place them onto an existing grid. These netcdf files will typically represent raster data. *write_netcdf* does the opposite and tries to save a grid and data to a similar format.

For this reason, I haven't added the *halo* keyword to the new functions since when reading data from a netcdf file created with *to_netcdf*, the halo would already be included in the file. Adding a halo is really only necessary when reading field data from a netcdf file and placing it onto an existing grid.

What do you think? Does that make sense?

kbarnhart:
@mcflugen I think that your are correct, but its hard to tell because I think there are some extra files. Some are files that were deleted but have now been resurrected. And some are files that are new (e.g., test output). Can you take a gander through the 36 changed files, remove those that are extra and then ping me. Then I'll do a closer look-through?

mcflugen:
@kbarnhart I'll have a look.

I'm not completely sure what's going on as I had just merged *master* into this branch and, it appeared at the time, the conflicts were resolved. I think the problem may be that the *v2_dev* branch was initially the target of this pull request but I later changed it to *master*.

landlab/grid/base.py
```diff
@@ -412,6 +414,88 @@ def __init__(self, **kwds):
         self._axis_units = tuple(np.broadcast_to(axis_units, self.ndim))
         self._axis_name = tuple(np.broadcast_to(axis_name, self.ndim))
 
+    def fields(self, include="*", exclude=None):
+        """List of fields held by the grid.
```
kbarnhart, line 418: 👍

landlab/layers/eventlayers.py
```diff
@@ -448,6 +448,11 @@ def event_layers(self):
         finally:
             return self._event_layers
 
+    @property
+    def at_layer(self):
+        """EventLayers for each cell."""
+        return self.event_layers
+
```
kbarnhart, line 455: I *think* this is supposed to be here and not a holdover, but just checking.

tests/io/netcdf/test_to_netcdf.py
```diff
@@ -0,0 +1,146 @@
+#! /usr/bin/env python
```
kbarnhart, line 1: I think a test for RadialModelGrid.to_netcdf() is missing...

And/or equivalent of the last test in test_from_netcdf.py

tests/io/netcdf/test_from_netcdf.py
```diff
@@ -0,0 +1,61 @@
...
+
+@pytest.mark.parametrize(
+    "grid_type", ["HexModelGrid", "RadialModelGrid", "RasterModelGrid"]
+)
+def test_from_grid(datadir, grid_type):
+    grid = from_netcdf(datadir / "test-{0}.nc".format(grid_type))
+    assert grid.__class__.__name__ == grid_type
+    assert_array_equal(grid.at_node["elev"], 1.0)
+    assert_array_equal(grid.at_node["temp"], 1.0)
+    assert_array_equal(grid.at_link["elev"], 0.0)
```
kbarnhart, line 61: I can't leave a comment on the binary files, but am I correct to see that those three files are all used by this file, right?
Should they be in a folder called data or does it matter?

mcflugen: Yes, they are in the folder `datadir`, which is a *pytest* fixture.

landlab/layers/eventlayers.py
mcflugen: Yes, I added it so that it would be more like *at_node*, *at_link*, etc.

tests/io/netcdf/test_to_netcdf.py
mcflugen: There are some tests for `RasterModelGrid.to_netcdf()` but I should add some more/better ones.

tests/io/netcdf/test_from_netcdf.py
kbarnhart: got it. sorry for being obtuse. was harder to figure that out than I expected in the GitHub interface....

PR #1038, opened 2019-11-01
Update gdp docstring, add setters, check wt elevation

DavidLitwin:
Updated gdp docstrings with math formatting. Also starts the joss paper for gdp.


kbarnhart:
@mcflugen do you know how to turn on building the docs for this branch (since it is on a fork). 

@DavidLitwin is working on making sure the math looks right (so local builds aren't sufficient b/c mathjax).

I know how to turn on branches.... but not branches on forks.

DavidLitwin:
@kbarnhart the other problem I'm having with this is that `black landlab` doesn't remove the invalid escape sequences that prevent it from passing on lint. Any ideas what to do there?

mcflugen:
@DavidLitwin You can fix your escape character problem by making your docstring a [raw string](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals). You can do this by just putting a `r` in front of the docstring string.

PR #1050, opened 2019-11-09
Fix #1049

### LINKED ISSUE #1049: v2, fields, tutorials
@mcflugen another bit of v2 codebase revisions + tutorials issue...

The [fields tutorial](https://github.com/landlab/tutorials/blob/v2_dev/fields/working_with_fields.ipynb)

gives a weird view when we use .keys... is there a better thing now that we are xarray based?

Also, something isn't working for keeping track of units.

Screen Shot 2019-11-08 at 7 05 33 PM

mcflugen:
This pull request addresses the issues found by @kbarnhart and described in #1049.

I've made only a couple small changes:
* `grid.at_node.keys()` now returns a list of field names at grid element. I believe this was caused by a change in *xarray.variables*
* when creating new fields, the *units* keyword is now passed all the way through so that it is saved in the field

With these changes, we now get:
```python
>>> grid.add_full("temperature", 2, at="node", units="m")
>>> grid.at_node.keys()
['temperature']
>>> grid.at_node.units["temperature"]
'm'
```

### DIFF
--- a/landlab/field/graph_field.py
+++ b/landlab/field/graph_field.py
@@ -263,7 +263,7 @@ class FieldDataset(dict):
         return self._ds
 
     def keys(self):
-        return self._ds.variables
+        return list(self._ds.variables)
 
     def set_value(self, name, value_array, attrs=None):
         attrs = attrs or {}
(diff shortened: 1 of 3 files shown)


kbarnhart:
🎉 thanks @mcflugen

kbarnhart:
I have confirmed that this fixes the issue with the field tutorial.

Issue #1048, opened 2019-11-09
v2, voronoi, and imshow_grid


kbarnhart:
@mcflugen I'm working on a branch in tutorials to update them to 2.0. When working on `set_BCs_on_voronoi.ipynb` I ran into the following error. I'm pretty sure its because `imshow_grid` hasn't been updated to deal with voronoi post graph2grid...?

<img width="1094" alt="Screen Shot 2019-11-08 at 5 24 41 PM" src="https://user-images.githubusercontent.com/6589339/68519196-ca1eaf00-024c-11ea-855f-25492076cf85.png">

mcflugen:
@kbarnhart Good catch. I think this one might be a little tricky. For this to work, we have to have a *scipy.spatial* representation of the voronoi grid (this is what `grid.vor` was), which is different from how we store the grid in a `VoronoiDelaunayGrid`.

When we first create a `VoronoiDelaunayGrid`, we use *scipy.spatial* to create the needed data structure but we then get rid of it because it doubles the size of the grid and is redundant. We *could* keep it around but it would only be for this use case.

I wonder if `landlab.plot.graph.plot_graph` is an acceptable alternative. If it is, great. If it isn't, perhaps we should change *plot_graph* instead of carrying around the extra data structure.

gregtucker:
I second the idea of not carrying around a heap of extra memory usage if it's not needed. But equally, it would be nice to have *imshow_grid* be grid-type agnostic. What would performance be like if we invoked *scipy.spatial* to create a *vor* data structure only when needed, i.e., when *imshow_grid* is called for a Voronoi?

kbarnhart:
@mcflugen I think that a main benefit of `imshow_grid` is that it is grid-type agnostic. I know I value using it it that way (and probably wouldn't use it as much if it weren't...). Voronoi is less well used, but hex is highly used (e.g., all the CA work)... So I think that something equivalent (and under the same function call) is required. 

I think it is OK to get rid of "show_elements".

This is to say that I don't think that `plot_graph` is an acceptable alternative to `imshow_grid` (unless what you mean to propose is that `imshow_grid` uses `plot_graph` underneath).

If creating the `spicy.spatial` DS is large in memory but fast in wall time, it could be created only when this function is called (but not stored).

If that is not possible, I'd recommend creating a PatchCollection (since pcolormesh only supports quadrilaterals).

mcflugen:
[...]

PR #1051, opened 2019-11-09
fix: imshowgrid in v2

kbarnhart:


kbarnhart:
Addresses #1048

kbarnhart:
@mcflugen this is the branch to take over on.

mcflugen:
@kbarnhart will you please have a look at this and let me know if it's working as you expect? I'm able to plot values located at cells of a *HexModelGrid* so, if it's not completely working, it's probably pretty close.

kbarnhart:
@mcflugen it almost works now...

The remaining thing to fix (I tried... but haven't finished... would be great if you could take a stab) is that color-for-closed doesn't presently work for for irregular grids. 

The last plot on "tutorials/boundary_conds/set_BCs_on_voronoi.ipynb" is a good example. 

https://github.com/landlab/tutorials/tree/v2_dev has the tutorials on it. 

Except for the component making tutorial and the two eco hydro everything now passes :)

mcflugen:
@kbarnhart That's great! 🎉 Thanks for your help. I'll take a stab at the rest tonight.

kbarnhart:
Thanks @mcflugen. If you end up thinking that it is worth adding show_elements back because the voronoi plots look weird.... probably easiest to add a LineCollection that shows all of the edges....

mcflugen:
@kbarnhart I've added back all the edges. I'm not sure if it's exactly the same as *show_elements* or not but the plots do look better.

I've fixed the *color_for_closed* issue but found a new problem when plotting a field defined at cells. It should be an easy fix.

mcflugen:
@kbarnhart could you have another look? I think it's all fixed now. I made a few changes,
* set cell edges to be black lines
* fixed issue with the face color of cells of closed boundary nodes
* fixed plotting of fields defined at cells
* changed *vmin*, *vmax* to be min/max over all values being plotted

kbarnhart:
@mcflugen other than setting the default to have black outlines, your changes look great.

This is because when you are plotting a small grid, sometimes you want to see the edges... other times, I don't want to see the edges...

I added back the `show_elements` option, and implemented it with a line collection. It plots the faces.

Its not 100% the same as the prior option (which showed the "half faces" extending away from the grid in between the nodes that don't have cells). I don't think its critical to put that back. 

Assuming this passes, I'll merge it.

kbarnhart:
@mcflugen how do you make a new release on the `conda-forge/label/dev` channel?

Once this and #1053 are merged, I'd like to make a new release so I can test the v2_dev branch notebooks on binder. 

I assume the next version is `v2.0.0b1`, right?

mcflugen:
@kbarnhart On your fork of *conda-forge/landlab-feedstock*
* checkout the *devel* branch
* create a new branch for *v2.0.0b1*
* update *meta.yaml*
* create a pull request to merge back into ***devel*** (it's important to merge it back into *devel*, not *master*)

PR #1055, opened 2019-11-10
Remove uses of old node and link status flags

mcflugen:
This pull request removes uses of the old node and link status flag constants. The old constants are still there, though, and can be used in the same way as before for backward compatibility. I've removed uses of the following constants,
* `CORE_NODE`
* `FIXED_VALUE_BOUNDARY`
* `FIXED_GRADIENT_BOUNDARY`
* `LOOPED_BOUNDARY`
* `CLOSED_BOUNDARY`
* `ACTIVE_LINK`
* `INACTIVE_LINK`
* `FIXED_LINK`

In doing all of this, I realized that sometimes it is desirable to want to use these constants without importing a *ModelGrid* (and then using e.g. `ModelGrid.BC_*`). For these use cases, I introduced two new *enum* classes, `NodeStatus` and `LinkStatus` (*enums* only became available in Python 3.4ish). These mirror those constants bound to `ModelGrid`,
* `NodeStatus.CORE` -> `ModelGrid.BC_NODE_IS_CORE`
* `NodeStatus.FIXED_VALUE` -> `ModelGrid.BC_NODE_IS_FIXED_VALUE`
* `NodeStatus.FIXED_GRADIENT` -> `ModelGrid.BC_NODE_IS_FIXED_GRADIENT`
* `NodeStatus.LOOPED` -> `ModelGrid.BC_NODE_IS_LOOPED`
* `NodeStatus.CLOSED` -> `ModelGrid.BC_NODE_IS_CLOSED`
* `LinkStatus.ACTIVE` -> `ModelGrid.BC_LINK_IS_ACTIVE`
* `LinkStatus.INACTIVE` -> `ModelGrid.BC_LINK_IS_INACTIVE`
* `LinkStatus.FIXED` -> `ModelGrid.BC_LINK_IS_FIXED`

Truthfully, I think I like these two new *enum* classes better. I also think they may fit in better when we address the problem of different components having different sets of boundary conditions as I think the solution will be to create a new class(es) to hold just boundary conditions.

### DIFF
--- a/docs/source/user_guide/grid.rst
+++ b/docs/source/user_guide/grid.rst
@@ -452,7 +452,7 @@ either *active* or *inactive* (Figure 3).
 A closed boundary is one at which no flux is permitted enter or leave, ever.
 By definition, all links coming into or out of a closed boundary node must be inactive.
 There is effectively no value assigned to a closed boundary; it will probably have a
-BAD_INDEX_VALUE or null value of some kind.
+grid.BAD_INDEX_VALUE or null value of some kind.
 An open boundary is one at which flux can enter or leave, but whose value is controlled
 by some boundary condition rule, updated at the end of each timestep.
 
--- a/docs/source/user_guide/landlab_one_to_two.rst
+++ b/docs/source/user_guide/landlab_one_to_two.rst
@@ -35,15 +35,26 @@ TODO List of these, including replacements if any.
 Changes to field creation
 -------------------------
 - no more `noclobber`, instead we have `clobber`
+- recommended field init (eg. at="")
 
 Changes to Boundary Condition Flags
 -----------------------------------
-- no more use of unbound flags (e.g., `from landlab import CLOSED_NODE`)
+- no more use of unbound flags AND unbound flags are not importable
+  from top level or grid submodule (e.g., `from landlab import CLOSED_NODE`)
   now we use `Grid.BC_NODE_IS_CLOSED`.
 
-Changes to Field initialization
--------------------------------
-- PUT HERE IF THIS HAPPENS
+  Old name                  New name
+  ========================= ====================================
+  BAD_INDEX_VALUE           ModelGrid.BAD_INDEX
+  CORE_NODE                 ModelGrid.BC_NODE_IS_CORE
+  FIXED_VALUE_BOUNDARY      ModelGrid.BC_NODE_IS_FIXED_VALUE
+  FIXED_GRADIENT_BOUNDARY   ModelGrid.BC_NODE_IS_FIXED_GRADIENT
+  LOOPED_BOUNDARY           ModelGrid.BC_NODE_IS_LOOPED
+  CLOSED_BOUNDARY           ModelGrid.BC_NODE_IS_CLOSED
+  ACTIVE_LINK               ModelGrid.BC_LINK_IS_ACTIVE
+  INACTIVE_LINK             ModelGrid.BC_LINK_IS_INACTIVE
+  FIXED_LINK                ModelGrid.BC_LINK_IS_FIXED
+
 
 Standardization and deprecation within the component library
 ------------------------------------------------------------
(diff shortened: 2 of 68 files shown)


kbarnhart:
@mcflugen what about `BAD_INDEX_VALUE`? If all of these flags should be bound, that should be bound too. 

My commit on this is:

491fe6672d144d106fa4bc45378ba67868e33757

mcflugen:
@kbarnhart I agree. After having done it, I was starting to second guess and wonder if there might be a better way. I've added it as an attribute of *ModelGrid*, along with the others.

What do you think about renaming `BAD_INDEX_VALUE` to just `BAD_INDEX`? Is that too drastic? Maybe just `BAD_INDEX` for the bound version? It just seems like the extra `_VALUE` doesn't really add anything except more typing and longer lines.

kbarnhart:
@mcflugen I'm most of the way through adding my commit (and a few other things into this PR)

I think that we "could" change the name... we are changing other names... but I'm not sure that removing _VALUE adds much. 

In my (currently being tested) changes I removed the flags from the top level and grid submodule __init__.py files. I did this because, if the point of binding the flags is to use the bound versions, then why make it easy to import the unbound versions. 

Let me know if you think this is extreme.

mcflugen:
@kbarnhart I like the idea of removing the unbound constants from the top-level and grid submodule imports. However, if we're going that far, it seems like we should just completely remove them. I imagine most everyone using those constants were importing them straight from *landlab* and so it would break their code and then they would use the new versions.

I'm fine with removing them completely.

kbarnhart:
I'm not sure what you mean by 

>  However, if we're going that far, it seems like we should just completely remove them.

mcflugen:
@kbarnhart I can't convince you about `BAD_INDEX` rather than `BAD_INDEX_VALUE`?

kbarnhart:
I don't have strong feelings about BAD_INDEX, so I'm find if you want to change it.

mcflugen:
@kbarnhart I meant that if, for example, `CORE_NODE` can not be imported from either `landlab` or `landlab.grid` but, instead, only from `landlab.grid.nodestatus`, then it's effectively removed from users. My thought was that if `CORE_NODE` can only be accessed through `landlab.grid.nodestatus`, we might as well get rid of it completely, which I am all for.

mcflugen:
5892542 So much trailing whitespace!

kbarnhart:
I understand. I think it makes sense to only expose them as bound to the new NodeStatus and LinkStatus objects.

SiccarPoint:
I'm watching this go past by email... A workflow of being able to hard-set the BCs of individual nodes on the grid is incredibly valuable, and I assume this won't be going away?? What will be the workflow to directly replace e.g.
```
mg.status_at_node[mynode] = CLOSED_NODE
```
?

kbarnhart:
```python
mg.status_at_node[mynode] = mg.NODE_IS_CLOSED
```

SiccarPoint:
Awesome

kbarnhart:
The basic idea is that it is valuable to bind the BC flags to the grid (and to use them instead of -1 or -4)

SiccarPoint:
Seems reasonable. Though I assume they are still 1s, 4s etc under the hood?

mcflugen:
@SiccarPoint Correct, the values under-the-hood haven't changed.

kbarnhart:
@mcflugen I'm going to let you finish this up. I think the remaining things are

- [ ] BAD_INDEX_VALUE > BAD_INDEX
- [ ] Remove unbound BC flags in node and link status files

Once those are done, this looks good to me to merge. 

And once it and #1056 are merged, I think we are ready for a `v2.0.0b1` release.

mcflugen:
@kbarnhart I've made those changes. Once this branch is merged, I'll prepare a *v2.0.0b1* release.

PR #1064, opened 2019-11-21
feat: add new way to set a color

kbarnhart:
@nathanlyons I realized that it would benefit the component tutorial (and similar uses) to be able to set a single color for a plot of the channel profiler (it can be repeatedly called, and the color can represent time). 

Its a small change @nathanlyons let me know if you think it's OK. 

I've made the change in the tutorials repo as well.


nathanlyons:
Good idea and works for Profiler.

PR #1065, opened 2019-11-23
component to model species evolution

nathanlyons:
This component, SpeciesEvolver simulates biologic evolution in response to landscape change.

Intro to files in the component folder:
- species_evolution.py: The component. It is species type agnostic. It also knows nothing about zones, described below, so approaches without zones can use this component.
- zone.py: These objects conceptually exist in a portion of the grid (the ``mask`` attribute indicates where). They help to evaluate landscape connectivity. It is tempting to think of zones as habitats, although this distinction is avoided because zones may not spatially align with habitats depending on the species implementation.
- zone_controller.py: This object manages `zone` objects and methods to batch populate zones with species.
- base_taxon.py: Subclasses of taxa inherit this class. BaseTaxon is not intended to be used itself.
- zone_taxon.py: This subclass of BaseTaxon is associated with `zone` objects. The landscape connectivity inferred from zones are used to evaluate macroevolution rules.
- record.py: A data structure to store variables relative to time. Data is stored in a dictionary. `SpeciesEvolver` and `ZoneController` both have the attribute, ``record_data_frame``. This attribute is a record instance transformed to a Pandas DataFrame that makes it easy to see and work with the data stored in the record.

This pull request includes a tutorial in `notebooks/tutorials/species_evolution`.


landlab/components/species_evolution/base_taxon.py
```diff
@@ -0,0 +1,129 @@
...
+
+        This property must be implemented in a subclass.
+        """
+        # pragma: no cover
+
+    @abstractmethod
+    def _evolve(self, dt, stage, record, id):
+        """Run the evolutionary processes of the taxon.
+
+        SpeciesEvolver loops through the evolution processes of extant taxa in
```
nicgaspar: This implies to me that there can be more than one 'stage' in a 'dt' but if this is the case, is a 'total_number_of_stages' required?

landlab/components/species_evolution/record.py
```diff
@@ -0,0 +1,190 @@
...
+            The value of `var_name` that has the type, T. `nan` is returned if
+            `var_name` does not exist in the record.
+        """
+        if var_name not in self.variables:
+            return np.nan
+
+        time = self._check_time(time)
+
+        idx = self._get_time_index(time)
+        return self._dict[var_name][idx]
```
nicgaspar: This, along with `set_value` and `increment_value` require that one knows the type of a specific var, so as not to set it incorrectly.  Is a `get_type` overkill? Maybe...

landlab/components/species_evolution/species_evolver.py
```diff
@@ -0,0 +1,673 @@
...
+
+    Model time and other variables can be accessed with the class attribute,
+    ``record_data_frame``. Time is tracked to construct phylogeny. The unit of
+    time is not considered within the component other than the record, and can
+    be thought of as in years or whatever unit is needed. Time is advanced
+    with the ``dt`` parameter of the ``run_one_step`` method.
+
+    The geographic ranges of the taxa at the current model time are evaluated
+    during the ``run_one_step`` method. Taxa persistence or extinction is
+    determined by the taxon objects. Creation of child taxa is also determined
```
nicgaspar: taxa and taxon are not intuitive words to me in that I keep thinking that taxa is singular and taxon is plural. This is just me, but if you changed "child taxa" to "taxa children" it seems more clear to me.

landlab/components/species_evolution/species_evolver.py
```diff
@@ -0,0 +1,673 @@
...
+    by these objects. Taxa metadata can be accessed with the attribute,
+    ``taxa_data_frame``.
+
+    Taxa are automatically assigned unique identifiers, ``uid``. Identifiers
+    are used to reference and retrieve taxon objects. Identifiers are assigned
+    in the order taxa are introduced to SpeciesEvolver.
+
+    The development of this component was inspired by SEAMLESS (Spatially
+    Explicit Area Model of Landscape Evolution by SimulationS). See Albert et
```
nicgaspar: I like that you reference James' work. But why not also reference your own? I think there is also a way to include references automatically with the code now. Something that @mcflugen built?

landlab/components/species_evolution/species_evolver.py
```diff
@@ -0,0 +1,673 @@
...
+    True
+
+    Additional examples of controller usage are provided in ZoneController
+    documentation.
+
+    The ``mask`` of the zone is True where the conditions of the zone function
+    are met. All nodes of the grid are included because the elevation of each
+    node is below 100.
+
+    >>> zc.zones[0].mask
```
nicgaspar: Why is it `zones[0]`? is the `[0]` because you could index the different zones if different ones existed?

landlab/components/species_evolution/species_evolver.py
```diff
@@ -0,0 +1,673 @@
...
+
+    . . . x . . .       key:    . node in zone mask
+    . . . x . . .               x node outside of zone mask
+    . . . x . . .
+
+    Run a step of both the ZoneController and SpeciesEvolver. Two zones exist
+    following this time step.
+
+    >>> delta_time = 1000
+    >>> zc.run_one_step(delta_time)
```
nicgaspar: Why do you have to run line 159 to get that there are two zones? Isn't that completely taken care of in line 158?

landlab/components/species_evolution/species_evolver.py
```diff
@@ -0,0 +1,673 @@
...
+
+    The split of the initial zone triggered speciation. Taxon 0 became extinct
+    as it speciated to child taxa, 1 and 2.
+
+    >>> se.taxa_data_frame
+         appeared  latest_time  extant
+    uid
+    0           0         1000   False
+    1        1000         1000    True
+    2        1000         1000    True
```
nicgaspar: Why does taxon 0 die and taxa 1 and 2 evolve? I thought what happens to taxa when a zone gets split up is under the control of the user. Is this just default behavior? If yes, maybe best to say?

landlab/components/species_evolution/species_evolver.py
```diff
@@ -0,0 +1,673 @@
...
+        },
+    }
+
+    def __init__(self, grid, initial_time=0):
+        """Instantiate SpeciesEvolver.
+
+        Parameters
+        ----------
+        grid : ModelGrid
+            A Landlab ModelGrid.
```
nicgaspar: I think this is actually grid agnostic, as in one could use any grid type, right? If I'm wrong, then maybe state RasterModelGrid here.

landlab/components/species_evolution/species_evolver.py
```diff
@@ -0,0 +1,673 @@
...
+
+        >>> se.taxa_data_frame
+             appeared  latest_time  extant
+        uid
+        0           0         1000   False
+        1           0         1000   False
+        2        1000         2000   False
+        3        1000         2000    True
+        4        1000         2000   False
+        5        1000         2000    True
```
nicgaspar: Somehow it is very confusing to me that when you show all the data in the `taxa_data_frame` it does not show ancestor information, but then on line 545, one can get the ancestor information using `get_taxon_objects`. If I'm missing something that has a simple explanation, maybe add it.

landlab/components/species_evolution/zone.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+@unique
+class Connection(IntEnum):
+    """Zone connection type.
+
+    Connection type represents the connectivity of zones in two time steps. It
+    is described with the pattern, x-to-y where x and y are the descriptive
+    counts (none, one, or many) of zone(s) at the earlier and later time step,
+    respectively. For example, a connection type of one-to-many is a zone in
```
nicgaspar: is `one-to-many` a zone in the earlier time step that spatially overlapped ONE ZONE BUT in the later time OVERLAPS MULTIPLE ZONES?

nicgaspar: I think that might be weird language that I used above, but it shows my confusion. Maybe `one-to-many` actually means one zone broke up into many zones? But then I am unclear how a `none_to_none` would be used, because how can you label a zone that didn't exist prior or currently?

landlab/components/species_evolution/zone_controller.py
```diff
@@ -0,0 +1,416 @@
...
+    · x · + + ·    · x · + + ·    · x · + + ·    · x · + + ·
+    · x · · · ·    · x · · · ·    · x · · · ·    · x · · · ·
+    · · x · o ·    · · @ · o ·    · · x · · ·    · · · · · ·
+    · · · · · ·    · · · · · ·    · · · · · ·    · · · · · ·
+
+    The grid boundary affects zone creation. Zones can include nodes along the
+    boundary, although because boundary nodes are not associated with cells
+    that have area, boundary nodes do not contribute to the area summation of
+    clusters. The area summation only takes into account the cells associated
+    with core nodes.
```
nicgaspar: I would be careful saying "nodes along the boundary". That could imply nodes along the perimeter, and in Landlab, those are always boundary nodes. But boundary nodes can also be in locations that are not the perimeter of the grid. So maybe reword to:

The grid perimeter affects zone creation. All perimeter nodes are boundary nodes in Landlab. Zones can include nodes along the perimeter, although because perimeter nodes are not associated with cells that have area, perimeter nodes do not contribute to the area summation of clusters. [NOW THIS IS ME GUESSING, BUT I THINK HAS TO BE RIGHT GIVEN THAT LAST SENTENCE] Any nodes that are designated as boundary nodes but do not lie on the perimeter of the grid are not included in the area summation. The area summation only takes into account the cells associated with core nodes.

nicgaspar: Also, given the discussion about perimeter/boundary, I would not include the 2 value in row five, column 6.

One more thing - given the discussion of D4 vs. D8, does this mean that a RasterModelGrid must be used?

landlab/components/species_evolution/zone_controller.py
```diff
@@ -0,0 +1,416 @@
...
+    zone clusters are along grid boundaries.
+
+    zone function  D8             D4             D8             D4
+    returned mask  min area = 0   min area = 0   min_area = 2   min_area = 2
+    * · · * * ·    + · · x x ·    + · · x x ·    + · · · · ·    · · · · · ·
+    · * · · * ·    · + · · x ·    · # · · x ·    · + · · · ·    · # · · · ·
+    · * · · · ·    · + · · · ·    · # · · · ·    · + · · · ·    · # · · · ·
+    * · · · * *    + · · · o o    $ · · · o o    + · · · o o    · · · · o o
+    · · · · * ·    · · · · o ·    · · · · o ·    · · · · o ·    · · · · o ·
+    · * * · · ·    · @ @ · · ·    · @ @ · · ·    · · · · · ·    · · · · · ·
```
nicgaspar: In the D8 min_area = 2 case I find it a bit confusing that you get rid of the "three x" case that exists in the D8, min_area = 0 case. I think you do this because two of the x's don't count because they are on the perimeter, but then do the pluses that are on the perimeter really count?

landlab/components/species_evolution/base_taxon.py
nathanlyons: Yes, the `_evolve` Taxon method can be called multiple times (each a "stage") in one time step (when run_one_step of SpeciesEvolver is called). Although `total_number_of_stages` parameter is not required.

SpeciesEvolver `run_one_step` and Taxon `_evolve` method are designed in such a way that taxon types can have different numbers of stages. ZoneTaxon requires 2 stages. Future taxon types may have a different number of stages. Its up to the Taxon creator to consider what's going on in the stages of other taxon.

landlab/components/species_evolution/record.py
nathanlyons: ...maybe or maybe not. The record is not exposed to users. Its just handy to use with Taxon types. So only those gaining deeper familiarity with SpeciesEvolver will use Record and will by then know the types. I've made note to consider this down the line.

landlab/components/species_evolution/species_evolver.py
nathanlyons: I relate and my intuition also switches taxon and taxa. I think I'm getting used to it.

In "Creation of child taxa is also", child is modifying taxa so flipping the words doesn't happen to fit here. I changed this part to

> Each taxon object determines if it persists or becomes extinct, and if it creates child ``Taxon`` objects.

landlab/components/species_evolution/species_evolver.py
nathanlyons: I will add the citation once we have it using `_cite_as`.

landlab/components/species_evolution/species_evolver.py
nathanlyons: Yes, that's all. I added some text to explain.

landlab/components/species_evolution/species_evolver.py
nathanlyons: I have considered moving the SpeciesEvolver call down to focus this part of the example on zones. Although, I want to always demonstrate these two running directly together to keep time in sync. So I split it up this way:

```python
    Run a step of both the ZoneController and SpeciesEvolver.
    >>> delta_time = 1000
    >>> zc.run_one_step(delta_time)
    >>> se.run_one_step(delta_time)
    
    Two zones exist following this time step.
    
    >>> len(zc.zones) == 2
    True
```

landlab/components/species_evolution/species_evolver.py
nathanlyons: What happens to taxa is programmed into the Taxon classes. Its up to the Taxon programmer, which is in this case me, because I made ZoneTaxon demoed here.

The tree directly below this section shows taxon 0 doesn't die, but evolves into taxa 1 and 2 because, as you pointed out, the zone splits.

I added this short summary under the tree:

> The split of the initial zone into two zones at time 1000 triggered taxon 0
    to evolve into two child taxon objects. Taxon 1 occupies a zone on one side
    of the mountain range, and taxon 2 occupies a zone on the other side.

landlab/components/species_evolution/species_evolver.py
nathanlyons: The component is grid agnostic. ZoneController is not. At some point I did indicate such. I must have overruled myself. Anyways, ZoneController grid parameter type is correct now.

landlab/components/species_evolution/species_evolver.py
nathanlyons: The data frame is to summarize taxa and get_taxon_objects with the ancestor parameter gets the objects that share that ancestor.

Also, taxa have multiple ancestors, so it wouldn't work well in this 2-d table. The ancestor in get taxon objects is a common ancestor. Parent uid could be added to table but its less crucial than other info. Decisions were made to minimize the amount of data added to the taxa data frame.

I'll consider explanations that don't add confusion.

landlab/components/species_evolution/zone.py
nathanlyons: The second comment is accurate. `none_to_none` might be useless. Including it did make programming the zone functions more consistent.

landlab/components/species_evolution/zone_controller.py
nathanlyons: Modified text given your first comment.

Removing the 2 value in the upper grid example maybe requires larger grid examples to demonstrate the key combinations of D4, D8 and min area. The node of the 2 as a part of that zone doesn't contradict perimeter considerations as well as how they are handled by ZoneController.

ZoneController is limited to RasterModelGrid. SpeciesEvolver is not. Changed as I described in a previous response.

landlab/components/species_evolution/zone_controller.py
nathanlyons: Thanks for the helpful questions.

The three x's are not removed in the D8, min_area = 0. They are removed in the D8, min_area = 2 case because only one of the x's are not on the perimeter.

In which case are you asking about the +'s? They are only removed in the D4, min_area = 2 case because the single +'s are diagonal from other +'s, therefore are in clusters below the area threshold.

landlab/components/species_evolution/base_taxon.py
nicgaspar: So basically `_evolve` is called once per stage, so if there were 7 stages, it would be called 7 times? That was not clear to me. I think the taxon type has to know what stage or stages it evolves at, right? So if a taxon does nothing at stage 1, and stage got a value of 1, then nothing happens. And that would all be controlled by the creator of the subclass, right?

Possibly this would be more clear to me if it stated explicitly that this might not do anything at some stages, depending on what stages the taxon actually evolves at.

landlab/components/species_evolution/record.py
nicgaspar: That helps. Because I'm just reading code who sees what isn't obvious to me yet.

landlab/components/species_evolution/species_evolver.py
nicgaspar: I see. Basically you would never want to run ZoneController without immediately running SpeciesEvolver, right? That makes sense now that you say it, but I appreciated you saying it. What if you said that explicitly? Something like:
Two zone exist following the `run_one_step` call to the ZoneController.
SpeciesEvolver should always be called immediately after ZoneController, so that zones and species do not get out-of-sync.

Maybe that's not exactly right, but you see where I'm going.

landlab/components/species_evolution/species_evolver.py
nicgaspar: OK, so maybe you could just include the text you just gave me - i.e. the Taxon programmer can change these details. That was not obvious to me here, but I thought it was the case.

landlab/components/species_evolution/species_evolver.py
nicgaspar: Got it. Don't worry about adding anything. It will probably just make things more confusing.

landlab/components/species_evolution/base_taxon.py
nathanlyons: What you describe is stated in the abstract _evolve method of the base class, exemplified in ZoneTaxon, and will be demoed in a future tutorial.

landlab/components/species_evolution/species_evolver.py
nathanlyons: Added text.

Issue #1066, opened 2019-11-24
Update GW on v2_dev


kbarnhart:
need a repeat of #1047 onto v2_dev. it has some merge conflicts, so I'm going to do it later.

kbarnhart:
Addressed by #1076

PR #1071, opened 2019-12-14
Add tutorial notebooks to repo

mcflugen:
This pull request adds the tutorial notebooks from [landlab/tutorials](https://github.com/landlab/tutorials) to the main *landlab* repository. By default, the notebook tests are skipped (since they take a lot of time to run), however, using the *--test-notebook* flag for pytest will run them along with all the rest.

I think we should add some documentation regarding the notebooks to the *readthedocs* and maybe have a table that points users to versions of the tutorials that match *landlab* releases.

Maybe something along the lines of:

landlab version | Download
--- | ---
latest | *landlab-tutorials-latest.tar.gz* |
2.0 | *landlab-tutorials-v2.0.tar.gz* |
1.10 | *landlab-tutorials-v1.10.tar.gz* |


kbarnhart:
This is great! I’ll look at it tomorrow. 

I like the idea you have for a table with the version downloads. I’ll think about an approach that is easy to maintain. 

Do you think it makes sense to add the teaching tutorials in the same PR?

kbarnhart:
@mcflugen one additional thing...

[This PR](https://github.com/landlab/tutorials/pull/82) includes work @DavidLitwin did for the groundwater component. It never made it into v2_dev because... nz+agu. 

I'm working on the PR to address #1066 right now (updating v2_dev with @DavidLitwin 's changes to the main source code).

I want to make sure getting this tutorial through the transition doesn't slip through the cracks...

kbarnhart:
quick question for @mcflugen 

did you make sure that all the teaching notebooks have been re-saved with no output? I will admit to deciding to ask this instead of looking through all the notebooks....

I did this for all of the other notebooks. 

I did a tiny amount of research on ways to automate this... 
e.g., https://jupyter-notebook.readthedocs.io/en/stable/extending/savehooks.html

mcflugen:
@kbarnhart I did. I wrote a script to do this and was planning to add the script as a pre-commit hook so that whenever someone commits a change to a notebook, that notebook will be checked for any saved output. Either the commit would aborted, or the output would be automatically removed.

The teaching tutorials are failing because they need to be updated to landlab v2. I think it's mostly that they use *FlowRouter* instead of *FlowAccumulator*. I'm not sure but I think that should be an easy fix. I'll have a look at it tonight. I think that's the only thing left for this pull request.

kbarnhart:
@mcflugen I think it would make sense to add the groundwater component tutorial (see [this PR](https://github.com/landlab/tutorials/pull/82) in the tutorials repo) since it also need `v2`-ifying. 

Now that #1066 is merged, this notebook will work with v2. 

Then in my subsequent PR I'll deal with the re-org of all the notebooks, etc.

mcflugen:
@kbarnhart I've fixed all but one of the teaching tutorials so that they are *v2-ified* and run without error. I'm having trouble with the *stream_power_channels_class_notebook.ipynb* notebook as it uses the old *channel_profiler*. I switched it to use *ChannelProfiler* but it's still not quite right. If you have any experience with the new and old channel profilers, would you please have a look? If not, who should we ask for help?

kbarnhart:
@mcflugen I can do it. I'm the right person to update something to use the ChannelProfiler.

.travis.yml
```diff
@@ -103,6 +86,6 @@ install:
 - pip install numpy
 - pip install -e .
 script:
-- pip install pytest pytest-cov pytest-datadir coveralls
-- pytest --cov=landlab --cov-report=xml:$(pwd)/coverage.xml -vvv
+- pip install pytest pytest-cov pytest-datadir coveralls jupyter nbformat dask holoviews
```
kbarnhart: quick thought: since jupyter, dask, and holoviews are things that are not landlab, but are required to run the notebooks... how do you think we best indicate this to users?

mcflugen: @kbarnhart I agree. I was thinking of doing something along those lines.

kbarnhart: cool. I'll just add it now since I'm working on fixing the notebook (mostly fixed now)

mcflugen:
I've reorganized our requirements and environments files. My goal was to reduce the amount of duplicated requirements.

* *requirements.txt*: This is the requirements needed by *landlab*. *setup.py* now reads in this file when setting the *install_requires* keyword (i.e. `pip install -e .` will install everything in *requirements.txt*).
* *requirements-testing.txt*: This is used in our CI to install stuff just needed for testing (e.g. coveralls, pytest, etc.).
* *notebooks/requirements.txt*: This contains additional requirements for running the notebooks (e.g. dask, holoviews, etc) but that aren't needed for *landlab* generally. This is used by our CI so that the notebooks can be tested, as well has for *Binder* (see the next item).
* *notebooks/environment.yml*: This is a conda environment file that is meant to be used for *Binder*. It contains *landlab* as a requirement and then points to *notebooks/requirements.txt* it install the notebook-specific requirements.

.travis.yml
mcflugen: @kbarnhart I've made some changes to address this. See what you think.

kbarnhart:
@mcflugen can you make me a collaborator on your repo. I have changes to make the stream power teaching tutorial work (and a few other things) and I need to push to your branch.

mcflugen:
@kbarnhart I've checked the "allow edits from maintainers" box, is that good enough?

kbarnhart:
@mcflugen I like the approach to managing files. I'm not 100% sure that it will work for binder, I'll investigate...

environment-dev.yml
```diff
@@ -2,21 +2,9 @@ name: landlab_dev
...
-- netcdf4
-- pyshp
-- pyyaml
-- scipy
-- statsmodels
-- pandas
-- xarray
 - pip:
-    - bmipy<1
+  - -r file:requirements.txt
```
kbarnhart: @mcflugen does this form mean that these dependencies will be installed with pip instead of with conda-forge?

I also think that we need this file to include the notebook dependencies (I'm pretty sure that it doesn't at present)

notebooks/environment.yml
```diff
@@ -0,0 +1,8 @@
+name: landlab_notebooks
+channels:
```
kbarnhart: I'm almost certain that it makes sense to get rid of this file and the associated req.txt in the notebook folder. 

Binder needs a file called environment.yml at the top level directory (almost certain of this... hard to check since binder is being kinda funny today)

We also have a somewhat complex binder config b/c we want an environment AND we want setup.py to get run.

environment-dev.yml
mcflugen: Yes, this means install everything that's in *requirements.txt* with *pip*. I don't think I've listed *jupyter* as a requirement anywhere, so, yes, we may need to add it.

notebooks/environment.yml
mcflugen: I would be surprised if *Binder* doesn't allow you to specify a configuration file of your choice. However, if it doesn't, we can just move this up a level.

I was thinking we ***don't*** want to run *setup.py* for *Binder*. My thought was that the *Binder* examples would be for users wanting to use the latest landlab release - not the development version.

kbarnhart: Binder has a lot of configuration file options (https://mybinder.readthedocs.io/en/latest/config_files.html#postbuild-run-code-after-installing-the-environment) that it considers in a relatively standard order. 

I'm not aware of a way to indicate which environment file one wants to use, just that if you want the most common approach, you need to have a file called environment.yml at the top level. 

I think you are right about using the release version of landlab. that makes sense.

notebooks/teaching/README.md
```diff
@@ -0,0 +1,51 @@
+
```
kbarnhart: @mcflugen can you help me understand why this .md was put here. I know that it means that the teaching folder will have this as a readme... but a bunch of this information is outdated AND I think we want to put most of this information on the readthedocs page rather than in the GitHub repo.... 

Is is just here so the content isn't lost?

kbarnhart:
For reference. This is the binder link to this branch

https://mybinder.org/v2/gh/mcflugen/landlab/mcflugen/add-notebooks-to-repo

notebooks/teaching/README.md
mcflugen: I just copied this over from the other repository without reviewing it. I agree that we don't need it.

environment-dev.yml
kbarnhart: Don't we want to use conda-forge for as many things as possible?

notebooks/teaching/README.md
kbarnhart: cool. I'll make sure to revise it when I do the doc revision related to tutorials that we discussed yesterday (and agreed should be a PR that happens after this one). 🎉

setup.py
```diff
@@ -75,17 +75,7 @@ def run(self):
...
-        "matplotlib",
-        "netcdf4",
-        "pyyaml",
-        "pyshp",
-        "scipy",
-        "statsmodels",
-        "pandas",
-        "xarray",
-    ],
+    install_requires=open("requirements.txt", "r").read().splitlines(),
```
kbarnhart: This is great!

mcflugen:
> On second thought... I'm not sure I think this makes sense for the following reasons:
> 1) I don't think it actually makes it possible for binder to work
> 2) Lots of things get installed via pip instead of conda-forge

1. I just tried *Binder* and, I think, things work with *pip* installing dependencies from a requirements file.
2. Yes, most everything is installed with *pip*. I'm not sure this is a problem, though. It's what we were already doing on Travis and AppVeyor.

My goal was to reduce the number of places dependencies were listed and using requirements files seemed to be the most flexible: *setup.py*, *.travis.yml*, *appveyor.yml*, and *Binder* all point to the same requirements file.

Our CI had already been setup to *pip* install everything so it seemed to make sense to extend that to *Binder* as well. Someone wanting to *conda* install the dependencies can still do `conda install --file=requirements.txt` to install things from *conda-forge*.

I'm definitely open to suggestions, though, but this seemed to be the easiest way to keep things straight.

PR #1047, opened 2019-11-08
Improvements on GDP

DavidLitwin:
Repeat of #1038, only from branch on landlab/landlab. Math is better formatted in the docstring of the GroundwaterDupuitPercolator, more setters and getters are added, check water table elevation above land surface elevation, and the begin files for the JOSS paper.

PR #1076, opened 2019-12-18
Repeat 1047 update GDP

kbarnhart:
This serves as the equivalent to #1047 on v2_dev

@DavidLitwin I got all but one test to pass. Can you take a look.


tests/components/groundwater/test_dupuit_percolator.py
```diff
@@ -176,3 +176,44 @@ def test_symmetry_of_solution():
...
+    wt = grid.add_zeros("node", "water_table__elevation")
+    _ = grid.add_zeros("aquifer_base__elevation", at="node")
+    wt[:] = elev + 1
+
+    # initialize the groundwater model
+    gdp = GroundwaterDupuitPercolator(grid, recharge_rate=0.0)
+    gdp.run_one_step(1)
+    assert_equal(wt[4], 1)
+
+    grid = RasterModelGrid((3, 3))
```
kbarnhart: @DavidLitwin this looks like a repeat of the above test. true?

DavidLitwin: The difference is the method  - with and without adaptive timestep solver, I think.

DavidLitwin:
Hi Katy, I just checked out `barnhark/repeat_1047` locally on my computer and the groundwater test returns an import error... Was this what you saw? Or something else?

`ImportError: cannot import name 'map_pairs_to_values' from 'landlab.graph.sort.ext.remap_element' `

DavidLitwin:
Never mind on that. I ran `python setup.py develop` and everything in `test_dupuit_percolator.py` passes for me.

kbarnhart:
@DavidLitwin the docstring doesn't pass for me. I suspect that this means I didn't do the merge quite right... and I couldn't tell where I might have gone wrong...

Also. I just split that test into two to make the difference clear. 

```
landlab/components/groundwater/dupuit_percolator.py F                    [100%]

=================================== FAILURES ===================================
_ [doctest] landlab.components.groundwater.dupuit_percolator.GroundwaterDupuitPercolator _
128     Advance timestep
129
130     >>> dt = 1E3
131     >>> for i in range(1000):
132     ...     gdp.run_one_step(dt)
133
134     Calculate surface water flux out of domain
135
136     >>> fa.run_one_step()
137     >>> np.testing.assert_almost_equal(gdp.calc_sw_flux_out(),0.0005077)
UNEXPECTED EXCEPTION: AssertionError('\nArrays are not almost equal to 7 decimals\n ACTUAL: 0.00025070206902455539\n DESIRED: 0.0005077')
```

DavidLitwin:
Ok, thank you for splitting it. Let me see if I can figure out the docstring test issue...

DavidLitwin:
It seems the velocity calculation was reverted to an older incorrect version when merged. I just fixed it, try now.

kbarnhart:
@DavidLitwin thanks for finding and fixing this issue. 

Might you take a look at the rest of the merge just to double check it? (You may have already done this)

Most of the times merges are easy, but this one was a bit complex...

DavidLitwin:
@kbarnhart I've done a line-by-line comparison between `barnhark/repeat_1047` and my latest branch, `dlitwin/gw_and_faster_depression_finding`. As near as I can tell, there are no other errors.

kbarnhart:
Thank you @DavidLitwin !!!

kbarnhart:
@mcflugen There are two links in our docs that are now broken with something like 

```
/Users/barnhark/projects/landlab/landlab/docs/source/user_guide/papers_presentations.rst.rst:26:broken link: https://doi.org/10.1130/G46219.1 (('Connection aborted.', OSError("(54, 'ECONNRESET')")))
make: *** [linkcheck] Error 2
```

Both the links work for me in a browser... and I reproduce the error locally with sphinx's link check 

any ideas about how to continue enforcing that links work, but not deal with weird connection errors...?

mcflugen:
@kbarnhart I saw that too. I was hoping it was just temporary and that it would fix itself. Unfortunately, it's been like this for a few days now so maybe not. I don't have any good ideas other than to turn off the link checks, which is not at all ideal.

kbarnhart:
I'm going to fiddle around with some of the [linkcheck options](https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-the-linkcheck-builder)

I feel pretty strongly about NOT turning off link check (just because of how easily it is to get stale links...) since we now have *almost* entirely functional links, I want to make sure we can distribute link fixing across developers...

kbarnhart:
I found a lame workaround with `linkcheck_ignore`

PR #1081, opened 2019-12-20
Post 2.0 documentation update

kbarnhart:


kbarnhart:
@mcflugen the prior release update failed because of some doc links (since the docs have now been updated). This is the doc updating PR... once it is in master, and then release, and then those two are synced, I'll make v2.0.0b3

PR #1084, opened 2019-12-21
Synch SHA

kbarnhart:


mcflugen:
@kbarnhart if this becomes a pain, maybe we can, instead, put a tag on *master* like *v2.0.0-dev* or something like that?

mcflugen:
@kbarnhart I've kick-started a new build on *conda-forge* for *v2.0.0b3*.

mcflugen:
@kbarnhart All tests have passed for conda-forge/landlab-feedstock#22. Any reason not to merge yet?

kbarnhart:
@mcflugen 

1) Only failure in CI is that the 2.7 deploy didn't work (which makes sense). We can make a PR that removes this. 
2) Thanks for making the build on conda-forge -- while we are in beta, will we need to do these by hand? Or will the bot start recognizing them?
3) I don't think we'll be able to sync the sha this way... So I think your plan to update the tags makes sense. 

I'll remove the unnecessary deploy stage.  Can you make a PR that updates [the release notes](https://landlab.readthedocs.io/en/release/development/practices/dev_guide_releases.html) with tagging instructions and what you did to hand create the release on conda-forge?

kbarnhart:
@mcflugen updating after this particular set of attempting to synchronize version, etc. I found that my version was still at v1.11... So I added a v2.0.0-dev tag locally to master and have pushed it. Hopefully we find that it doesn't trigger anything. Presuming that is the case, I think we just make two tags with every release (not sure why we didn't think of that...)

PR #1094, opened 2020-01-07
update papers/presentations

gregtucker:
Added some 2019 papers and 2018 presentations to the docs


mcflugen:
@gregtucker Wow! That's getting to be a long list! 🎉

PR #1098, opened 2020-01-13
remove conda-forge channel from environment-dev.yml

kbarnhart:
@amanaster2 has brought to our attention an install issue that this may fix. Either way, since we now use pip for the dev environment, it is redundant.


kbarnhart:
@mcflugen if this passes, I think its OK to merge, right? 

nowhere do we truly test this environment (we just use the various `requirements-xxx.txt` files for travis and appveyor...)

Do you think we need to test this file?

@amanaster2 and @gregtucker I've added text as we discussed on the call, might you review it?

kbarnhart:
:wave: @mcflugen @amanaster2 @gregtucker

mcflugen:
@kbarnhart We could test this file but I'm not sure it would be worth the longer build times. What do you think?

I don't see why this can't be merged.

kbarnhart:
agreed. merging.

Issue #1109, opened 2020-01-15
Create builds for Python 3.8?


kbarnhart:
@mcflugen I was just testing some of our install instructions... I blindly made an empty conda environment and installed landlab with 

```
pip install --pre landlab
```

this totally failed with `ModuleNotFoundError: No module named 'numpy'`

It took me a moment to figure out that this was because the environment I had created was python 3.8. 

Perhaps we should start building for 3.8. Do you know of any barriers to doing that?

Error log below:

```
Last login: Wed Jan 15 09:42:27 on ttys000
(base) barnhark@rgnt-2-71-169-edu ~ % conda create --name test
Collecting package metadata (current_repodata.json): done
Solving environment: done


==> WARNING: A newer version of conda exists. <==
  current version: 4.8.0
  latest version: 4.8.1

Please update conda by running

    $ conda update -n base -c defaults conda



## Package Plan ##

  environment location: /Users/barnhark/opt/anaconda3/envs/test



Proceed ([y]/n)? y

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
# To activate this environment, use
#
#     $ conda activate test
#
# To deactivate an active environment, use
#
#     $ conda deactivate

(base) barnhark@rgnt-2-71-169-edu ~ % conda activate test
(test) barnhark@rgnt-2-71-169-edu ~ % conda install pip
Collecting package metadata (current_repodata.json): done
Solving environment: done


==> WARNING: A newer version of conda exists. <==
  current version: 4.8.0
  latest version: 4.8.1

Please update conda by running

    $ conda update -n base -c defaults conda



## Package Plan ##

  environment location: /Users/barnhark/opt/anaconda3/envs/test

  added / updated specs:
    - pip


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    setuptools-45.0.0          |           py38_1         646 KB  conda-forge
    ------------------------------------------------------------
                                           Total:         646 KB

The following NEW packages will be INSTALLED:

  ca-certificates    conda-forge/osx-64::ca-certificates-2019.11.28-hecc5488_0
  certifi            conda-forge/osx-64::certifi-2019.11.28-py38_0
  libcxx             conda-forge/osx-64::libcxx-9.0.1-1
  libffi             conda-forge/osx-64::libffi-3.2.1-h6de7cb9_1006
  ncurses            conda-forge/osx-64::ncurses-6.1-h0a44026_1002
[...]

PR #1111, opened 2020-01-15
Add Python 3.8 builds to AppVeyor and Travis

mcflugen:
This pull request addresses issue #1109 as described by @kbarnhart. That is, we don't build/test/deploy (to PyPI) with Python 3.8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
 - osx
 env:
   matrix:
+  - CONDA_ENV=3.8
   - CONDA_ENV=3.7
   - CONDA_ENV=3.6
   global:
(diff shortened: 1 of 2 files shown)


mcflugen:
@kbarnhart Could you have a look at this?  The docs are failing to build because of a link that *linkcheck* think is broken (but I don't think it is) but, otherwise, things seem to be building with Python 3.6, 3.7, and 3.8.

kbarnhart:
@mcflugen looks like the doc error was a fluke. Now passes. Thanks for addressing this!

mcflugen:
Thanks @kbarnhart!

PR #1116, opened 2020-01-17
SpeciesEvolver (make pretty and circular import)

kbarnhart:
@nathanlyons I ran make pretty locally which ended up changing a bunch of SpeciesEvolver files. Reorganizing the import order also resulted in a circular import (which I fixed). 

So this PR
- Runs make pretty on SpeciesEvolver (and friends)
- Fixes the circular import

I hope this is OK with you... LMK if it is not and I'll just close the PR.

### DIFF
--- a/landlab/components/species_evolution/__init__.py
+++ b/landlab/components/species_evolution/__init__.py
@@ -6,13 +6,7 @@
 """
 from .species_evolver import SpeciesEvolver
 from .zone import Zone
-from .zone_taxon import ZoneTaxon
 from .zone_controller import ZoneController
+from .zone_taxon import ZoneTaxon
 
-
-__all__ = [
-    'SpeciesEvolver',
-    'Zone',
-    'ZoneTaxon',
-    'ZoneController'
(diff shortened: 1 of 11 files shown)


nathanlyons:
@kbarnhart, thanks for doing this.

Would it work if we hold up on doing this for a bit? I've done work locally that may conflict, at least for my simple mind, and its at a critical time to avoid complications. I won't forget to handle this.

kbarnhart:
@nathanlyons sounds good. I removed comparable changes in #1117 so now that can be merged and won't affect the species evolver component (that PR now makes no changes to the species_evolver submodule).

At some point, can you add the following to the docstring (obviously changing the references as you see fit). 

```
References
----------
**Required Software Citation(s) Specific to this Component**

None Listed

**Additional References**

None Listed
```

nathanlyons:
@kbarnhart It worked out that I made pretty and included the new ref section in the component docstring in #1152. So I think this PR can be closed.

Issue #1108, opened 2020-01-15
JOSS Review: GroundwaterDupuitPercolator, address discoverability


kbarnhart:
In the [JOSS Review Issue](https://github.com/openjournals/joss-reviews/issues/1935) @rreinecke has made the following comment regarding discoverability of the tests, documentation, notebooks, dependencies for the this component. 

> Concerning the submodule I strongly disagree but that is something you need to decide and will not be part of this review. Nevertheless, I already have a problem in verifying the functionality of the software as I cannot find according unit tests (and how to run them). If they are somewhere else in the project please add a detailed description to your repository e.g. a list of submodules etc.. Please also explain there where to find which software component. I'm aware that there is another official Landlab page but as my time, and the time of the other reviewer and of course potential future users, is limited please consider to extend your readme. Also install instructions are missing together with a list of dependencies. Furthermore, did you implement a simple example model which demonstrates the capabilities of the new package? If so where? And where do I find the according documentation?

I will address this issue on a forthcoming Pull Request.

kbarnhart:
I've created two PRs to address this issue.

First, #1110 updates core landlab documentation regarding installation, tests, and dependencies. 

Second, #1112 updates the submodule Readme. 

Once these two PRs are merged, I'll ping @rreinecke to asses whether we have addressed his concerns, or if there are still outstanding issues that need to be addressed on this point.

kbarnhart:
@rreinecke -- In discussion with @DavidLitwin we realize that we are not sure what you mean by:

> Please also explain there where to find which software component.

Could you clarify so we can make sure to address this aspect of your comment.

rreinecke:
@kbarnhart Certainly. I'm referring to the main README in the landlab project. I know this review is for the groundwater component only, but as you choose to maintain it in this repository my comments will of course also address possible issues with the landlab project itself.

I'm missing a comprehensible README that:
* Tells me how to install the software
* What the dependencies are
* Were to find examples
* Maybe a short example
* How to run the tests
* Guides me through the package structure of the project
* Where to find the extended documentation
* And tells me who is responsible and how to contribute

I think everything is somewhere in your project and probably on your homepage but it is very cumbersome to find in the gitlab repository (which is under review not your homepage). Please revise the main README. They are many other github projects that do a great job in providing the information mentioned above. Maybe checkout other landing pages to improve yours.

kbarnhart:
@rreinecke thanks for clarifying. I'll re-ping you here when I've revised.

[...]

PR #1120, opened 2020-01-22
Add'l doc and readme updates

kbarnhart:
@DavidLitwin I think this should do for addressing #1108.

Let me know what you think.


kbarnhart:

Issue #1105, opened 2020-01-15
JOSS Review groundwater component: JOSS paper


rreinecke:
Please consider adding a short paragraph on:
State of the field: Do the authors describe how this software compares to other commonly-used packages?

kbarnhart:
[Linking to the JOSS Review Issue](https://github.com/openjournals/joss-reviews/issues/1935)

DavidLitwin:
I have addressed this [here](https://github.com/landlab/landlab/commit/bb2db392b868a9f96f6d2a8829911d3187856d7c) by updating the docstring for the component to caution users about model assumptions. 

The supplemental theory page [here](https://landlab.readthedocs.io/en/master/reference/components/dupuit_theory.html#dupuit-theory) also describes the assumptions made when using the Dupuit-Forcheimer approximation, and how this affects the governing equations solved.

DavidLitwin:
Referencing pull request #1122.

rreinecke:
@DavidLitwin very nice paragraph. I think this provides a much more complete view on the issue. (even though I like to again to self-promote my publication here: https://www.geosci-model-dev.net/12/2401/2019" - maybe be future research into this topic ;))

rreinecke:
@DavidLitwin can be closed as soon as it is merged.

PR #1122, opened 2020-01-23
Address #1105

DavidLitwin:
Changes to the JOSS paper for GroundwaterDupuitPercolator, adding a paragraph on state of field, what other groundwater models are available.


joss/paper.md
```diff
@@ -35,7 +35,9 @@ bibliography: papers.bib
...
...ailable for different purposes. Fully coupled groundwater and surface water models such as PARFLOW [@kollet_integrated_2006], CATHY [@camporese_surface-subsurface_2010], and PIHM [@qu_semidiscrete_2007] solve the three-dimensional Richards equation for variably saturated flow, and couple this with precipitation and runoff components. MODFLOW [@langevin_modflow_2019] also solves the three dimensional Richards equation, and may be coupled to precipitation and runoff models, as in GSFLOW [@regan_gsflow_2018]. More parsimonious models are also available that solve the hillslope storage Boussinesq equations, which model groundwater flow in quasi-two dimensional hillslopes [@marcais_dynamic_2017, @broda_low-dimensional_2012] using the Dupuit-Forcheimer approximation. Two-dimensional implementations of the Boussinesq model are common, but do not appear to be widely available as open source packages. The simplicity and computational efficiency of this method is advantageous for capturing the first-order effects of groundwater flow on other Earth surface processes. More sophisticated groundwater models may be necessary depending on the hydrological features that the user intends to capture.
```
kbarnhart: Some small thoughts on content and clarity:
* "as in GSFLOW" -> "as is implemented in GSFLOW"?
* I might define parsimonious (or use the words that define it instead of that term). My sense is that different communities have different ideas about what a parsimonious model is and how one is identified. 
* You nicely define the Richards equation and the D-F approximation. I'd recommend adding a similar statement of what exactly the Boussinesq equations are. I'd keep your statement that the Boussinesq models are implemented for gw flow in 2D hill slopes, but add what they are. You might even introduce this at the end of the prior paragraph, after the statement "free surface with groundwater flow driven by gradients in water table elevation".

joss/paper.md
```diff
@@ -35,7 +35,9 @@ bibliography: papers.bib
...
+
+Implementations of the Boussinesq model often encounter numerical instabilities as the water table intersects the surface and groundwater return flow occurs through a seepage face, due to the presence of a discontinuity in the energy gradient from inside the hillslope (where it is determined by the water table) to the seepage face (where it is determined by the topography). @marcais_dynamic_2017 introduced a regularization that smooths the transition between subsurface flow and surface flow. Although introduced for numerical stability and not based on physical principles, this smoothing may reproduce the effect of subgrid heterogeneity where saturation within a grid cell is unlikely to be homogenous or well-reproduced by a binary condition (saturated vs unsaturated).
```
kbarnhart: * Should "instabilities as the" be "instabilities where" or "instabilities when". I'm not entirely certain, so you should keep it as is if you feel strongly. 
* I would split the first sentence into two changing ", due" to ". This is due"

DavidLitwin: Addressed in most recent commit!

joss/paper.md
DavidLitwin: Addressed in most recent commit! Thanks :)

kbarnhart:
This only failed because one of the six travis tests on push timed out. I'm merging.

kbarnhart:

Issue #1106, opened 2020-01-15
JOSS review groundwater component: Tutorial


rreinecke:
Can you please add a step that verifies whether the mass balance is conserved also with varying step sizes and recharge rates?

What about different boundary conditions? Could you at least summarize what else is possible?

kbarnhart:
[Linking to the JOSS Review Issue](https://github.com/openjournals/joss-reviews/issues/1935)

DavidLitwin:
Yes, can do. I think that all of these are good ideas. It won't happen today but I will work on it when I'm able.

DavidLitwin:
Referencing pull request #1123 Thank you for suggesting this, I think it adds a lot to the tutorial!

PR #1123, opened 2020-01-23
Update groundwater tutorial for #1106

DavidLitwin:
Add tests for conservation of mass with changing timestep and recharge rate, and add detail on setting fixed gradient boundaries. Also corrected the initial storage term in the mass conservation plots.


kbarnhart:
@DavidLitwin there were a few places where I felt like a small amount of additional explanatory text would benefit the tutorial. I've made notes with markdown cells that say:

TODO: ADD SOME EXPLANATORY TEXT HERE

This seemed easier to add on a commit rather than as comments.

kbarnhart:
The explanatory text added a lot. Thanks @DavidLitwin !

kbarnhart:

Issue #1118, opened 2020-01-20
Groundwater module error


rclavera:
Dear Landlab team,

I was testing the groundwater component. The example used have the next magnitudes:
number of rows:        2761
number of columns:  2528
number of nodes:      6979808
number of cells:         6969234
number of core cells: 6969234
area of a cell:       1.0

This error appears running the example:
calc_recharge_flux_in
    return np.sum(self._grid.area_of_cell * self._recharge[self._cores])
ValueError: operands could not be broadcast together with shapes (6969234,) (4946676,)

I understand that 
“self._grid.area_of_cell” is an array of 6969234 (cells)
“self._recharge” is an array of 6979808 (nodes)
“self._recharge[self._cores]” is an array of 4946676 (core nodes)

What can I do?
Thank you very mcuh

kbarnhart:
@rclavera thank you for making this issue. With the information you have provided it is really difficult to assess the issue. 

Could you please provide a code-snippet that contains all of the code necessary to reproduce your error (e.g., small model grid, all synthetic values, no input files). This will make it possible to reproduce, diagnose, assess the error you've found. 

tagging @DavidLitwin

rclavera:
@kbarnhart ,
Here I attached the code
[test.py.zip](https://github.com/landlab/landlab/files/4089573/test.py.zip)
sorry, but the other file (compressed) occupies 35.7 MB. What can I do?
Thank you

rclavera:
@kbarnhart thanks, 
I'm working on it reducing the image .
Is not Landlab v2 available in PYPI?

kbarnhart:
@rclavera Landlab v2 is on pypi but you need to specify the `--pre` flag. 

```
pip install --upgrade --pre landlab
```

should work to upgrade. Let me know if it doesn't work for you. Note that we presently don't have python 3.8 builds on either conda-forge or pypi. we should have them in the next few days. 

I've now created a PR (#1119) to fix the [update instructions](https://landlab.readthedocs.io/en/master/install/index.html#id2)

rclavera:
@kbarnhart, Landlab v2 has been installed. Thanks

rclavera:
I attach the files to test the problem I have.
[Archive 2.zip](https://github.com/landlab/landlab/files/4101473/Archive.2.zip)

kbarnhart:
@DavidLitwin could you please have a look at this?

DavidLitwin:
[...]

PR #1124, opened 2020-01-24
Address #1118

DavidLitwin:
Small change to gw component flux calculations, correctly indexing area_of_cell.


kbarnhart:
@DavidLitwin I think part of the problem is that area_of_cell is a number of cells long array, so indexing is tricky if there are inactive cells. I switched to using area_of_cell_at_node, so it can be indexed correctly be self._cores. 

Could you add a test of this part of the code that has a grid with at least one interior node which is inactive (so that grid.node_at_cell is not the same as grid.core_nodes)

DavidLitwin:
@kbarnhart Ah, I didn't know that method existed. Looks good though! I'll see what I can do as far as a test.

kbarnhart:

DavidLitwin:
@kbarnhart I'm getting an error with this method.

`AttributeError: 'RasterModelGrid' object has no attribute 'area_of_cell_at_node'`

Am I missing something?

kbarnhart:
you are right... I mistyped. it is cell_area_at_node

tests/components/groundwater/test_dupuit_percolator.py
```diff
@@ -219,3 +219,31 @@ def test_wt_above_surface_adaptive_run_step():
     gdp = GroundwaterDupuitPercolator(grid, recharge_rate=0.0)
     gdp.run_with_adaptive_time_step_solver(1)
     assert_equal(wt[4], 1)
+
+
+def test_inactive_interior_node():
```
kbarnhart: looks good. thanks @DavidLitwin

kbarnhart:
Addressed #1118

Issue #1125, opened 2020-01-24
JOSS Review groundwater component: Install error with depricated python version


rreinecke:
With Python 2.7.17 and `python setup.py develop
` you get an error message:
`Traceback (most recent call last):
  File "setup.py", line 12, in <module>
    import versioneer
  File "../landlab/versioneer.py", line 1719
    file=sys.stderr)
        ^
SyntaxError: invalid syntax
` I'm aware that this version is deprecated with this year but it is still pretty common on many systems.

kbarnhart:
@rreinecke Landlab no longer supports python 2.7. Is there somewhere you think we should more clearly communicate this?

rreinecke:
> @rreinecke Landlab no longer supports python 2.7. Is there somewhere you think we should more clearly communicate this?

Yes. As I already suggested in the setup.py and a small note in the readme

rreinecke:
Is it possible to run the tests and build landlab without using conda? I personally really dislike it and would appreciate it if I do not need to install it on my machine.

kbarnhart:
Forgive me if I missed something, but I didn't see where you recommended mentioning this in the `setup.py` and the `readme.rst`. We already list the python versions in the `setup.py` file as classifiers, but I'll address this by:
* adding `python_requires` and
* adding a note about supported versions in the readme. 

Regarding building landlab and running the tests without conda: I expect that if you install all of the dependencies by whatever means you like (e.g., all as binaries, all from source, some mixture), you should be able to install landlab from source without using conda. This approach, however, is not something that we support.

kbarnhart:
[Linking to the JOSS Review Issue](https://github.com/openjournals/joss-reviews/issues/1935)

rreinecke:
@kbarnhart I just noticed it now when trying to run the tests on my machine. As soon as #1126 is closed this can be closed as well.

mcflugen:
@rreinecke 

> Is it possible to run the tests and build landlab without using conda? I personally really dislike it and would appreciate it if I do not need to install it on my machine.

It is, and, in fact, our CI is run using *pip* to install the dependencies, not *conda*.

PR #1126, opened 2020-01-24
Improve clarity of dropping Py27 support

kbarnhart:
Addresses comments raised in #1125.

@mcflugen might you take a look.  I'm not sure if there are additional things we need to test in adding `python_requires`.


kbarnhart:

kbarnhart:
@mcflugen 🏓

PR #1127, opened 2020-01-24
Fix DepthDependentDiffuser and DepthDependentTaylorDiffuser docstrings

kbarnhart:
It was recently pointed out that the DepthDependentDiffuser and DepthDependentTaylorDiffuser docstrings reflect incorrectly the equations used by these components. 

This PR fixes this. 

It also cleans up the init to use the initialized_output_fields function.

### DIFF
--- a/landlab/components/depth_dependent_diffusion/hillslope_depth_dependent_linear_flux.py
+++ b/landlab/components/depth_dependent_diffusion/hillslope_depth_dependent_linear_flux.py
@@ -19,7 +19,7 @@ class DepthDependentDiffuser(Component):
 
     .. math::
 
-        q_s = D S H^* (1.0 - exp( H / H^*)
+        q_s = - D S H^* (1.0 - exp( - H / H^*)
 
     where :math:`D` is is the diffusivity, :math:`S` is the slope, :math:`H` is
     the soil depth on links, and :math:`H^*` is the soil transport decay depth.
--- a/landlab/components/depth_dependent_diffusion/hillslope_depth_dependent_linear_flux.py
+++ b/landlab/components/depth_dependent_diffusion/hillslope_depth_dependent_linear_flux.py
@@ -199,29 +199,16 @@ class DepthDependentDiffuser(Component):
         self._K = linear_diffusivity
         self._soil_transport_decay_depth = soil_transport_decay_depth
 
-        # create fields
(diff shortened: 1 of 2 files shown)


landlab/components/depth_dependent_diffusion/hillslope_depth_dependent_linear_flux.py
```diff
@@ -19,7 +19,7 @@ class DepthDependentDiffuser(Component):
 
     .. math::
 
-        q_s = D S H^* (1.0 - exp( H / H^*)
+        q_s = - D S H^* (1.0 - exp( - H / H^*)
```
gregtucker: Not sure about the leading minus. If S = dz/dx, then it's needed, but if S is positive downward, i.e. S = -dz/dx, then I think it should not be there.

landlab/components/depth_dependent_taylor_soil_creep/hillslope_depth_dependent_taylor_flux.py
```diff
@@ -24,7 +24,7 @@ class DepthDependentTaylorDiffuser(Component):
 
     .. math::
 
-        q_s = DSH^* ( 1 + (S/S_c)^2 + (S/Sc_)^4 + .. + (S/S_c)^2(n-1) ) (1.0 - exp( H / H^*)
+        q_s = - DSH^* ( 1 + (S/S_c)^2 + (S/Sc_)^4 + .. + (S/S_c)^2(n-1) ) (1.0 - exp( - H / H^*)
```
gregtucker: Same as above

landlab/components/depth_dependent_diffusion/hillslope_depth_dependent_linear_flux.py
kbarnhart: In the actual code we calculate as:

```python
        # Calculate flux
        self._flux[:] = (
            -self._K
            * slope
            * self._soil_transport_decay_depth
            * (1.0 - np.exp(-H_link / self._soil_transport_decay_depth))
        )
```

implying S = dzdx. Probably worth explicitly stating this to avoid confusion. Will add.

kbarnhart:
only fail was a timeout on Travis CI push, so I'm going to merge.

PR #1128, opened 2020-01-28
StreamPowerSmoothThresholdEroder indexing bug

kbarnhart:
@gregtucker I found this while updating terrainbento to LL2. 

self._thresholds is an at-node array, so it needs to be indexed here.


kbarnhart:
a few other comments:
1) not sure why this came up as an issue now and not with the prior versions of terrainbento
2) there should be a test.

PR #1130, opened 2020-01-28
feat: add setters for erodibility for some stream power like components

kbarnhart:
This PR adds setters for the fluvial erodibility coefficient in 
- FastscapeStreamPower (and by extension StreamPowerSmoothThresholdEroder)
- StreamPowerEroder
- ErosionDeposition
- Space

These are necessary to support updating terrainbento to be compliant with LL2.


kbarnhart:
@mcflugen once this is merged, I'd like to make a new release of landlab (v2.0.0b4). Is that OK with you?

Issue #1069, opened 2019-12-04
Groundwater component anomalies


rclavera:
Hi everybody,
I've been tested the groundwater component. First at all, I want to congratulate you for the incredibly good job.
On the other hand, I would like to comment some issues:
I run the next example (code simplified):

[groundwaterExample.py.zip](https://github.com/landlab/landlab/files/3921944/groundwaterExample.py.zip)

1.- In the results we can appreciate an anomaly in the water table (see the attached video ‘water_table.ogv’
[water_table.ogv.zip](https://github.com/landlab/landlab/files/3921922/water_table.ogv.zip)
): it is really strange and I suppose it is a numerical instability.
In other examples with lower thickness (difference between water table and aquifer base)  does not appear
2.- Does the component consider that the seepage only occurs in the line created by water table surface and topographic elevation surface?
![surface_water360](https://user-images.githubusercontent.com/47635648/70145132-b164b600-169f-11ea-9ef7-dd9915427d41.png)
3.- What are the black horizontal lines near the seepage of the previous image?

kbarnhart:
Thanks for the comment and providing the example. I’ve tagged @DavidLitwin , the primary author of the component.

DavidLitwin:
Hi @rclavera, thank you for your contribution to testing this component. I'll do my best to answer your questions here.

1. 
 Just to be clear, the instability feature that you are seeing is this one, correct?
![image](https://user-images.githubusercontent.com/45630470/70253834-36ec7280-1741-11ea-8b77-7e9267f66abf.png)
The figure above shows surface water specific discharge, from the test case that you included. I have seen something like this early in the process of development, but I thought it had been resolved (albeit indirectly, through a variety of changes made for stability). As of right now, I don't have a solution for this, but I would very much like to. If you have any suggestions, I am happy to hear. One thing that I can say is that it does seem to appear when hydraulic gradient is very small (aquifer base is flat, water table is flat or nearly flat). This may provide some insight when looking for a solution. 

2. 
I am interpreting your second question as whether the component only considers seepage at the line of intersection between water table and topographic elevation. To get a better sense of what is going on here, I would look at the equation for surface water seepage:
![image](https://user-images.githubusercontent.com/45630470/70255068-63a18980-1743-11ea-98ed-f9be843acaea.png)

[...]

PR #1131, opened 2020-01-28
Address stability issues in GroundwaterDupuitPercolator

kbarnhart:
Address stability issues identified by @rclavera in #1069.


landlab/components/groundwater/dupuit_percolator.py
```diff
@@ -269,7 +269,8 @@ def __init__(
         porosity=0.2,
         recharge_rate=1.0e-8,
         regularization_f=1e-2,
-        courant_coefficient=0.01,
+        courant_coefficient=0.5,
+        vn_coefficient = 0.8,
     ):
```
kbarnhart: I think you will need to add something to the docstring indicating what this is.

landlab/components/groundwater/dupuit_percolator.py
```diff
@@ -334,6 +335,7 @@ def __init__(
 
         # save courant_coefficient (and test)
         self._courant_coefficient = courant_coefficient
+        self._vn_coefficient = vn_coefficient
 
     @property
     def courant_coefficient(self):
```
kbarnhart: I might add a similar property for the VN condition.

kbarnhart:
@DavidLitwin I think that this approach makes sense. I've made a couple of comments in the code.

DavidLitwin:
@kbarnhart I've made a few changes wrt your comments and a few other things that came up as I was looking at the code. One thing to note - I dropped the cos(theta) from the stability definition, because to arrive at the von Neumann criteria, I assume advection is negligible, in which case cos(theta)~1.

landlab/components/groundwater/dupuit_percolator.py
```diff
@@ -294,6 +295,15 @@ def __init__(
...
             zero.
+            Default = 0.5
+        vn_coefficient: float (-)
+            The multiplying factor C for the condition dt >= C*dx^2/(4D),
+            where D = Kh/n is the diffusivity of the Boussinesq
+            equation. This arises from a von Neumann stability analysis of
+            the Boussinesq equation when the hydraulic gradient is small.
+            This parameter is only used with ``run_with_adaptive_time_step_solver``
+            and must be greater than zero.
+            Default = 0.8
```
kbarnhart: Great!

landlab/components/groundwater/dupuit_percolator.py
```diff
@@ -356,6 +367,31 @@ def courant_coefficient(self, new_val):
...
+        return self._vn_coefficient
+
+    @vn_coefficient.setter
+    def vn_coefficient(self,new_val):
+        """ set coefficient for the diffusive timestep condition in
+        the adaptive timestep solver. """
+        if new_val <= 0:
+            raise ValueError("courant_coefficient must be > 0.")
+        self._vn_coefficient = new_val
+
```
kbarnhart: Great!

landlab/components/groundwater/dupuit_percolator.py
```diff
@@ -334,6 +344,7 @@ def __init__(
 
         # save courant_coefficient (and test)
         self._courant_coefficient = courant_coefficient
+        self._vn_coefficient = vn_coefficient
```
kbarnhart: Very minor comment, and something that I recently learned... 

If you do 


```python
self.courant_coefficient = courant_coefficient
self.vn_coefficient = vn_coefficient
```
(removing the leading underscores) then you invoke the setters and use the checking you've built into them.

DavidLitwin: Interesting, should I do that? And if I do, do I remove leading underscores on all appearances of those variables?

kbarnhart: I would recommend you do. 

When you do this, the assignment on these lines will invoke the setter, which will check the value provided and then assign the value to the private variable`self._courant...` (because that is what the setter does. It is a way to ensure that the checking created in the setter is used in the init. 

Regarding whether there is a difference between using `self._name` of var or `self.name_of_var` (which points to `self._name_of_var `through the `@property`) internally... this is beyond my python knowledge. I might expect that using `self.name_of_var` might be marginally slower. tagging @mcflugen if he has any thoughts that we can both learn from.

DavidLitwin: ok, made some changes to use setters in more places in init. Let me know what you think.

kbarnhart: looks good to me.

kbarnhart:
@DavidLitwin a v2.0.0b4 release is imminent. do you want this to be part of it?

DavidLitwin:
@kbarnhart, if it's ok, I would prefer to hold off until I get some feedback from someone who knows the numerical methods better than I do. Thank you for checking though! I expect I should have a better idea what I want to do by the end of next week.

DavidLitwin:
@kbarnhart, I'm ready to go ahead with this. I checked about the new stability criteria and I think it is the correct thing to do. I've updated some docs as well for clarity, and removed the shear stress method as discussed with you and @gregtucker. Now the component does not have dimensions!

kbarnhart:
@DavidLitwin presuming this passes, I think its good to go. I did the following:
- addressed lint
- turned unit agnostic to True

nice work!

Issue #1100, opened 2020-01-13
Remove cruft and organization


kbarnhart:
We discussed removing cruft on the the call this morning. I've compiled some of the places we might look/remove. We can discuss each of these on this issue

Looks clear to me it should be removed:
- [x] landlab/testing (these are things that should be replaced with pytest or numpy testing?)

Already identified:
- [x] duplicate field creation (issue #1058)

Uncertain:
- [x] landlab/io/gebco 
- [x] landlab/io/vtk (its clear that having this would be good, but it is not done, I gather)
- [x] ~landlab/plot/event_handler.py (this and the next are very old and seem to me to be unused)~ KEEP
- [ ] landlab/plot/video_out.py 

I didn't look at (assigned to @mcflugen):
- [x] landlab/grid
- [x] landlab/graph
- [x] landlab/field

Other things I noticed:
 - [ ] Should landlab/framework get a new name or be moved to landlab/bmi. it seems everything in there has to do with bmi. 
- [ ] Its not clear why some things are in landlab/core/utils.py and some things are in landlab/utils... perhaps the distinction should be clearer. 

:wave: @landlab/core

nathanlyons:
`event_handler` still works and is used to click on imshow_grid plots (eg in spyder, interactive jupyter plots) to get the node info where clicked.

kbarnhart:
@nathanlyons thank you! I have removed it from the above list.

nathanlyons:
Could the `animate_landlab_output` tutorial be the recommended method/replacement to `video_out`? Something like the latter is nice to make quick video, although not being able to animate grids and deep customization of plot elements is limited.

nathanlyons:
Regarding distinctions of utilities, I think especially the user audience (sensu Barnhart et al, in progress) could benefit from having a groupings of 

- things that fall under earth surface dynamics (eg, `fault_fact_finder`, `flow_distance`, and `watershed`) and
- things associated with programming tasks, objects, and data structures (eg, `add_halo`, `count_repeats`, `decorators`).

The programmy things could live in landlab/core/utils/ along with `utils.py`. The earth surface dynamics things in their current location.

Maybe some things will be hard to fit in these designations (eg `source_tracking_algorithm`, maybe bad example) in which case if there is not a clear reason, I advocate to leave with the earth surface dynamics things.

kbarnhart:
Three of four uses of the 
```
[...]

PR #1149, opened 2020-02-10
Remove cruft from landlab.grid.structured_quad

mcflugen:
In support of #1100, this pull request removes cruft from the *landlab.grid.structured_quad* subpackage. Much of the removed code is duplicated in the graphs subpackage and the remaining code (*landlab.grid.structured_quad.links*) should probably be moved there as well.


kbarnhart:
@mcflugen I'm going to let you handle updating/merging this one. But I think it is good.

Issue #1104, opened 2020-01-14
Unclear which components require specific units


vaga3461:
I've created a grid and specified that I want topographic elevation to be in kilometers, and now I'm using FastscapeEroder on that grid. "FastscapeEroder.units" says that FastscapeEroder elevations are in meters. It's unclear to me if I can use this component on a grid with different units.

nicgaspar:
Landlab does not explicitly handle units. So you can have elevation in your units of choice, just make sure that you are being careful about unit conversion. That is, drainage area also has units, and those units are dependent on the units of dx and dy.

kbarnhart:
@vaga3461 and I discussed this a bit and based on that conversation I have some additional thoughts:

At present almost all components are unit agnostic (OverlandFlow, KinwaveImplicitOverlandFlow are two I can think of that are not, I know there are others). However, we indicate (for fields and input parameters), units instead of dimensions. 

We have discussed presenting dimensions instead of units. However, each time this has come up (I have been present for two versions of this conversation and know that more occurred before I joined the dev team), we end up deciding that the default values themselves have units, and thus we should use those units instead of dimensions. 

However, unless a user digs into the source code to determine whether a component truly is unit agnostic or not, it is not clear which landlab components are unit agnostic and which ones are not. 

I would propose that we consider adding a piece of metadata to components which indicates:

a. I am unit agnostic, just be consistent with your units and you'll be fine. 
b. I do actually care about my units and you must look at my units carefully to use me correctly. 

What do others think. 

Ultimately, I'm open to any solution which makes unit agnostic/not agnostic clearer to a user. 

:wave: @landlab/core

gregtucker:
I agree a unit required/agnostic indicator would be a useful addition to component metadata. Perhaps an optional header item: `_unit_agnostic = True`

kbarnhart:
On the call we've decided that it makes sense to 

* Add a `_unit_agnostic = True` to component base class, then add `_unit_agnostic = False` to those components for which it really matters. 
* Add public attribute to Component. Add information to info printed about component
* Add information to the Component tutorial. 

@landlab/core anyone want to volunteer to do this.

PR #1151, opened 2020-02-12
Address Units/Dimensions

kbarnhart:
This PR implements what was discussed on #1104. I've prioritized this because it is discussed in the paper (which I plan to submit by the end of the week). 

Specifically it:

* Add a _unit_agnostic = None to component base class, then add _unit_agnostic = False to those components for which it really matters, and _unit_agnostic = True where it doesn't matter.
* Add public attribute to Component. Add information to info printed about component
* Add information to the Component tutorial.
* Adds a page to the user guide which describes how landlab deals with units. 

Where I observed a unit conversion within a component (in the init/run/other calculation) I wrote unit_agnostic=False. Based on my perusal, the components for which this is the case are: 

GroundwaterDupuitPercolator
KinwaveImplicitOverlandFlow
KinwaveOverlandFlowModel
LandslideProbability
LateralEroder
OverlandFlow
OverlandFlowBates
PotentialEvapotranspiration
PotentialityFlowRouter
Radiation
SedDepEroder
SoilMoisture
SoilInfiltrationGreenAmpt
SpatialPrecipitationDistribution
VegCA
Vegetation

If others could please look at these components and indicate whether they agree/disagree that would be much appreciated. 

I also observed something that is (sorta) troubling... There are cases in which the only reason a component is unit agnostic is that there is an inconsistency between units of different inputs (this is the case in the KinwaveOverlandFlowModel component highlighted in the "making components" tutorial.... This occurs because Mannings N is specified as s/m^(1/3) but precipitation and infiltration rates are specified with time units of hours...

I would argue that we should look for cases like these and remove all of them (but I'm probably over reacting). 

The other main reason that a component is non-unit-agnostic is that there is a non-user-defined coefficient that has units.

tagging @landlab/core


saisiddu:
@kbarnhart: I agree with PotentialEvapotranspiration, Radiation, SoilMoisture,  VegCA, and Vegetation. I am pretty sure LandslideProbability is also NOT unit agnostic (agreeing with your decision). Thank you!

DavidLitwin:
@kbarnhart the shear stress method in the groundwater component may be the only thing that's unit  dependent (SI), because it relies on Manning's equation. For everything else, while I've given default values in SI (m and s), there's no reason why it couldn't be in ft or hr, so long as units were consistent (elevations, grid spacing, hydraulic conductivity and recharge).

DavidLitwin:
Also @kbarnhart, this may be up for debate, but I'm not sure the shear stress calculation even belongs in gdp? It was useful to me at one point, but it (a) does not depend on flow being created with the groundwater component (b) depends on a running a FlowAccumulator.

nathanlyons:
@kbarnhart I did not find a unit agnosticism that I disagreed with.

PR #1155, opened 2020-02-22
Add layer reducer

mcflugen:
This pull request adds a new method, *reduce*, to *EventLayers*. The *reduce* method combines blocks of *EventLayers* into single blocks. So, for instance, if there are 200 layers, calling *reduce* could reduce the number of layers to 20 but averaging every 10 layers into a single layer. This can be useful when running a model that generates a large number of layers but only the most recently deposited layers are of importance and deeper layers can be archived to save memory.

As a demonstration of how this could be used, the following is a code snippet from the *Sequence* model,
```python
    if (self.grid.event_layers.number_of_layers - self._n_archived_layers) % 20 == 0:
        self.grid.event_layers.reduce(
            self._n_archived_layers,
            self._n_archived_layers + 10,
            age=np.max,
            percent_sand=np.mean,
            porosity=np.mean,
            t0=np.sum,
            water_depth=np.mean,
        )
        self._n_archived_layers += 1
```
In this example, after 20 new layers have been added layers 10 through 19 (as measured from the surface layer) are average into a single layer and archived. This ensures that there are always 10 layers at the top that are at full resolution. Notice the way that layer attributes are combined can be specified per attribute (e.g. the maximum of *age* is used, while *porosity* is averaged).

### DIFF
--- a/landlab/layers/eventlayers.py
+++ b/landlab/layers/eventlayers.py
@@ -876,9 +876,14 @@ class EventLayers:
         array([[ 1.5,  1.5,  1.5],
                [ 1. ,  2. ,  0.5]])
 
+        Combine all of the layers into a single layer.
+
         >>> layers.reduce()
         >>> layers.dz
         array([[ 2.5,  3.5,  2. ]])
+
+        Add two additional layers to the top. The bottom-most layer is row
+        0, and the two new layers are rows 1 and 2.
 
         >>> layers.add([1., 2., .5])
         >>> layers.add([1., 2., .5])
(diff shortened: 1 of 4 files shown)


landlab/layers/eventlayers.py
```diff
@@ -661,16 +869,147 @@ def add(self, dz, **kwds):
...
+        >>> layers.dz
+        array([[ 2.5,  3.5,  2. ]])
+
+        >>> layers.add([1., 2., .5])
+        >>> layers.add([1., 2., .5])
+        >>> layers.dz
+        array([[ 2.5,  3.5,  2. ],
+               [ 1. ,  2. ,  0.5],
+               [ 1. ,  2. ,  0.5]])
+        >>> layers.reduce(1, 3)
```
gregtucker: Suggest adding a comment pointing out that (if I'm right) this combines the top 2 layers, which are 1 and 2, because layers are numbered upward from bottom (so the deepest layer is 0).

mcflugen:
@gregtucker Thanks! I haven't added anything to the notebooks that demonstrate how to use layers, but I will as part of this pull request.

Issue #1138, opened 2020-02-04
Assigning node fields to NMG in read_shapefile


pfeiffea:
@mcflugen and @kbarnhart : 

I’m trying to use a shapefile to create an NMG to run NST. Issue I’m having: we need to assign topographic elevations for each node on the network. The way read_shapefile works, there isn’t an obvious way to set node elevations. Furthermore, if I just had a list of node elevations and I wanted to somehow bring them in, it gets complicated—the indexing of input links/nodes doesn’t match landlab link/node indexing. 

Jon alluded to this issue in a conversation the other day, and I’m coming to appreciate that this is a trickier issue than I’d understood. 

Is there an existing example of how to do this?

kbarnhart:
@pfeiffea I'll work on an answer to this on the bus tonight.

kbarnhart:
@pfeiffea I'm going to need to tap @mcflugen in here for some help. A couple of thoughts

read shape file reads a poly line type shape file, creates a network grid, and assigns any fields located on the shape file to be located _at link_. So it is unclear to me where in the shape file itself the elevation location would be stored (or do the point geometries also have z attributes?) Could you clarify on this point?

One simple approach to try, is that if you have x and y and z locations that _you know_ align with the grid nodes, then you can sort them so that the increase by x and then by y (which I think means sort first by y and then by x, but @mcflugen will correct me if I'm wrong). np.argsort and np.take_along are the two functions you'll want to check out. 

The issue you are facing with the ordering is that landlab's grid will re-order the nodes that make up the link endpoints and links based on the landlab ordering scheme. I'm pretty sure that this means that the code I wrote originally ([link to the line here](https://github.com/landlab/landlab/blob/7b15af7fb877bbf91d24c77d236863fe6a89d1b1/landlab/io/shapefile/read_shapefile.py#L171)) to assign at-link attributes to the grid is wrong and these values need to be re-ordered so that they are assigned to the correct links.

 @mcflugen could you point me in the right direction for re-ordering these values so that they can be assigned to the grid.

mcflugen:
@kbarnhart, @pfeiffea 

There is a *sort* keyword for *NetworkGraph* that controls if the graph elements should be sorted (*landlab*-style) or not. The way things are now, *NetworkModelGrid* always sets this value to *True* (i.e. elements are always sorted).

Would one possibility be to set this value to *False* so that nothing is reordered?

[...]

PR #1158, opened 2020-03-03
Barnhark/improve read shapefile

kbarnhart:
Currently in progress, but eventually this will address @pfeiffea's #1138


kbarnhart:
@pfeiffea I now have added all the capabilities we discussed over on #1138. It works on the example test case, but doesn't work on the example shape file you sent... I was looking at:

MethowSubBasin.shp and MethowSubBasin_Nodes_4.shp (correct?)

MethowSubBasin.shp has 29 links (so 30 nodes expected in MethowSubBasin_Nodes_4.shp). 
But MethowSubBasin_Nodes_4 had 59 points. I tried to delete obvious duplicates... and failed.

pfeiffea:
Hi @kbarnhart I just made the fix to the nodes shapefile (very odd... I wonder how Jon was processing those in his Matlab version...) and pushed the changes. That should fix the issue.

kbarnhart:
@pfeiffea I've now completed this. I added a few other things that I thought would be very useful. Specifically:

1) A way to indicate that only a subset of shape file attributes (for both node and link shape files) are imported as fields. 
2) A way to change the name of the shape file attributes to a new name on import. 
3) A way to indicate that you really want a shape file attribute being imported to have a specific data type. 

Hopefully you think these things are useful! 

Once this passes, I'll ask @mcflugen  to review.

pfeiffea:
amazing @kbarnhart, this looks perfect.

kbarnhart:
@pfeiffea glad you think so. 

As I was writing the tests, those were the three features that I felt would be useful in actually using this for application.

kbarnhart:
@mcflugen any chance you can look at this?

kbarnhart:
@mcflugen a quick ping to see if you have a moment to look at this. If there are more pressing things, do them

PR #1161, opened 2020-03-04
Resolve memory leak, fix resulting tests

### LINKED ISSUE #1156: Memory leak in LakeMapperBarnes
I'm running the lake mapper in a landscape evolution model in order to route flow through depressions, particularly those that occur because of the randomized initial condition. I found that the HPC node I am using was consistently running out of memory, and I have narrowed down the source of this memory use increase to the lake mapper. 

Attached is a minimal reproducible example, with a plot showing how memory usage increases linearly with iteration in a for loop. Any support on this would be helpful. I would like to get this up and running soon.

[lake_mapper_mem.zip](https://github.com/landlab/landlab/files/4256769/lake_mapper_mem.zip)

SiccarPoint:
For very dynamic length objects (e.g., heaps), allocation of the data structure during __init__ results in memory leaks. This PR resolves this issue for the LakeMapperBarnes by moving the object declaration down into the scope of run_one_step.

This resolves #1156


DavidLitwin:
@kbarnhart and @gregtucker I went ahead and merged these changes, given that all tests passed. I realize now that I'm not sure I officially have permission (from Landlab folks) to be merging into Master... I'll make sure to get approval if this should happen again.

gregtucker:
@DavidLitwin the main requirement is to get review and approval from at least 1 other person, which you did for Dan, so that's fine. Good on you both for figuring this out.

Issue #1176, opened 2020-04-03
Updating flow accumulation without updating flow directions


DavidLitwin:
When working with the groundwater model, there are circumstances where the hydrological state may be updated much more frequently than the geomorphology. Because of that, it would be efficient to separate updating flow direction from updating accumulated flow. 

From a brief email exchange with @kbarnhart, it seems like the best way to do this is to use `FlowAccumulator.accumulate_flow(self, update_flow_director=False)`, but there is a lack of clarity in the documentation about how to do this in combination with a flow director that's already instantiated. I'll attach some simplified code to give you an idea of what I'm after. I'm not yet sure whether this is doing what I think it should be, as noted in the inline comments.

[separate_flow_direction_accum.zip](https://github.com/landlab/landlab/files/4428541/separate_flow_direction_accum.zip)

kbarnhart:
@DavidLitwin what you have written should accomplish what you want. There is something about your code that doesn't make sense to me, and I've noted it below. 

I'll take a stab on improving the docstring, link to it here, and ask for your recommendations on it. 
In looking at this, I've also noticed that IF a depression finder is passed (as is not the case in your example), you can turn off recalculating flow directions, but you can't turn off recalculating depressions. So I'll add the ability to do that too. 

```python
import numpy as np
from landlab import RasterModelGrid
from landlab.components import (
    GroundwaterDupuitPercolator,
    FlowDirectorD8,
    FlowAccumulator,
    FastscapeEroder,
    LakeMapperBarnes,
)

# create grid, sest boundary conditions
grid = RasterModelGrid((10, 10), xy_spacing=10.0)
grid.set_closed_boundaries_at_grid_edges(True, True, False, True)

# add fields.
elev = grid.add_zeros("topographic__elevation", at="node")
grid.add_zeros("aquifer_base__elevation", at="node")
elev[:] = 1 + 0.1 * np.random.rand(len(elev))
wt = grid.add_zeros("water_table__elevation", at="node")
wt[:] = elev

# initialize components
gdp = GroundwaterDupuitPercolator(grid)
fd = FlowDirectorD8(grid)
fa = FlowAccumulator(
    grid, flow_director=fd, runoff_rate="surface_water__specific_discharge"
)
lm = LakeMapperBarnes(grid)
sp = FastscapeEroder(grid, discharge_field="surface_water__discharge")

# run geomorph
for i in range(100):

    # find flow directions on updated topography
    fd.run_one_step()

    # route flow through depressions in topography
[...]

PR #1178, opened 2020-04-03
Improve FlowAccumulator docstrings

kbarnhart:
@DavidLitwin asked a question on #1176 that identified that the docstrings on the FlowAccumulator method accumulate_flow didn't actually describe what that function did. So I improved the docstrings. I also:
- Added the ability to turn off updating depression finding. 
- Removed a false statement from the documentation of LossyFlowAccumulator.


kbarnhart:
@mcflugen could you take a look at this. 

The docs stage is currently failing with an error associated with GraphFields. I can't reproduce it locally... Tagging you since you are likely to be able to decide how GraphFields should be represented in the docs. 

```
Warning, treated as error:
896/home/travis/build/landlab/landlab/landlab/field/graph_field.py:docstring of landlab.field.graph_field.GraphFields.groups:1:duplicate object description of landlab.field.graph_field.GraphFields.groups, other instance in reference/field/index, use :noindex: for one of them
897Makefile:22: recipe for target 'html' failed
898make: *** [html] Error 2
899make: Leaving directory '/home/travis/build/landlab/landlab/docs'
900The command "make -C docs clean html linkcheck" exited with 2.
```

kbarnhart:
@mcflugen do you understand why all the builds are failing with 
```
conda: command not found
The command "conda config --set always_yes yes --set changeps1 no" failed and exited with 127 during .
```

PR #1182, opened 2020-04-20
Remove cruft from field subpackage

mcflugen:
The pull request removes cruft from *landlab.field*. In particular, I've removed the following classes (all of which were unused):
* *ModelDataFieldsMixIn*
* *ModelDataFields*
* *ScalarDataFields*


kbarnhart:
Other than failing tests on appveyor this looks good to me.

mcflugen:
@kbarnhart Yeah, those *AppVeyor* builds are a bummer. I'm not sure what's up with them.

mcflugen:
@kbarnhart I know what the issue is and am trying to fix it on another branch. The problem is that we should be using Python 3 semantics for compiling cython extensions.

PR #1180, opened 2020-04-06
landslide component updated for V2

ChristinaB:
added feature of running landslide component with depth to water grids for hydrologic forcing as an alternative to recharge

PR #1183, opened 2020-04-20
Fix URL for miniconda download

mcflugen:
The URL to the *Miniconda* download changed a couple weeks ago, which caused our CI on Travis to fail (since we use *Miniconda* to install Python). This pull request fixes the URL.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,9 +72,9 @@ before_install:
   echo "backend: Agg" > $HOME/.matplotlib/matplotlibrc
 - |
   if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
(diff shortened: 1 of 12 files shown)


mcflugen:
This should fix tests that error-out on current pull requests (#1180, for example @ChristinaB)

mcflugen:
I've also added fixes for the *AppVeyor* builds. The problem was that newer versions of cython default to a language version of 3. This required changing *pyx* files to use:
* using integer division
* relative imports
* x.shape[0] instead of x.size to get the length of 1d arrays

I've also change the sphinx version requirement to be less than 3. This was necessary because sphinx version 3 produced warnings not present in version 2. I'll fix the warnings in another pull request so that we can use a newer version on sphinx.

mcflugen:
@kbarnhart It looks like your merge from master only changed some documentation so I'm going to merge this before the tests finish running again.

Issue #1177, opened 2020-04-03
ErosionDeposition/Space mass conservation and interior nodes that drain to themselves


kbarnhart:
Through careful and creative toy experiments @vaga3461 has identified that the way ErosionDeposition and Space handle sediment results in "evaporation" of sediment at ALL nodes that flow to themselves. This is reasonable behavior in only one circumstance: when that node is located at model grid edge and is an open boundary. 

Under other circumstances (e.g., closed boundary, bottom of an interior drained pit) the sediment should be deposited locally. 

There has been some discussion (mostly between @cmshobe @gregtucker and myself) about exactly how this should be handled. 

I am very very unlikely to be the person who actually fixes this, but since I've now had a few conversations about it I'm taking these notes for all our benefit (tag: @gregtucker @cmshobe @vaga3461 @BCampforts).

In the absence of a depression finder, there is no information about the size of "lakes" and thus the only real option for mass conservation is to dump all the sediment in the single pit node. I *think* however, that we will still lose any sediment that goes to fraction of fines. 

In the presence of a depression finder there are some options:
1) Option easiest is to handle this in the same way as without a depression finder. 
2) If a lake exists and has connectivity to an open boundary, how much sediment should get trapped in the lake (since we do route water and sediment through the lake). 
3) If a lake exists and it does not have any connectivity to an open boundary (e.g., a true Endorheic basin), how should sediment be distributed across it.?

A related, but distinct issue is #1172. A PR that addresses that should probably be merged before this issue is addressed.

SiccarPoint:
I tangled with this a little in the course of trying to get the SedDepEroder and also a lake fill algorithm going. The "just drop all the sediment" is extremely prone to a. inducing model instability (if it's just dropped); b. grindingly slow run times (if normal timestep limited procedures are followed; and c. ...the same thing proceeding to crashes, as there are a subset of combinations of sediment flux and topography that aren't solvable under b without introducing some rules like your 2 and 3 (e.g., more than one node supplying a single pit node with sediment, IIRC).

I don't have much specific advice, but to say I suspect fixing this is going to be no fun at all. Ironically, you may find that the easiest and most robust solution is a, i.e., just turn off the rate limiting and dump all the sediment. Or, retain the mass loss and note this in the documentation. 

[...]

Issue #1172, opened 2020-03-20
Erosion of flooded nodes in space


BCampforts:
Hello, 

I might be overlooking something but from what I see, I have the impression that flooded nodes are still eroded in the space.py component even when the erode_flooded_nodes flag is set to False and when using the run_one_step_basic definition. There seems to be a fix for this in the run_with_adaptive_time_step_solver definition on lines 625-627 but this does not apply for the solver with a fixed time step. A solution would be to remove the 625-627 lines and add them in the _calc_erosion_rates definition by passing 'flooded_nodes' as an argument and setting 
self._sed_erosion_term[flooded_nodes] = 0.0
self._br_erosion_term[flooded_nodes] = 0.0

gregtucker:
Vanessa and I unearthed a related issue in ErosionDeposition: nodes that are either flagged "flooded" by a lake/depression identifier, or are self-draining, *should* have a deposition rate equivalent to the incoming sediment flux, but do not. I hope to look at this and the above SPACE issue in the next day and a half.

BCampforts:
Regarding deposition in space.py: Charlie also noticed there should probably be a += on line  469 rather than an =

kbarnhart:
@gregtucker if I recall correctly you @cmshobe and I had a couple of conversations about what should happen when water enters a flooded node. If memory serves, we considered:
a) sediment is deposited at a rate of dHdt (some sediment is deposited some is passed downstream) and
b) all sediment is deposited. 

I think you favored (b), @cmshobe favored (a), and I wasn't certain which made more sense. 

I *think* what you are indicating is that it should be (b), right?

Whatever the case, I think that we should do the same thing to SPACE and ErosionDeposition. 
 
@BCampforts based on a brief assessment I think your diagnosis and approach are correct.

gregtucker:
@kbarnhart yes, that's right. This was prompted by a simple experiment that Vanessa ran: a 5x5 domain with all closed boundaries, cells in one row at an arbitrarily high elevation, the rest at zero. Mass should be conserved within the domain, but isn't if you use (a) above because there is currently no mechanism to pass sediment downstream if there is no downstream defined.

cmshobe:
Ohhh ok this is coming back now. I had forgotten about the flow routing part (i.e. lakes being their own receiver). Couple of thoughts:

[...]

PR #1186, opened 2020-04-24
Gt/ed space tweaks

gregtucker:
This PR changes how ErosionDeposition and Space handle flooded nodes. The previous erode_flooded_nodes option is removed. Functions to identify the "flood" status of nodes are now moved into GeneralizedErosionDeposition, and they handle two cases:
1 - a flood mapper is used, as indicated by the existence of a `flood_status_code` field: core nodes flagged as _FLOODED are tagged as True
2 - a flood mapper is not used (no `flood_status_code` field): nodes with slope <= 0 are tagged as True
For the latter case, both ErosionDeposition and Space will now deposit in any such nodes at a rate equal to the node's sediment influx rate (`_qs_in`) divided by cell area.
To the best of my knowledge, this should resolve issues #1177 and #1172.


kbarnhart:
@gregtucker I don't think this is quite right yet. I added some tests to demonstrate mass conservation, and at present they don't all pass. I'm going to take a break, so feel free to take a stab at this.

kbarnhart:
I have figured part of this out. porosity was being used to expand erosion AND deposition. haven't fixed everything but making some progress.

kbarnhart:
Here are the major things I needed to change/improve in order to conserve mass. 

### For Space:

1) the calculation of Es needs to be multiplied by (1-phi) so that the material going into the qs_in calculation is MASS (and thus there are no problems with distinguishing/mixing/unmixing material that is eroded from sediment and material eroded from bedrock). 

2) In the analytical solution for space, in two of the three cases outlined by Shobe equations (Eq 32, 34, and 35) there was an additional numerical instability. Specifically, if H<<H* then exp(H/H*) becomes infinite. This means that for both the calculations of R(t) (see pt 3.) and H(t) there are actually five cases, not three. Fortunately when exp(H/H*) is very large, the equations get easier to solve. I chose H/H*<100 as a threshold to for the cases based on where numpy gives overflow errors (somewhere between exp(100) and exp(1000)). I think anything greater than H/H*>10 probably is equivalent. 

3) In order to conserve bedrock correctly Equation 36 in Shobe et al (2017) needs to be integrated numerically to acknowledge that the rate of bedrock erosion will change within a tilmestep. This became apparent when I wrote tests for mass conservation that start with H=0. The results of dRdt would depend on whether dRdt was calculated with the initial sediment thickness of 0 (too much bedrock would be eroded)  or at the end of the timestep (not enough bedrock would be eroded). This was addressed by solving for R via a numerical integral. This presently happens in a loop in python. It can be cythonized at some point....

Less critical changes. I'd always had a very hard time following the analytical cases. I made them more understandable (for me). They are at least more commented, and connected with the paper equations. 

### For ErosionDeposition:

Mass conservation with phi~=0 is a tricky beast here. Unless it is assumed that ALL material has a porosity (which is basically like saying K is slightly higher) then I don't think you can demonstrate mass conservation beyond one timestep. Because we don't distinguish between materials, once material is is deposited (and poofed), subsequent erosion may mine through material with varying porosity (but no record of the porosity structure), and any attempt to calculate how much mass has been moved is a totally *&@#show.

I've modified it so that only material that is deposited is poofed by 1/(1-phi). This means material that is picked up, dropped, picked up, dropped may be multi-poofed. 

I am more than open to proposals to make this better. Note, however, that the mass conservation tests (described next) were written with an understanding of how E/D treats porosity.

### Tests:
the file `tests/components/erosion_deposition/test_ero_dep_mass_conservation.py` contains 2 tests (with 36 total test parameter combinations) of mass conservation for both Space and ErosionDepostion.  

it considered an entirely closed system and one in which there is an outlet. It permutes almost anything you might permute (space/ED, depressionFinder/None, no initial soil/initial soil, multiple phi, solver used), and uses dt!=1, vs!=1 to ensure that nothing is lurking behind a multiply by 1 bug. 

These enforce mass conservation in one timestep to 10^{-1} (though typically closer to 10^{-14}/numerical precision)

### One additional change:
1. When DepressionFinderAndRouter is run on a grid that has no open boundaries, it gives a completely obtuse error. It now gives a meaningful error and there is a test to demonstrate that it will do this. 

@gregtucker it is ready for you to look at.

gregtucker:
The above changes remove `phi` as a usable argument to ErosionDeposition. The rationale is explained in the following.

# Notes on Porosity in ErosionDeposition

## Basic definitions

Let subscripts $b$ and $m$ denote bulk and "mineral" quantities, respectively. The definitions and relations are:

- $Q_m$ - sediment flux, in mineral volume per time
- $Q_b = Q_m / (1-\phi)$ - "bulk" sediment flux, in deposit equivalent volume per time
- $E_b$ - entrainment rate, in bulk (deposit equivalent volume per time per area)
- $E_m = E_b (1-\phi )$ - same, in mineral volume
- $D_b$ - deposition rate, bulk (equivalent vertical accumulation rate)
- $D_m=(1-\phi) D_b$ - deposition rate, mineral volume only

## Theory

Sediment flux in terms of mineral volume is described by:

$\frac{\partial Q_m}{\partial x} = W (E_m - D_m)$.

where $W$ is width (when discretized, cell area is used for $W\Delta x$).

The entrainment equation gives $E_b$ rather than $E_m$:

$E_b = KA^mS^n$,

so we need to convert with:

$E_m = (1-\phi) E_b$.

The deposition term, however, gives us $D_m$ directly:

$D_m = V Q_m / A$.

So far we have:

$\frac{\partial Q_m}{\partial x} = W ((1-\phi)KA^mS^n - V Q_m / A)$.

For bed erosion/deposition rate (neglecting tectonics/baselevel), we have:

$\frac{\partial \eta}{\partial t} = D_b - E_b = \frac{VQ_m}{A(1-\phi)} - KA^mS^n$.

So, if we calculate entrainment in bulk terms, and deposition in mineral volume terms, we need to multiply by $(1-\phi)$ to convert entrainment rate into sediment addition, and need to divide by $(1-\phi)$ to convert deposition specific flux into aggradation rate.

Suppose, for example we have a numerical calculation that looks like this:

- `entrainment_rate` = $E_b$
- `depo_rate` = $D_m$

Then we would need to have the calculation of Qs multiply `entrainment_rate` by $1-\phi$. Likewise, to compute the net rate of erosional/depositional elevation change:

`net_ero_rate = entrainment_rate - depo_rate / (1 - porosity)`

Suppose, however, that we operated in terms of bulk equivalent flux, $Q_b$, instead of $Q_m$. According to the relationships above,

$(1-\phi) \frac{\partial Q_b}{\partial x} = W (E_m - D_m)$, which implies

$\frac{\partial Q_b}{\partial x} = \frac{W}{1-\phi} (E_m - D_m)$

$ = \frac{W}{1-\phi} \left( (1-\phi) E_b - \frac{VQ_m}{A}\right)$.

$ = W \left( E_b - \frac{VQ_m}{(1-\phi) A}\right)$.

Now, because $Q_b = Q_m / (1-\phi )$ above, we can simply substitute $Q_b$ on the RHS. This, along with substituting in the entrainment expression, gives us:

$\frac{\partial Q_b}{\partial x} = W \left( E_b - \frac{VQ_b}{A}\right)$.

So the porosity factor drops out as long as we are will to accept that the calculated sediment flux, $Q_b$, is to be interpreted as a **bulk** flux: not mineral volume per time, but rather the bulk deposit volume that would accumulate if all material were deposited.

## Current solution

I've modified the PR to do the following for the ErosionDeposition component:

- Removed use of phi in calculating poofing factors. This implies, as in the analysis above, that fluxes are bulk-equivalent, rather than mineral volume only.
- Retained the *parameter* `phi` for backward compatibility: previous codes that pass a `phi` parameter will still work.
- Updated the mass balance tests of ErosionDeposition component (only) so that porosity is no longer considered.
- Removed `phi` from arguments in those tests that directly instantiate ErosionDeposition (as opposed to those that instantiate a component that might be ED or might be Space).
- Removed tests made obsolete by the removal of `phi`

What I have not done:

- I haven't removed the parameter `phi` from the call signature in the mass balance tests.
- There is no warning if a user passes `phi` to the ErosionDeposition component (but there is a note in the parameter list).

landlab/components/erosion_deposition/erosion_deposition.py
```diff
@@ -53,6 +54,16 @@ class ErosionDeposition(_GeneralizedErosionDeposition):
     transport-limited response style. v_s == 1 means equal contributions from
     transport and erosion, and a hybrid response as described by Davy & Lague.
 
+    Unlike other some other fluvial erosion componets in Landlab, in this
```
gregtucker, line 57: Nice additional explanatory text :)

landlab/components/erosion_deposition/erosion_deposition.py
```diff
@@ -349,27 +354,15 @@ def _calc_erosion_rates(self):
...
-        """Calculate change in rock and alluvium thickness for a time period
-        'dt'.
-
-        Parameters
-        ----------
-        dt : float
-            Model timestep [T]
-        """
-
+    def _calc_qs_in_and_depo_rate(self):
```
gregtucker, line 356: Nice encapsulation in new function!

landlab/components/space/space.py
```diff
@@ -450,91 +447,138 @@ def run_one_step_basic(self, dt=1.0):
...
-        pos_flood = (self._q > 0) & (blowup) & (self._slope > 0) & (flooded)
-        self._soil__depth[pos_flood] = (
-            self._depo_rate[pos_flood] / (1 - self._phi)
-        ) * dt
 
-        # non-positive slopes, not flooded
-        non_pos_not_flood = (self._q > 0) & (blowup) & (self._slope <= 0) & (~flooded)
-        self._soil__depth[non_pos_not_flood] += (
-            self._depo_rate[non_pos_not_flood] / (1 - self._phi) * dt
+        # there is a numerical blow up as well, which is not discussed in the
```
gregtucker, line 476: Great to identify and address this.

landlab/components/space/space.py
```diff
@@ -450,91 +447,138 @@ def run_one_step_basic(self, dt=1.0):
...
-            (self._q > 0) & (~blowup) & (self._slope <= 0) & (~flooded)
-        )
-        self._soil__depth[neg_slope_not_flooded] += (
-            self._depo_rate[neg_slope_not_flooded] / (1 - self._phi) * dt
-        )
+        # Equation 12 gives dRdt, and Equation 36 gives R. However, we don't
+        # include dH/dt within timestep in integrating for R.
+        # This matters when we are starting with very little soil and increasing.
+
+        # to do this right I think we need a numerical integral for
```
gregtucker, line 562: Good at least to have this on the radar screen, thanks for leaving the breadcrumbs!

landlab/components/space/space.py
```diff
@@ -666,6 +679,8 @@ def run_with_adaptive_time_step_solver(self, dt=1.0):
 
             # Next we consider time to exhaust regolith
             time_to_zero_alluv[:] = remaining_time
+
+            # poof deposition by phi
```
gregtucker, line 683: yes

landlab/components/space/space.py
```diff
@@ -685,3 +700,52 @@ def run_with_adaptive_time_step_solver(self, dt=1.0):
 
             # Update remaining time and continue
             remaining_time -= dt_max
+
+
+def _dRdt(t, a, b, c, d, H0):
```
gregtucker, line 705: Nice encapsulation of rate of regolith/sed change

tests/components/depression_finder/test_lake_mapper.py
```diff
@@ -1997,3 +1997,16 @@ def test_find_lowest_node_on_lake_perimeter_c():
     assert find_lowest_node_on_lake_perimeter_c(
         node_nbrs, flood_status, elev, nodes_this_depression, pit_count, BIG_ELEV
     ) == (0, 2)
+
+
+def test_all_boundaries_are_closed():
```
gregtucker, line 2002: Great that this new check is also tested

tests/components/erosion_deposition/test_ero_dep_mass_conservation.py
```diff
@@ -0,0 +1,122 @@
+import pytest
```
gregtucker, line 1: So great to have mass balance tests! As noted in my PR comment, I did some tweaks to this.

gregtucker:
Over to you @kbarnhart ... 

The changes you made to Space look good to me, and it's so great to have this wonderful additional battery of meaningful tests!

See what you think of my removal of `phi` and modifications to related tests.
I'm ok with the PR in its present state (assuming tests pass; the component stuff does for me locally, and this time I remembered to run "make pretty") but let me know if you think we should continue to iterate.

landlab/components/erosion_deposition/erosion_deposition.py
```diff
@@ -172,7 +182,10 @@ def __init__(
         K : float, field name, or array
             Erodibility for substrate (units vary).
         phi : float
-            Sediment porosity [-].
+            Sediment porosity [-]. NOTE: This parameter is retained for backward
+            compatibility, but has no impact on the solution because all
+            calculations are performed in terms of bulk equivalent volume
+            or depth.
```
kbarnhart, line 188: @gregtucker I think all your changes are great. the only thing I would change is that I would argue that we should remove phi here and as something in the keyword arguments because it is not an input. We can make a note in the docstring that says that it previously was an option, but now it is not. I think leaving it will get people into the same trouble that **kwds did. 

I agree that we wan't to do as little as possible to make LL2 break pre LL2 code, but I think this is a case in which it should. If a user was using a pre v2.0 version of ErosionDeposition and they were passing phi, it did something. Now it doesn't. So no matter what we aren't maintaining backward compatibility.

However, I think this could be a really good opportunity to use **kwds. Specifically, we add **kwds, and if phi is passed by a user, we raise an error with an informative message indicating how and why this component is changed. This means that if a user had code that used phi as a keyword argument, we are making it easy for them to know what has changed (they don't have to dig through the source code, etc). 

I'm happy to implement this, but will wait to hear what you think.

gregtucker: @kbarnhart you make a fair point about breaking backward compatibility: we've already done it here, so might as well go all the way. More honest to have the component complain than to have it produce slightly different numbers. 

**kwds seems like a reasonable solution *if* we also guard against passing anything else, such as valid parameters that are just mis-spelled. In other words, **kwds needs to be empty or just contain phi. Do you want me to implement? Having now finished the above stuff sooner than I thought would be possible, I have some time.

kbarnhart: I'm on it. But I think it would be best if you looked at my error text to make sure it is as comprehensible as possible.

kbarnhart:
@gregtucker now back ready for you to look at.

tests/components/erosion_deposition/test_erodep.py
```diff
@@ -22,7 +22,29 @@ def test_route_to_multiple_error_raised():
     fa.run_one_step()
 
     with pytest.raises(NotImplementedError):
-        ErosionDeposition(mg, K=0.01, phi=0.0, v_s=0.001, m_sp=0.5, n_sp=1.0, sp_crit=0)
+        ErosionDeposition(mg, K=0.01, v_s=0.001, m_sp=0.5, n_sp=1.0, sp_crit=0)
```
gregtucker, line 25: Nice!

gregtucker:
@kbarnhart this looks great! I made a minor edit to the message regarding phi. I think we're good to go :-)

PR #1189, opened 2020-04-28
Fix manylinux builds

mcflugen:
This pull request fixes creating *manylinux* builds for landlab on Travis using Docker. It seems the issue was that a new version of the *auditwheel* command now returns an error code (1) if it's given a non-platform wheel to repair. Since we just looped over all wheels in the *wheelhouse* folder, passing each one to *auditwheel*, *auditwheel* would return an error and the script would exit prematurely before creating further *manylinux* distributions.

To fix the problem, I simply allow *auditwheel* to return a non-zero code and continue with the loop.

### DIFF
--- a/.ci/travis/build_wheels.sh
+++ b/.ci/travis/build_wheels.sh
@@ -14,7 +14,7 @@ done
 
 # Bundle external shared libraries into the wheels
 for whl in /io/wheelhouse/*.whl; do
-    auditwheel repair "$whl" --plat $PLAT -w /io/wheelhouse/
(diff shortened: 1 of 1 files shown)


kbarnhart:
@mcflugen I don't see why this wouldn't work. Though I agree that there is no way to know for sure without a new release.

mcflugen:
@kbarnhart We could test it on another pre-release version? I just wish the CI builds didn't take *soooooooo* long.

kbarnhart:
@mcflugen agreed on both counts

mcflugen:
@kbarnhart I just added a *v2.0.0b8* tag from this branch to test the *manylinux* build. I guess we should know if it works in about 3 hours... 🤞

PR #1187, opened 2020-04-27
Merge master into release for v2.0.0 

kbarnhart:

PR #1191, opened 2020-04-29
Merge master into release

kbarnhart:


kbarnhart:
Do you understand why this doesn't see the prior merge I made on #1187. Is it because that was a squash merge?

How do you think I should handle this so that in the future its is cleaner?

mcflugen:
@kbarnhart Not sure why this doesn't see your prior merge. I couldn't definitely be wrong, but I don't think it would be because of the squash merge.

mcflugen:
@kbarnhart I see the Travis build failed in the testing-the-docs-stage with a broken link. I'm going to go ahead and treat that as a pass.

Looks good! 👍

kbarnhart:
Would you squash-merge this? Or one of the other options?

mcflugen:
Yay! 🎉

PR #1199, opened 2020-05-15
Minor cleanup

kbarnhart:
I was testing the NetworkSedimentTransporter binder instances and found that 
(a) one of the links was broken (also in SpeciesEvolver and Dupuit), so I fixed it.
(b) some notebooks had been saved with output (so I deleted the output). 

I also found that the binder instance is still on 2.0.1

import landlab
landlab.__version__
2.0.1

not sure why... conda forge and pip have the new 2.1.0 binaries... and binder is configured by our environment.yml file... 

But making sure that binder works for the network sediment transporter in the near term would be good... 

also tagging @pfeiffea and @jonczuba


mcflugen:
@kbarnhart This *binder* issue is strange. Maybe the old environment is cached on *binder* and not being updated? Since the notebooks require *v2.1* to work, it makes sense to specify this in the environment file. We should do this.

I just updated the [*CSDMS* *JupyterHub*](https://csdms.rc.colorado.edu) to *landlab==2.1.0* and it seems to be working with the new *NetworkSedimentTransporter* notebooks. This is to say, I don't think there are any package version conflicts preventing *landlab* from updating to the newest version.

kbarnhart:
@mcflugen I can specify the version number on this branch, and then I suspect it makes sense to make 2.1.1 to make sure binder sees the right environment file. Most of the binder URLs in the docs point to landlab/landlab/release.

kbarnhart:
@mcflugen I just double checked and all pointers to binder go to release. 

search in repo for: binder.org/v2/gh/landlab/landlab/

mcflugen:
@kbarnhart That makes sense.

I'm ok if you want to merge this before the tests finish.

kbarnhart:
Cool. I'm testing that it works on binder (which really serves as the real test). 

Currently building...

https://mybinder.org/v2/gh/landlab/landlab/barnhark/minor_cleanup?filepath=notebooks%2Fwelcome.ipynb

OK to make v2.1.1 ?

Also, you somehow managed to master, release, and v2.1.0 on the same commit hash... how did you do that?

kbarnhart:
@mcflugen all the new NST notebooks ran on binder so I'm merging. 

Not going to make a release without your input since I want to learn how to do the magic same-commit-hash juju.

PR #1200, opened 2020-05-16
v2.1.1

mcflugen:


kbarnhart:
@mcflugen fyi I tested binder and it now works as expected. thanks!

PR #1202, opened 2020-05-19
fix: remove fetched folder.

kbarnhart:
@mcflugen without these extra two lines the tests can only pass once for me. otherwise the folder `methow` persists and the test fails with an error saying (smartly) that the folder is already there.


kbarnhart:
@mcflugen OK to merge?

Issue #1204, opened 2020-05-20
Add Height Above Nearest Drainage (HAND) component? 


DavidLitwin:
Height Above Nearest Drainage (HAND) is a relatively new topographic metric that quantifies the elevation difference between any point in a DEM and the nearest point that is identified as a stream channel (Nobre et al 2011). In this way, it is the vertical complement of a flow path length, and average HAND is the vertical complement of 1/DD. Lately, it's appeared in the hydrological literature for prediction of areas of saturation, and even in a "calibration-free" rainfall runoff model (Nobre et al 2011, Gao et al 2018). 

I intend to use this metric to analyze synthetic topography generated from a landscape evolution model, but it would serve equally well for DEM analysis. To my knowledge, there are no python libraries that are currently set up to calculate it. HAND feels like a good fit for landlab in the spirit of the DrainageDensity and ChannelProfiler components. 

My intuition (and later a conversation with @kbarnhart) tells me that all of the information needed to calculate HAND should already be available from the single direction (D4, D8) FlowDirector components, but I am not an expert in these and it would likely be difficult for me to do this alone. [Nobre et al (2011)](https://github.com/landlab/landlab/files/4657155/Nobre.et.al_2011_Height.Above.the.Nearest.Drainage.a.hydrologically.relevant.new.terrain.model.pdf) describe the method they used to calculate HAND, attached here. I think it would be a substantial contribution to Landlab that others may find very useful too. Is anyone willing to help me on this?
@kbarnhart, @mcflugen, or @gregtucker?

Cited:
Gao, H., Birkel, C., Hrachowitz, M., Tetzlaff, D., Soulsby, C., & Savenije, H. H. G. (2018). A simple topography-driven and calibration-free runoff generation module. Hydrology and Earth System Sciences Discussions, 787–809. https://doi.org/10.5194/hess-2018-141

Nobre, A. D., Cuartas, L. A., Hodnett, M., Rennó, C. D., Rodrigues, G., Silveira, A., et al. (2011). Height Above the Nearest Drainage – a hydrologically relevant new terrain model. Journal of Hydrology, 404(1), 13–29. https://doi.org/10.1016/j.jhydrol.2011.03.051

kbarnhart:
@DavidLitwin I think this is a great idea and would make a great contribution. I'm not able to commit a lot of time, but am happy to provide guidance on using the data structures built by the flow director components.

DavidLitwin:
[...]

PR #1207, opened 2020-05-26
Add HeightAboveDrainageCalculator component

DavidLitwin:
Addressing the need I described in #1204, this PR includes a new component that calculates the height of every node above the nearest drainage node (the union of nodes specified with a channel mask and open boundary nodes).

### DIFF
--- a/landlab/components/hand_calculator/hand_calculator.py
+++ b/landlab/components/hand_calculator/hand_calculator.py
@@ -10,6 +10,7 @@ import numpy as np
 from landlab import Component
 from landlab.utils import return_array_at_node
 
+
 class HeightAboveDrainageCalculator(Component):
     """
     Calculate the elevation difference between each node and its nearest
@@ -125,7 +126,7 @@ class HeightAboveDrainageCalculator(Component):
         },
     }
 
-    def __init__(self, grid, channel_mask = "channel__mask"):
+    def __init__(self, grid, channel_mask="channel__mask"):
         """
         Parameters
         ----------
--- a/landlab/components/hand_calculator/hand_calculator.py
+++ b/landlab/components/hand_calculator/hand_calculator.py
@@ -115,9 +115,7 @@ class HeightAboveDrainage(Component):
 
         # Downstream drainage node
         if "downstream_drainage__node" in grid.at_node:
-            self._downstream_drainage_id = grid.at_node[
-                "downstream_drainage__node"
-            ]
+            self._downstream_drainage_id = grid.at_node["downstream_drainage__node"]
         else:
             self._downstream_drainage_id = grid.add_zeros(
                 "downstream_drainage__node", at="node", dtype=int
@@ -125,31 +123,29 @@ class HeightAboveDrainage(Component):
 
         # height above nearest drainage
         if "height_above_drainage__elevation" in grid.at_node:
-            self._hand = grid.at_node[
-                "height_above_drainage__elevation"
-            ]
+            self._hand = grid.at_node["height_above_drainage__elevation"]
         else:
             self._hand = grid.add_zeros(
                 "height_above_drainage__elevation", at="node", dtype=float
             )
-
 
     @property
     def channel_mask(self):
         return self._channel_mask
 
     @channel_mask.setter
-    def channel_mask(self,new_val):
+    def channel_mask(self, new_val):
         self._channel_mask = new_val
-
 
     def run_one_step(self):
 
-        self_draining_nodes = sum(self._receivers == np.arange(self._grid.number_of_nodes))
+        self_draining_nodes = sum(
+            self._receivers == np.arange(self._grid.number_of_nodes)
+        )
         if self_draining_nodes != len(self._grid.boundary_nodes):
             warn(
-            "Pits detected in the flow directions supplied. "
-            "Pits will be treated as drainage nodes."
+                "Pits detected in the flow directions supplied. "
+                "Pits will be treated as drainage nodes."
             )
 
         self._downstream_drainage_id[:] = 0
@@ -158,7 +154,9 @@ class HeightAboveDrainage(Component):
 
         for i in range(self._grid.number_of_nodes):
 
-            if i == self._receivers[i] or is_drainage_node[i]: #started on a boundary, depression, or in channel
+            if (
+                i == self._receivers[i] or is_drainage_node[i]
+            ):  # started on a boundary, depression, or in channel
                 self._downstream_drainage_id[i] = i
 
             else:
@@ -173,4 +171,4 @@ class HeightAboveDrainage(Component):
                         cur_node = downstream_id
 
         nearest_drainage_elev = self._elev[self._downstream_drainage_id]
-        self._hand[:] = self._elev-nearest_drainage_elev
+        self._hand[:] = self._elev - nearest_drainage_elev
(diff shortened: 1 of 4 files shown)


DavidLitwin:
I thought I would get this started here, @kbarnhart  and @gregtucker. The component works and is fast enough for the size grids I work on without cythonizing it. That might be a project for a later date. Component unit tests to come in the next few days!

DavidLitwin:
I should add that what I've written is a kind of minimal version of HAND. It doesn't include the slope-area calculations to create or update the channel mask like the DrainageDensity component, nor does it first fill pits like the methodology specified in Nobre et al 2011 paper, though I believe the component should warn if pits are detected. I'll leave it up to the user to determine the correct channel mask and get the flow routing how the want it outside of the component.

landlab/components/hand_calculator/hand_calculator.py
```diff
@@ -0,0 +1,195 @@
+# -*- coding: utf-8 -*-
+"""Landlab component to calculate height above nearest drainage."""
+from warnings import warn
+
+import numpy as np
+
+from landlab import Component
+
+
+class HeightAboveDrainage(Component):
```
gregtucker, line 10: Usually the convention is to have class and component names fit in the syntax of "X is a Y", for example: "is a string," "is an array," "is a RasterModelGrid," "is a LinearDiffuser". Consider modifying the name.

landlab/components/hand_calculator/hand_calculator.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+    Calculate the elevation difference between each node and its nearest
+    drainage node in a DEM.
+
+    This component implements the method described by Nobre et al (2011). A
+    single direction flow director (D8 or steepest descent) must be run prior
+    to HeightAboveDrainage to supply the flow directions. This component does
+    not fill depressions in a DEM, instead it treats them as drainage nodes.
```
gregtucker, line 18: "DEM, but rather treats them..." or "DEM; instead, it treats them..."

gregtucker:
@DavidLitwin this is gorgeous, bravo!

One thought I had is that you might be able to implement a faster algorithm without too much work. I suspect that the current algorithm might be O(N log N) or something like that, because you need to visit cells multiple times. There might be a relatively easy O(N) algorithm. A side effect of running the FlowAccumulator is the creation of the ordered array (`flow__upstream_node_order`), which lists node IDs in downstream-to-upstream order. The FlowAccumulator includes a "delta" array, which is a list containing, for each node, the array index where that node's donor list begins (`flow__data_structure_delta`). I suspect that by using these two together, you could make an algorithm that walks upstream on the stack, setting the absolute height of the nearest drainage for the immediate donors to each node. Then once you have that, calculating the height difference would be a one-liner. My very quick (probably flawed) algorithm is:

For each node n on the downstream-to-upstream array:
  If n is a drainage (or open boundary), set its nearest-drainage elevation to its own elevation
  If n is not a closed boundary node:
    For each donor j (obtained from the "delta" array):
      Set nearest-drainage elevation to n's value

hand_array = elev_array - nearest_drainage_elev_array

You can obtain the donors from the "delta" array: for a given node n, delta[n-1] contains the starting index, and delta[n] contains the ending index (non-inclusive). These are indices into the `flow__upstream_node_order` array. In other words, to get the IDs of donors to n:

`flow__upstream_node_order[delta[n-1]:delta[n]]`

where I'm using the shorthand "delta" for the delta array. In some cases, it will be empty (no neighbors), in some cases just one element, and in some cases two or more.

DavidLitwin:
Thank you @gregtucker, I think you have demystified for me how to use the stack! I will give it a shot and see if I can come up with something that works equivalently to what I have here.

kbarnhart:
@DavidLitwin this is great! I'll take a look by the end of the week. 

I also agree that it shouldn't be this component's job to make the drainage map.

landlab/components/hand_calculator/hand_calculator.py
DavidLitwin: Changed to HeightAboveDrainageCalculator, more in line with the HackCalculator.

landlab/components/hand_calculator/hand_calculator.py
DavidLitwin: Changed this as well, thanks!

DavidLitwin:
I've spent a while working on changing the form to use the stack structure and I still can't get it. As I understand the approach you suggested, @gregtucker : 

```
        for n in self._node_order:

            if is_drainage_node[n]:
                nearest_drainage_elev[n] = self._elev[n]
                if not n in self._grid.closed_boundary_nodes:
                    donors = self._node_order[self._delta_structure[n-1]:self._delta_structure[n]]
                    nearest_drainage_elev[donors] = nearest_drainage_elev[n]

        self._hand[:] = self._elev - nearest_drainage_elev
```

Where `_node_order` is `flow_upstream__node_order` and `_delta_structure` is the `flow__data_structure_delta`. The donors seem to miss some nodes that I was getting with flow direction alone. Even though it would run faster if I could get this to work, I'm tempted to leave it how it is (a) for the sake of my time, (b) because there's nothing in the component that needs the drainage area calculated with the flowaccumulator, so it seems kind of nice to be able to leave it with just flowdirector dependency. What do you think?

Also, I don't think I understand the docstring tests! Any advice on that would be appreciated.

gregtucker:
It's not obvious to me why the algorithm above would miss out some nodes... I agree for now it's not a terrible idea to stick with the algorithm you already have. Ultimately though I suspect someone is going to want to run this on a huge lidar DEM, and that's where an O(N) algorithm would really shine.

DavidLitwin:
That makes sense. A project for another day! I'll wait until someone makes an issue, or I want to do it myself.

kbarnhart:
I think I can answer the question. The nodes that are donors are not indexed by delta in the stack, but in an array called "D", which I don't think we store anymore. 

Braun and Willet Table 1 and Equation 10 are the right place to look. 

There is, however, another way. Key thing to note is that the stack order means that all of your downstream neighbors will be processed before you. This means that you can look downstream and (if you are not a drainage node) you can see what the drainage elevation is. 

A key assumption here is that the way drainage nodes are specified needs to work with the stack order... I think this is almost always the case. 

```python
receiver = grid.at_node['...']
nearest_drainage_elev= empty() 
for n in self._node_order:
    r = receiver[n]
    # if not drainage node set drainage elevation to downstream. 
    if not is_drainage_node[r]:
        nearest_drainage_elev[n] = nearest_drainage_elev[r]
    else: # set elevation of drainage to self. 
        nearest_dranage_elev[n] = elev[n]
self._hand[:] = self._elev - nearest_drainage_elev
```
** this pseudocode is probably almost right, but may have 🐛 s

gregtucker:
@kbarnhart yes that's it! You can look *downstream*. That should work!

DavidLitwin:
@kbarnhart that worked! The only bug is that the conditional statement needs to be whether the current node is drainage, rather than the node downstream. As far as I can tell, what I just sent should be complete.

DavidLitwin:
Ah wait, there is one thing that might need to be clarified. Currently channel__mask is a non-optional argument to init, but there's also an optional channel__mask in _info that I never do anything with. What's the protocol here?

DavidLitwin:
@gregtucker I think I'm all set here, so if you have no further comments I'll merge!

PR #1210, opened 2020-06-05
Gt/updates to flow routing

gregtucker:
This PR adds LakeMapperBarnes as a possible lake mapper to use with the FlowAccumulator. It also removes the need for duplicate execution of 3 of the stack-related functions.

### DIFF
--- a/landlab/components/flow_accum/flow_accumulator.py
+++ b/landlab/components/flow_accum/flow_accumulator.py
@@ -690,7 +690,6 @@ class FlowAccumulator(Component):
         keyword arguments, tests the argument of runoff_rate, and
         initializes new fields.
         """
-        # print('FA>init')
         super().__init__(grid)
         # Keep a local reference to the grid
 
@@ -723,12 +722,7 @@ class FlowAccumulator(Component):
 
         self._node_cell_area = node_cell_area
 
-        # STEP 2:
-        # identify Flow Director method, save name, import and initialize the correct
-        # flow director component if necessary
-        self._add_director(flow_director)
-        self._add_depression_finder(depression_finder)
-
+        # STEP 3:
         # This component will track of the following variables.
         # Attempt to create each, if they already exist, assign the existing
         # version to the local copy.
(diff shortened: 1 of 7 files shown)


kbarnhart:
👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

SiccarPoint:
Nice! Just a note to say that if you're evolving topography, it's unlikely this is the fill you will want. Although it's MUCH superior to the old algorithms on general topography, in a landscape with or heading towards few depressions, it is considerably worse performing...

landlab/components/flow_accum/flow_accumulator.py
```diff
@@ -957,15 +959,27 @@ def _add_director(self, flow_director):
...
+                "reroute_flow",
+                "surface",
+                "method",
+                "fill_flat",
+                "fill_surface",
+                "redirect_flow_steepest_descent",
+                "reaccumulate_flow",
+                "ignore_overfill",
+                "track_lakes",
+            ]
```
mdpiper, line 982: Would [__slots__](https://docs.python.org/3/reference/datamodel.html#slots) be useful here? This is a speculative comment, not a review comment.

gregtucker:
Thanks @SiccarPoint and @mdpiper. To address the performance issue, I added a line that tests for the presence of pits (grid.at_node['flow__sink_flag'] is nonzero for any core node) as a condition for running the lake mapper. I also fixed the second/third issue. I don't really understand `__slots__` but would be interested to learn more.

SiccarPoint:
Good idea!

SiccarPoint:
Great idea

landlab/components/flow_accum/flow_accumulator.py
```diff
@@ -1122,8 +1143,8 @@ def accumulate_flow(self, update_flow_director=True, update_depression_finder=Tr
             # At the moment, no depression finders work with to-many, so it
             # lives here
             if self._depression_finder_provided is not None:
-                if update_depression_finder:
-                    self._depression_finder.map_depressions()
+                if update_depression_finder and self.pits_present():
+                    self._depression_finder.update()
```
kbarnhart, line 1147: This is the problem line for #1276

PR #1211, opened 2020-06-11
Add callback function to GroundwaterDupuitPercolator

DavidLitwin:
In my work I've found it useful to be able to get some information about the  state of the groundwater model during the intermediate steps taken by the adaptive timestep solver. In particular I am  using grid info to determine transient runoff generation patterns and develop storage-discharge relationships.

Here I've added the option for a user to supply a function that is executed every substep of the adaptive timestep solver.  It could append to a list outside of the model instance, print soemthing, write to a file, etc. The function supplied should have two required arguments, the grid and the substep duration, which I think should be sufficiently general for almost anything you would want to do as far as analysis and testing. 

There is an example of how I'm using this in the component tests. I suspect that some testing should happen in __init__ to make sure that the function supplied is of the right type and has the right arguments, but I didn't know how to go about this exactly. Any other ideas you have on this would be appreciated!

### DIFF
--- a/landlab/components/groundwater/dupuit_percolator.py
+++ b/landlab/components/groundwater/dupuit_percolator.py
@@ -577,7 +577,7 @@ class GroundwaterDupuitPercolator(Component):
                 "water table above elevation surface. "
                 "Setting water table elevation here to "
                 "elevation surface"
-                )
+            )
             self._wtable[self._wtable > self._elev] = self._elev[
                 self._wtable > self._elev
             ]
@@ -660,7 +660,7 @@ class GroundwaterDupuitPercolator(Component):
                 "water table above elevation surface. "
                 "Setting water table elevation here to "
                 "elevation surface"
-                )
+            )
             self._wtable[self._wtable > self._elev] = self._elev[
                 self._wtable > self._elev
             ]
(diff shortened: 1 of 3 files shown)


DavidLitwin:
@gregtucker I tried a few things but I can't figure out what's going on here with the docs build. Do you have any ideas?

gregtucker:
@davidlitwin I suggest testing to see if you can build the docs on your branch locally. To do so, navigate to landlab/docs and run "make html". Actually, you might might try it first with master, then with your branch, to see whether there are any differences in the build output that might indicate the problem.

kbarnhart:
So the error is 
```
/home/travis/build/landlab/landlab/landlab/components/groundwater/dupuit_percolator.py:docstring of landlab.components.groundwater.dupuit_percolator.GroundwaterDupuitPercolator:163:Inline strong start-string without end-string.
```

It refers to line 163 of the docstring (not the document, which I find to be a pain). In this case, the docstring for andlab.components.groundwater.dupuit_percolator.GroundwaterDupuitPercolator is split in two strings (one at the class definition and one at the init). No, I did not count to line 163 of the docstring... I googled the error and found [this stack overflow](https://stackoverflow.com/questions/30454549/a-literal-in-restructuredtext) and figured out that I should look for *s. 

In this case, the key thing is that ** is the notation for strong emphasis. 

The **kwds in the docstring provides a "strong start string" but there is no matching end string. 

An additional note on sphinx errors and warnings, sometimes they are the line before or after the actual problem code. 

Also, `make docs` will only work on linux.

kbarnhart:
@DavidLitwin this is awesome.

kbarnhart:
@mcflugen looks like there is a bit of a conflict between lint and sphinx quality checking...

@DavidLitwin needs to put ** in a docstrning (which, understandably sphinx doesn't like). 

When I escaped it to `\*\*` lint yells with:
```
landlab/components/groundwater/dupuit_percolator.py:330:52: W605 invalid escape sequence '\*'
```
do you see any danger to adding W605 to the ignored lint items?

mcflugen:
@kbarnhart Maybe turn the docstring that's giving trouble into a raw string? That should, I think, eliminate the *invalid escape sequence* error.

kbarnhart:
@mcflugen that worked for me locally. thanks!

kbarnhart:
@DavidLitwin looks like this should pass. I'll approve it and let you merge.

DavidLitwin:
Thanks for sorting this out @kbarnhart!

PR #1217, opened 2020-07-07
Gt/matrix builder

gregtucker:
Adds a utility with two functions that build a matrix for a finite-difference or finite-volume solution. Includes a fairly extensive tutorial

### DIFF
--- a/docs/source/reference/components/hand_calculator.rst
+++ b/docs/source/reference/components/hand_calculator.rst
@@ -1,0 +1,7 @@
+HeightAboveDrainageCalculator: Calculate height above nearest drainage
+----------------------------------------------------------------------
+
+.. automodule:: landlab.components.hand_calculator.hand_calculator
+    :members:
+    :undoc-members:
+    :show-inheritance:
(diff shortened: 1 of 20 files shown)


gregtucker:
@mcflugen I added a generic function `links_by_tail_and_head_status`. Kept the specific functions (which now call the generic one) for convenience, but maybe we should remove them to avoid clutter, what do you think?

mcflugen:
@gregtucker Looks great!

It looks like you've removed the specific functions. That makes sense to me.

I've added a new generic function, called *links_with_node_status*. It does the same as your *links_by_tail_and_head_status* function but I thought the name was a little better. I kept the old name in case you liked it better but ultimately we should only have one.

The *links_with_node_status* function I added accepts *None* for either *status_at_head* and *status_at_tail* so that a user can identify links based on the status of either its head or tail while ignoring the status of the other (if both *status_at_head* and *status_at_tail* are *None*, all the links are returned).

gregtucker:
Great idea to allow specification of just head or tail status, @mcflugen. I also like your name, so I removed the original function in favor of your version.

Issue #1228, opened 2020-08-08
Unfortunate change in stream_power_channels_class_notebook


kxwhipple:
In the migration to landlab 2 something inside StreamPowerEroder, seems to have changed such that the network structure in the landscapes produced when running this classroom exercise (default parameters) has degraded significantly. Back under landlab 1 the network structure was a reasonable dentritic pattern and the landscape looked realistic. At present the drainage network structure produced (seemingly with no changes in the notebook, so I'm guessing the change is inside StreamPowerEroder) is in a word unfortunate and very unrealistic to the point students will not recognize the output as a landscape.  
Here's the topography currently generated after 1 Ma:
![drainage_network_landlab2_output](https://user-images.githubusercontent.com/40966478/89717243-52abb680-d969-11ea-8318-bd1386cfc140.png)

Here's what it looked like in the previous version:
![drainage_network_landlab1_output](https://user-images.githubusercontent.com/40966478/89717260-70791b80-d969-11ea-9425-295a959ebb70.png)

I think you can see what I mean. I'd really like to keep using this notebook for teaching, but if the landscape doesnt look realistic at all it would really undermine the learning.  Thanks
Kelin

gregtucker:
Thanks for pointing this out @kxwhipple! I agree it looks weird. The difference you're seeing is the change from D8 flow routing to D4 routing as the default option in the FlowAccumulator component. It's easy to switch back to D8. If @nicgaspar agrees, we can make this change in the notebook (I'll submit a pull request shortly).

In the meantime, it's an easy change to the notebook. In Code Block 5, change the line:

`frr = FlowAccumulator(mg1)  # intializing flow routing`

to

`frr = FlowAccumulator(mg1, flow_director='FlowDirectorD8')  # intializing flow routing`

PR #1230, opened 2020-08-09
use D8 routing in channel evolution notebook

gregtucker:
This PR simply adds the D8 routing option (`flow_director='D8'`) in the teaching notebook on stream channel erosion and landscape evolution. Channels look more realistically dendritic with D8 routing.

Addresses #1228


nicgaspar:
Thanks for doing this @gregtucker

PR #1233, opened 2020-08-17
Fix doc bugs

elbeejay:
Was looking at the documentation for landlab and came across a few things that didn't seem quite right. 

1. In the `grid` [section of the user guide](https://landlab.readthedocs.io/en/master/user_guide/grid.html#adding-data-to-a-landlab-grid-element-using-fields) several `=` signs were displayed as `-`. The `add_field` function was shown with the inputs `value_array` and `at='node'` flipped. Also when written as `clobber=False` this would not work if the previous code block defining the field 'elevation' with 0s has been run. 

2. In the [FlowAccumulator documentation](https://landlab.readthedocs.io/en/master/reference/components/flow_accum.html#landlab.components.flow_accum.flow_accumulator.FlowAccumulator) it looked like one of the example code blocks from the docstring didn't render. I think it is because of a missing blank line, so I added one.

I assume it is okay to open a PR to the master branch as I didn't see an obvious 'dev' or 'develop' branch. But my apologies if this isn't quite the right way to open a PR for this project.

-Jay

### DIFF
--- a/docs/source/user_guide/grid.rst
+++ b/docs/source/user_guide/grid.rst
@@ -151,7 +151,7 @@ whose length is equal to the number of nodes in the grid.
 
 .. code-block:: python
 
-    z - mg.add_zeros("elevation", at="node")
+    z = mg.add_zeros("elevation", at="node")
 
 Here *z* is an array of zeros. We can verify that *z* has the same length as the number of nodes:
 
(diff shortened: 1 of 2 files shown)


kbarnhart:
@elbeejay thanks so much for noticing these doc build/formatting errors and making a PR. Its very much appreciated! 🎉  

Submitting it to `master` is exactly the right thing to do. 

I'm going to double check the docs in these places look right on my local build, and will then merge.

kbarnhart:
I checked these on a local build of the html and they look great. Thanks again @elbeejay, merging.

PR #1235, opened 2020-08-18
Small NST Joss paper fixes

kbarnhart:
A few changes to the JOSS paper .md and .bib files (for reference formatting, primarily), also fixing @mcflugen's affiliation. 

fyi: @pfeiffea


mcflugen:
Thanks fixing that @kbarnhart (although I would love to work with you at the USGS!)

PR #1237, opened 2020-08-31
Small changes to JOSS paper

kbarnhart:
Address comments from [this comment](https://github.com/openjournals/joss-reviews/issues/2341#issuecomment-682831094) from JOSS reviewer @zsylvester (thanks!)


kbarnhart:
@mcflugen 🆘 ... looks like some new errors associated with notebook testing.

mcflugen:
@kbarnhart I had a quick look but wasn't able to reproduce the error. The problem looks to be with the `--ExecutePreprocessor.timeout=None` option. Maybe instead of `None`, we could just use a large number?

I'll have a closer look at this tomorrow.

kbarnhart:
Thanks @mcflugen!

Did a bit of googling and based on [this](https://nbconvert.readthedocs.io/en/latest/execute_api.html#execution-arguments-traitlets) part of nbconvert's docs it looks like None and -1 are the valid "no timeout" options... Have switched it to -1.

kbarnhart:
@mcflugen looks like this may just be a passing fluke (and/or the -1 change was sufficient)....

kbarnhart:
only failure is a linkcheck timeout in docs. so merging.

PR #1240, opened 2020-09-18
add cython version of priority queue

gregtucker:
Adds a cythonized version of the StablePriorityQueue class created by @SiccarPoint. This one is called StablePriorityQueueC to differentiate it. The hope is that it increases performance of LakeMapperBarnes, and perhaps future uses of a PQ (note: there is also one implemented in CellLab; ultimately they should probably be merged).

Differences:
o Testing of "removed" entries uses != instead of "is not", because I believe use of nan as code for "REMOVED" should unconditionally pass the "is not" test.
o The peek method in StablePriorityQueue has the potential for an infinite loop if the first item was flagged as remove (I think the first bullet point kept it from being triggered in the tests). Modified it to use a counter to keep accessing the next item until a non-REMOVED one is found. However, I don't know whether this is the desired behavior? Alternative behavior would be to return None if the first item is REMOVED (in which case the while loop should not be needed...?) This required changing one of the test outcomes (I still don't understand why that test passed).
o Removed the tracking of tasks_ever_added, to prevent ballooning of memory usage.
o Replaced the use of an iterator for counting with a simple integer and increment.

@SiccarPoint can you take a quick look and see whether you think this does the trick?

### DIFF
--- a/landlab/utils/__init__.py
+++ b/landlab/utils/__init__.py
@@ -13,6 +13,7 @@ from .source_tracking_algorithm import (
     track_source,
 )
 from .stable_priority_queue import StablePriorityQueue
+from .cfuncs import StablePriorityQueueC
 from .watershed import (
     get_watershed_mask,
     get_watershed_masks,
--- a/landlab/utils/cfuncs.pyx
+++ b/landlab/utils/cfuncs.pyx
@@ -1,0 +1,86 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import numpy as np
+cimport numpy as np
+cimport cython
+
+import heapq
+
+
+cdef double LARGE_ELEV = 9999999999.0
+
+
+cdef class StablePriorityQueueC:
+    """Implements a stable priority queue, that tracks insertion order; i.e.,
+    this is used to break ties.
+
+    See https://docs.python.org/2/library/heapq.html#priority-queue-implementation-notes
+    & https://www.sciencedirect.com/science/article/pii/S0098300413001337
+    """
+    cdef public list _pq
+    cdef dict _entry_finder
+    cdef float _REMOVED
+    cdef int _counter
+
+    def __init__(self):
+        self._pq = []  # list of entries as a heap
+        self._entry_finder = {}  # mapping of tasks to entries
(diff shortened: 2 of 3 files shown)


SiccarPoint:
I believe you have an issue here... The peek is no longer working right. i.e., the part of the testing that does this:
'''
task = q.peek_at_task()
assert(task == 'b')
task = q.pop_task()
assert(task == 'a')
'''
Should not ever be permitted, as by definition the peek should not change the queue structure. This should return 'b' again in this case.

The difficulty is around the 'float("inf")' issue, as you spotted. I *think* what I have was actually working correctly, since the below is also working. It's the actual same bit of memory being allocated, so the value doesn't matter. The inf is simply a dummy value in this case, so in a manual output review it would be obvious what was happening.
'''
>>> a = float("inf")
>>> b = a
>>> c = a
>>> float("inf") is not float("inf")
True
>>> float("inf") != float("inf")
False
# both as you noted in your text, but:
>>> b is not c
False
# ...which is equivalent to the code as written, and behaving correctly
'''
So, whatever the tests were doing before _is_ the correct behaviour, I think.

gregtucker:
Thanks @SiccarPoint. In the original Python version of the `peek_at_task` method, how do you avoid an infinite loop in the case where the top-level task (`self._pq[0]`) is flagged as `_REMOVED`?

```
    def peek_at_task(self):
        """Return the lowest priority task without removal.

        Raise KeyError if empty.
        """
        while self._pq:
            priority, count, task = self._pq[0]
            if task is not self._REMOVED:
                return task
        raise KeyError("peeked at an empty priority queue")
```

SiccarPoint:
This implementation followed [this documentation](https://docs.python.org/2/library/heapq.html) fairly closely. I've clearly had a bit of a brainfade implementing the peek. This should I think look like this:

```
    def peek_at_task(self):
        """Return the lowest priority task without removal.

        Raise KeyError if empty.
        """
        while self._pq:
            priority, count, task = self._pq[0]
            if task is not self._REMOVED:
                return task
            # If we get here, the highest priority entry is not valid, so we can clear it safely:
            _ = heapq.heappop(self._pq)
        raise KeyError("peeked at an empty priority queue")
```
This should restore the ability to proceed through the loop.

PR #1241, opened 2020-09-20
add Wavefront OBJ output

gregtucker:
Simple utility to write a Landlab grid and 1 field (default `topographic__elevation`) to a Wavefront OBJ file, for import into visualization programs like Blender.

### DIFF
--- a/tests/io/test_write_obj.py
+++ b/tests/io/test_write_obj.py
@@ -49,12 +49,6 @@ f 5// 6// 9//
 """
 
 
-def remove_newline_chars(mystr):
-    mystr = mystr.replace("\n", "")
-    mystr = mystr.replace("\r", "")
-    return mystr
-
-
 def test_write_to_filelike(tmpdir):
     grid = RasterModelGrid((3, 3), xy_spacing=2.0)
     z = grid.add_zeros("topographic__elevation", at="node")
--- a/tests/io/test_write_obj.py
+++ b/tests/io/test_write_obj.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python3
+
+"""unit tests for landlab.io.obj module"""
+
+import os
+import tempfile
+from landlab import HexModelGrid, RasterModelGrid
+from landlab.io import write_obj
+
+LITTLE_HEX_OBJ = """# landlabgrid
+#
+g landlabgrid
+v 1.0 0.0 0.0
+v 3.0 0.0 0.0
+v 0.0 1.732051 0.0
+v 2.0 1.732051 1.0
+v 4.0 1.732051 0.0
+v 1.0 3.464102 0.0
+v 3.0 3.464102 0.0
+f 4// 1// 2//
+f 4// 3// 1//
+f 5// 4// 2//
+f 6// 3// 4//
+f 7// 4// 5//
+f 7// 6// 4//
+"""
+
+LITTLE_RAST_OBJ = """# landlabgrid
+#
+g landlabgrid
+v 0.0 0.0 0.0
+v 2.0 0.0 0.0
+v 4.0 0.0 0.0
+v 0.0 2.0 0.0
+v 2.0 2.0 1.0
+v 4.0 2.0 0.0
+v 0.0 4.0 0.0
+v 2.0 4.0 0.0
+v 4.0 4.0 0.0
+f 5// 4// 1//
+f 1// 2// 5//
+f 6// 5// 2//
+f 2// 3// 6//
+f 8// 7// 4//
+f 4// 5// 8//
+f 9// 8// 5//
+f 5// 6// 9//
+"""
+
+
+def remove_newline_chars(mystr):
+    mystr = mystr.replace("\n", "")
+    mystr = mystr.replace("\r", "")
+    return mystr
+
+
+def test_write_obj():
+
+    grid = HexModelGrid((3, 2), spacing=2.0)
+    z = grid.add_zeros("topographic__elevation", at="node")
+    z[3] = 1.0
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        fname = os.path.join(tmpdirname, "test_tri.obj")
+        write_obj(fname, grid)
+        file = open(fname)
+        contents = file.read()
+        contents = remove_newline_chars(contents)
+        assert contents == remove_newline_chars(LITTLE_HEX_OBJ)
+
+    grid = RasterModelGrid((3, 3), xy_spacing=2.0)
+    z = grid.add_zeros("topographic__elevation", at="node")
+    z[4] = 1.0
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        fname = os.path.join(tmpdirname, "test_quad.obj")
+        write_obj(fname, grid)
+        file = open(fname)
+        contents = file.read()
+        contents = remove_newline_chars(contents)
+        assert contents == remove_newline_chars(LITTLE_RAST_OBJ)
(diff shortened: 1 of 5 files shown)


gregtucker:
The code includes tests with raster and hex grids, but the tests only verify that the files are created. @mcflugen any suggestions on how to test file contents? The following code creates a test hex and test raster conversion---note that MacOS will actually visualize them directly in finder!

```
from landlab import RasterModelGrid, HexModelGrid
from landlab.io import write_obj

rg = RasterModelGrid((3, 3), xy_spacing=2.0)
z = rg.add_zeros('topographic__elevation', at='node')
z[4] = 1.0
write_obj('test_quad.obj', rg, 'topographic__elevation')
```

File contents:
```
# landlabgrid
#
g landlabgrid
v 0.0 0.0 0.0
v 2.0 0.0 0.0
v 4.0 0.0 0.0
v 0.0 2.0 0.0
v 2.0 2.0 1.0
v 4.0 2.0 0.0
v 0.0 4.0 0.0
v 2.0 4.0 0.0
v 4.0 4.0 0.0
f 5// 4// 1//
f 1// 2// 5//
f 6// 5// 2//
f 2// 3// 6//
f 8// 7// 4//
f 4// 5// 8//
f 9// 8// 5//
f 5// 6// 9//
```

```
hg = RasterModelGrid((3, 2), spacing=2.0)
z = hg.add_zeros('topographic__elevation', at='node')
z[3] = 1.0
write_obj('test_tri.obj', hg, 'topographic__elevation')
```

File contents:

```
# landlabgrid
#
g landlabgrid
v 1.0 0.0 0.0
v 3.0 0.0 0.0
v 0.0 1.732051 0.0
v 2.0 1.732051 1.0
v 4.0 1.732051 0.0
v 1.0 3.464102 0.0
v 3.0 3.464102 0.0
f 4// 1// 2//
f 4// 3// 1//
f 5// 4// 2//
f 6// 3// 4//
f 7// 4// 5//
f 7// 6// 4//
```

gregtucker:
Also tagging @kbarnhart since you might be interested in this functionality for your work.

kbarnhart:
@gregtucker happy to review this. May take a few days since I'll verify that I can open the files in Blender (a good excuse to check it out 👍 ).

Re testing the file contents, my recommendation would be to create an example output file and verify that it is the same as the test result (e.g., compare that reading in the file as a string yields the identical string, either line by line or whole-string). This can sometimes be tricky given windows/linux line endings. However, looking at the Blender docs it looks like its very important for the format to just use \n as the line ending on all platforms https://docs.blender.org/manual/en/2.80/addons/io_scene_obj.html, so probably makes sense to verify that this occurs on windows as part of a test.

gregtucker:
Thanks @kbarnhart. Can anyone point me to a good existing example in our code base of a file-output test?

kbarnhart:
Not off the top of my head. But the best place to look would be in tests/landlab/io. There are a number of example data files in there. Look for usage of the tempfile package since that is what I've used (copying @mcflugen) for making temporary directories/files for this sort of test. 

https://docs.python.org/3/library/tempfile.html

Since this is a flat text file you should just be able two write the output in a temporary file, read it in, read the known truth in and do a straight == comparison.

mcflugen:
@gregtucker Thanks for this! 🎉 I'll review it later today.

mcflugen:
@gregtucker I made a few changes, please have a look and let me know what you think.

Two main changes:
* I changed *write_obj* so that the *path* argument can either be a string or a file-like object. I primarily did this just to make your doctest look a little nicer - no temporary folder and it now displays the contents of the file.
* More tests

I know that *write_obj* writes data to a file and that's why it's under `landlab.io`, but we don't want users to read/write *landlab* grids from/to *OBJ* files as that's not their intended purpose. Since *OBJ* files are really just for visualization, and *netCDF* for loading and saving grids, what do you think of moving *write_obj* under `landlab.plot`? I'm not advocating for one location over another, just thinking out loud. @kbarnhart do you have any thoughts?

kbarnhart:
Only that I've earmarked some time tomorrow afternoon to look closely at this (and test with Blender). So expect comments then!

mcflugen:
Excellent. Thanks @kbarnhart! 😺

tests/io/test_write_obj.py
```diff
@@ -0,0 +1,119 @@
...
+f 6// 5// 2//
+f 2// 3// 6//
+f 8// 7// 4//
+f 4// 5// 8//
+f 9// 8// 5//
+f 5// 6// 9//
+"""
+
+
+def remove_newline_chars(mystr):
```
kbarnhart, line 52: @gregtucker it looks like you are no longer using this. perhaps remove.

kbarnhart (changes requested):
This is great @gregtucker. Super cool functionality and the documentation is great!. 

Three things beyond my very minor comment below. 

1. I think you need to add text to an existing and/or create a new rst file in landlab/docs/source/reference/io in order for your docstrings to manifest in the docs. 

2. Its not entirely clear to me that we are testing we are making files that will work on windows (e.g., that the newline character is always \n instead of \r\n. I **think** that what you are doing is correct b/c of the documentation of the print function [here](https://docs.python.org/3/library/functions.html#print) which indicates that the newline used is \n, but I think it would be worth explicitly testing (e.g., verify the newlines are \n on all ci, which will include windows, and/or open a file on a windows machine). The latter sounds harder to me... 

3. I tried it out and found that the axes might be mis-specified. But a testament to this being usable... it took me no time at all to try and Blender was super easy to install and use. 

```python
from landlab.components import StreamPowerEroder, FlowAccumulator
from landlab import imshow_grid
from landlab.io import write_obj


grid = RasterModelGrid((100, 200), 0.1)
z = grid.add_zeros("topographic__elevation", at="node")

fa = FlowAccumulator(grid)
sp = StreamPowerEroder(grid)

dt= 10000
U = 0.001
for i in range(100):
    z[grid.core_nodes] += dt*U
    fa.run_one_step()
    sp.run_one_step(dt)
    
imshow_grid(grid, z, output="fig.png")
write_obj("test_obj.obj", grid, clobber=True)
```
Yields the expected topography (I made it really small b/c otherwise it was HUGE in Blender. 

fig

But it was not oriented the way I expected in Blender. 
Screenshot from 2020-09-25 16-40-36

kbarnhart:
@mcflugen I think this should stay under landlab.io. While it is a file type that is used for visualization it ultimately is a file that is being written (and theory one could create a reader that could read a .obj into a landlab grid). If we made a vtk writer (an output time primarily for visualization) I'd also argue it belongs in IO. And similarly one _could_ write a reader for it. 

But this is a molehill. 

And I'm definitely a fan of thinking about where things live in the package... :+1:

mcflugen:
@kbarnhart Agreed. *landlab.io* probably makes the most sense. I just wanted to bring it up before we merge this PR and then bring *vtk* back.

mcflugen:
Regarding the orientation issue @kbarnhart points out, it appears that, for the object to appear in the orientation you would expect, the *HexModelGrid* example should produce the following,
```
# landlabgrid
#
g landlabgrid
v 1.0 0.0 0.0
v 3.0 0.0 0.0
v 0.0 0.0 -1.732051
v 2.0 1.0 -1.732051
v 4.0 0.0 -1.732051
v 1.0 0.0 -3.464102
v 3.0 0.0 -3.464102
f 4// 1// 2//
f 4// 3// 1//
f 5// 4// 2//
f 6// 3// 4//
f 7// 4// 5//
f 7// 6// 4//
```
That is, the points are specified as *(x, z, -y)*. That seems so odd that I am probably missing something.

tests/io/test_write_obj.py
gregtucker: done

gregtucker:
Thanks @kbarnhart and @mcflugen. For 3D modeling, I'm not sure the orientation matters, so leaving it as x/y/z for now.

PR #1248, opened 2020-10-08
Siccar point/fix 1219

SiccarPoint:
Addresses #1219. I have some local fails related to netCDF which I suspect are known? New test passes fine.


gregtucker (approved):
Looks great, will merge

Issue #1250, opened 2020-10-21
Missing docs for flow routing/accumulation function


kbarnhart:
I noticed when responding to @AlexLipp on #1249 that the docstrings for the functions located in 

landlab\components\flow_accum\flow_accum_bw.py
and 
landlab\components\flow_accum\flow_accum_to_n.py

are not in the sphinx docs.

kbarnhart:
Fixed. Closing.

Issue #1249, opened 2020-10-20
Accumulating generic fields (not just runoff) using FlowAccumulator


AlexLipp:
Hi all, 

LandLab v1 had a function `find_drainage_area_and_discharge` as part of the FlowAccumulator component which allowed discharge at any point in a catchment to be calculated, given a variable runoff across the model grid. 
A very helpful (possibly unintended) use for this function, which colleagues and I made use of, is that it can be used to accumulate any generic variable that enters a drainage network. For example, if you feed the function (in place of the `runoff` argument) a `bedrock_incision` array, it then outputs the cumulative sediment flux at any point in the landscape (as the `discharge` output of the tuple). This has been a very useful functionality for us!  

However, this specific function has been removed as part of LandLab2. To produce the same results, it is now necessary to instantiate a new FlowAccumulator for every different field you want to accumulate (sediment flux, pollutant etc...), and then run it one step. This involves overwriting a field (`surface_water__discharge`) at every point, which throws a warning (albeit not an error). The major issue with this is that this whole process is around 50x slower than the method using LandLab1. For our purposes we need to able to run this many many times, and as such this reduction in speed is essentially breaking.  

My main request/question is whether it is possible to reintroduce the `find_drainage_area_and_discharge` function that takes variable runoff, but doesn't involve instantiating a new FlowAccumulator everytime. In otherwords, a way to calculate discharge given variable run-off for an already existing drainage network. Alternatively, is it possible to implement a generic `accumulate_field` function which takes as input a raster map of values and then accumulates them along the flowpaths. This is useful, as stated, for calculating cumulative sediment fluxes, but I can imagine uses for any number of things which travel along drainage networks. 
I was going to try defining such a function myself for LandLab2 but, looking at the source code, the original function in LandLab1 is basically just a call to something called `_accumulate_bw` which appears imported from a library called `.cfuncs` which I can't find the source of.  

Additionally, if there is a way to do this using LandLab2 components that I'm not aware of, please let me know! I have hunted around but to no avail...

I'm keen to upgrade fully to LandLab2 because of increased functionality (specifically the better stream profiling) and to stay aligned with the community but I can't find a way to efficiently implement this 'generic accumulator' which is essential for what I'm trying to do. Currently I have LandLab2 and LandLab1 in separate environments for separate, but switching between them is not ideal!

Cheers! 
Alex

kbarnhart:
@AlexLipp this function still exists. Import 
```python
[...]

PR #1251, opened 2020-10-21
add flow accum docs

kbarnhart:
Addresses #1250 

FYI #1249


kbarnhart:
Only fail was a broken link on one of the two docs builds. So am going to merge this.

Issue #1266, opened 2021-02-07
Drainage density of non-raster grid


DavidLitwin:
I was just trying to use the DrainageDensity component on a HexModelGrid, and found that this is currently not possible. I looked into the code and found that `length_of_d8`  is given as an argument to the function `_calc_dists_to_channel`, which will only be applicable to RasterModelGrids. It seems like there should be a simple way to choose either `length_of_link` or `length_of_D8` depending on the grid type and flow director used? Could someone advise on a fix for this? 

Thanks!
David

PR #1268, opened 2021-02-09
use length_of_link or length_of_d8 as appropriate; add test; modify i…

gregtucker:
…nline doc

Addresses #1266


DavidLitwin (commented):
This looks like it should do it, thanks @gregtucker!

DavidLitwin (approved):

PR #1269, opened 2021-02-14
Gt/simple submarine diffuser

gregtucker:
Diffusion-based marine sediment transport model that uses a water-depth-dependent diffusivity, with:
- diffusivity above wave base is uniform (except within tidal range)
- diffusivity below wave base decreases exponentially with water depth
- if specified tidal range is nonzero, diffusivity varies smoothly across mean sea level using a tanh function (so it is about 88% of maximum at the mean low tide level, and 12% of maximum at the mean high tide level)
- above mean sea level (i.e., on land), apart from a small diffusivity associated with tidal effects, diffusivity is arbitrarily tiny (1e-20 m2/y), with this small positive value intended to avoid division by zero in the base class LinearDiffuser.


tests/components/marine_sediment_transport/test_simple_submarine_diffuser.py
```diff
@@ -0,0 +1,146 @@
...
+    assert_array_equal(df, [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 0.0, 1.0, 1.0, 1.0, 1.0, 1.0])
+
+    ssd = SimpleSubmarineDiffuser(grid, tidal_range=2.0)
+    df = ssd.depth_function(depth)
+    assert_array_almost_equal(
+        df, [0.5, 0.5, 0.5, 0.5, 0.5, 0.880797, 0.119203, 0.5, 0.5, 0.5, 0.5, 0.5]
+    )
+
+
+if __name__ == "__main__":
```
mcflugen, line 142: Remove this block?

mcflugen (changes requested):
Looks good @gregtucker! 👍 It looks like there is a stray XCode file that should be removed (*default.profraw*).

tests/components/marine_sediment_transport/test_simple_submarine_diffuser.py
gregtucker: done

gregtucker: and stray default.profraw file removed

mcflugen (approved):

PR #1270, opened 2021-02-15
Use GitHub Actions for CI

mcflugen:
This pull request adds GitHub Actions for our continuous integration tests and will also remove Travis and AppVeyor.

We now have GitHub Actions that check the following:
* code is formatted for *black*
* there is no link using *flake8*
* all unit tests and doctests pass
* all tutorial and teaching notebooks run
* the documentation builds (this is only run if any documentation file has changed)

These tests are run on Linux, Mac, and Windows for Python 3.[678]. As before, all of these tests need to pass before a pull request can be merged.

To get the tests to pass, I had to do the following:
* fix a small error in our shapefile reader. This was possibly caused by a relatively small API change in the *shapefile* package
* remove some lint and reformat the codebase with *black*

Some things still to do (but in a separate pull request):
* deploy to the PyPI using GitHub actions
* build landlab with Python 3.9


mcflugen:
@gregtucker If you could give me a 👍 or 👎 on this, that would be great. Once this is merged, I'll move on to your pull request (#1269).

@kbarnhart There's nothing for you to do here but I thought I would mention you in case you were interested in our continuous integration revamp.

kbarnhart:
@mcflugen - thanks for the heads up. I'll check out what you did, and eventually port similar changes to TerrainBento and Umami... I will probably ask for recommendations from you when I do that.

mcflugen:
@kbarnhart I would be happy to help!

gregtucker (approved):
Looks great!

(There's an Untitled.ipynb file in components folder, is this needed?)

mcflugen:
@gregtucker Thanks! Good catch, that notebook is definitely not needed!

Issue #1278, opened 2021-02-24
Testing major version changes against Terrainbento?


kbarnhart:
Terrainbento's test suite represents an addition set of tests of Landlab. Many of the tests are based on analytical solutions, and sometimes changes in Landlab can unexpectedly break terrainbento. 

For example, many tests started to fail in Terrainbento after #1210 (see #1276, #1277 , and TerrainBento/terrainbento/issues/156)

I'd like to start a discussion about whether it is worth testing against terrainbento for landlab releases. The main reason not to test is that the tests take a while. The main reason to test is that terrainbento has a ton of example use cases, that include tests of Landlab's behavior that extend beyond what is in the standard test suite. 

@mcflugen @gregtucker

gregtucker:
I think it would be a good idea, for releases (not for every Landlab PR). Ideally, anyone who modifies any of the components that TB actually uses would also test against TB. At the moment, the group of people who are likely to do this is probably small enough that we could simply send word to this group that it's a good idea to run the TB tests before merging.

kbarnhart:
@gregtucker agreed that this should only be for releases. @mcflugen and I discussed this briefly the other day when he gave me pointers about updating the terrainbento CI.  In progress at TerrainBento/terrainbento/pull/158.

In theory once I have working github actions for the terrainbento CI this should be pretty simple to implement. 

In reflecting on this a bit more, I think the other reason to do this is that terrainbento probably represents the most extensive collection of tests that landlab components interacting.

Going to leave this open if others feel compelled to discuss.

mcflugen:
I think this makes a lot of sense. We should probably test as many of landlab's [dependent packages](https://github.com/landlab/landlab/network/dependents) as possible. Some others could include:
* DepuitLEM
* Compaction
* Umami
* Sequence
* Plume

I think all of those come with test suites.

We could then have a separate GitHub actions file for each of these that runs on a version tag. A question then would be what version of the dependent package do we test with? The latest development version? The latest release? I think we would probably want to test the latest release, in which case the test would have to checkout the latest version of the dependent package and then run its tests.

Using *terrainbento* as an example, maybe the workflow file could look something like the following?
```yaml
# test-terrainbento.yml
name: Test TerrainBento

on: [push, pull_request]
on:
  push:
    tags:
      - v[0-9]+.[0-9]+.[0-9]+

jobs:
  build-and-test:

    runs-on: ${{ matrix.os }}

    defaults:
      run:
        shell: bash -l {0}

[...]

PR #1277, opened 2021-02-24
remove pit present logic

kbarnhart:
Addresses #1276 

It is not slick in that it doesn't identify a way to decide whether it is necessary to run the depression finder, but it fixes the errors induced in terrainbento.

Issue #1276, opened 2021-02-24
FlowAccumulator needs to run the depression finder even if no pits are present. 


kbarnhart:
Investigating TerrainBento/terrainbento/issues/156 I've found that the divide nodes get labeled as flooded, and thus, when erode_flooded_nodes is set to False, they turn into pinnacles. 

This seemed pretty odd to me, until I looked at the change history of the FlowAccumulator. In #1210 I saw that re-running the depression finder only occurs when pits are present. 

```python
if update_depression_finder and self.pits_present():
                     self._depression_finder.update()
```
This was probably meant to save time, but it means that if there are no pits, but there are flooded nodes, those flooded nodes don't get re-labeled. Because all of the stream power components can turn on/off erosion on flooded nodes depending on user specification, getting those flooded nodes re-labeled is important. Removing `and self.pits_present()` has no (local) impact on Landlab tests. ~Terrainbento tests are in progress (they take 1 hour), but many that were failing are now passing. I suspect they will all pass and will update this in the morning.~ Removing this line completely addresses the issue in Terrainbento (all tests pass with this line modified).

@gregtucker 

MWE 
```python
from landlab import RasterModelGrid
from landlab.components import FlowAccumulator, FastscapeEroder
from landlab.components.depression_finder.lake_mapper import _FLOODED

import matplotlib.pylab as plt

grid = RasterModelGrid((3, 21), 100)
grid.set_closed_boundaries_at_grid_edges(False, True, False, True)
z = grid.add_zeros("topographic__elevation", at="node")

fa = FlowAccumulator(
    grid, 
    depression_finder="DepressionFinderAndRouter", 
    routing='D4')
sp = FastscapeEroder(grid, K_sp=0.001, erode_flooded_nodes=False)

for i in range(10):
    fa.run_one_step()
    sp.run_one_step(2000)
    z[grid.core_nodes] += 10
    
    where_flooded = fa.depression_finder.flood_status == _FLOODED
    
    plt.plot(grid.x_of_node[grid.core_nodes], z[grid.core_nodes])
    plt.plot(grid.x_of_node[where_flooded], z[where_flooded], "ko")
```
<img width="396" alt="Screen Shot 2021-02-23 at 8 52 40 PM" src="https://user-images.githubusercontent.com/6589339/108945014-14ca6300-7619-11eb-8851-abeb8045e4fa.png">

gregtucker:
[...]

PR #1281, opened 2021-03-03
Test the TerrainBento package

mcflugen:
This pull request attempts to add a GitHub action that tests a package that's dependent on *landlab*. In this case, *TerrainBento*.

### DIFF
--- a/.github/workflows/test-terrainbento.yml
+++ b/.github/workflows/test-terrainbento.yml
@@ -1,0 +1,64 @@
+name: Test TerrainBento
+
+on: [push, pull_request]
+
+jobs:
+  build-and-test:
+    # We want to run on external PRs, but not on our own internal PRs as they'll be run
+    # by the push to the branch. Without this if check, checks are duplicated since
+    # internal PRs match both the push and pull_request events.
+    if:
+      github.event_name == 'push' || github.event.pull_request.head.repo.full_name !=
+      github.repository
+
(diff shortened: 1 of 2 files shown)


mcflugen:
@kbarnhart It looks like we should be able to test *terrainbento* as we discussed in #1278. One thing I've noticed, though, is that the *terrainbento* tests won't run on Python 3.9 because of the removal of a deprecated keyword in `nbformat.read`. Since it wasn't doing anything, anyway (at least since Python 3.6 or so), you can read a notebook in *tests/test_all_notebooks.py* with the following (no "encoding" keyword),
```python
nb = nbformat.read(fp.name, nbformat.current_nbformat)
```
I also had to install *jupyter_contrib_nbextensions* before running the tests. I don't see it in any of your requirements files so maybe it used to be installed along with something else? Or maybe there's something else weird going on.

kbarnhart:
@mcflugen I'm planning on updating the terrainbento notebook tests to look like your modification to landlab's notebook tests... will probably get to that sometime next week.

kbarnhart:
@mcflugen this looks great. It is failing as I'd expect because of #1277 / #1276. Seems to me that the order of operations is 

1. Eventually merge #1277 and make a new release of landlab. 
2. I update the terrainbento CI, including improving the notebook tests and make a new terrainbento releas. 
3. This PR should then pass and could be merged.

PR #1285, opened 2021-03-19
Clean up metadata files

mcflugen:
This pull request cleans up some of *landlab*'s metadata files.
* The changelog is now *CHANGES.rst* and is in a standard format that *zest.releaser* understands.
* I've added an *AUTHORS.rst* file to hold a list of all the *landlab* contributors.
* Updated the version and other metadata in *setup.py*
* Removed *versioneer* for handling versions, we'll now use *zest.releaser*.
* The long description (as seen on PyPI) is now constructed from the concatenation of *README.rst*, *AUTHORS.rst*, and *CHANGES.rst*.
* Removed stray files that shouldn't have been in version control.
* Updated the *MANIFEST.in*.


gregtucker (commented):
Looks great!

PR #1291, opened 2021-04-30
Add advanced solver to GroundwaterDupuitPercolator

DavidLitwin:
It has long frustrated me that the adaptive timestep solver for the GDP is both slow and not particularly accurate with time varying recharge. The most concerning area of accuracy for me was that, despite the efforts of the regularization, the solver still often winds up with a water table above the surface, which either has to be tolerated (not ideal) or chopped off to the surface elevation (violating conservation of mass). 

I've been tinkering with the idea that accuracy could be improved by improving the solution to the ODE  that arises in the conservation of mass statement if you assume the groundwater flux term has fixed values. The ODE for thickness is then only dependent on thickness where it appears in the exfiltration term. This is a much simpler problem to solve than the full PDE, as essentially all grid nodes can be treated independently as their own initial value problems. It seems like it may offer some advantages. 

Here I've finally written something that does that, using scipy solve_ivp, which uses a 4th order explicit Runge-Kutta scheme by default. I've written it to take advantage of the vectorized option to solve all IVPs on the nodes in a single call. My initial tests suggest that it is an improvement, at least in some areas, though I need to do more testing. It is slower, but with RK45, not as slow as I was worried it would be. As far as I know, this does not affect the timestep stability conditions (although I increasingly believe that the Von Neumann stability condition is the primary factor that matters here), so this solution is still embedded in the adaptive timestep approach. I've called the new solver "run_with_adaptive_time_step_plus_ivp_solver". Time for a shorter name?

I'm starting this PR to make you all aware that this is happening, and get any input you might have on it. I'm especially interested in ways to speed this up, including the possibility that we could start parallel processing options. 

Known issues:
- I don't have support for areas where regolith thickness is zero here. It might be easy to put this in, as it is in the regular adaptive timestep solver, but I haven't yet.
- Unhandled runtime warnings for various things


DavidLitwin:
Hey @gregtucker or @mcflugen, could you take a look at this when you get a chance? I'm curious what you think of the approach. I've also made some changes to tests, and fixed a typo in the component documentation string. 

I've given up on changing the stability constraints for now. Courant condition doesn't appear to be a limiting factor always - I've exceeded it by 2x and still had stable solutions when von Neumann is met. On the other hand, von Neumann alone doesn't quite do it. In other cases instability features emerge in aquifer thickness around 0.5x the von Neumann condition. The two together seem to work well, though it is probably conservative.

gregtucker:
@DavidLitwin this looks great. I don't have time to do a proper review, but the overall approach seems promising. Other things to consider for performance improvement include cythonizing and using Dask. (Note that we're hoping to weave Dask into some of Landlab's core numerical routines in the not-too-distant future.)

DavidLitwin:
Sounds good, thanks @gregtucker. @mcflugen would you have a chance to look at this? If you do, I would be happy to hear any other suggestions you have about speeding the component up.

DavidLitwin:
Hi all, after discussion with Ciaran, I have decided to use a different approach to improve the GDP solution. More discussion will be over there.

PR #1292, opened 2021-05-06
Add at_layer variables at elements other that cell

mcflugen:
This pull request adds the ability of a user to add layers at elements other than cells. Previously, the *at_layer* variables could only be at cell elements.

To allow layer variables to be defined on all grid elements, I've created a new naming convention: `at_layer_<location>` (*location* is one of *node*, *link*, *patch*, *corner*, *face*, *cell*, or *grid*). This may not be the best syntax but one nice thing about it is that if you split the second part on underscores, you get the dimension of the array (e.g. the dimensions of *at_layer_node* is `("layer", "node")`). I'm open to alternatives, though. This is, for the most part, only used internally so changing things at a later time shouldn't be an issue.


mdpiper:
@mcflugen Just FYI, I've started my review, but I have a bit to go/understand yet.

mdpiper (dismissed):

mdpiper:
The naming scheme is cool!

PR #1300, opened 2021-05-27
Fix xarray 'axis' keyword error in map function

mcflugen:
The pull request fixes an error caused by a new version of *xarray* that raises an exception when a *numpy* ufunc is passed to *xarray* map function. To fix the error, I simply replaced `np.sum` with `xr.Dataset.sum`.


mcflugen:
@nathanlyons Could you please have a quick look at this pull request? With the newest version of *xarray* you can no longer pass *numpy* ufuncs (i.e. *np.sum*, *np.mean*, etc.) to *Dataset.apply* (and *Dataset.map*). Replacing those functions with the equivalent *Dataset* functions seems to fix things.

PR #1301, opened 2021-06-07
fix numpy dep warning related to np.int (replaced with just int)

gregtucker:
Resolve a numpy 1.20 deprecation warning arising in the CTS cellular automaton package: use int instead of np.int

https://numpy.org/devdocs/release/1.20.0-notes.html


gregtucker:
@mdpiper I know you're in the thick of ESPIn but this should be quick, it's just replacing a few np.int's with int's (might even be fun to share with the students as a simple example of the process of PR, review, approval, and merge...)

gregtucker:
Oops, hang on, this does not play nicely with cython (then why did the tests pass?)...

gregtucker:
...confirmed that this is going to take a little more digging before being review-ready...

gregtucker:
This turns out to be stickier than it looks, requiring a deeper dive than I have time for in the near future. Closing.

PR #1302, opened 2021-06-13
Improved solution for GroundwaterDupuitPercolator

DavidLitwin:
As I referenced in #1291, there have been some outstanding problems with the existing solvers for the GDP. As Ciaran pointed out, the solution to the ODE that I was passing to the scipy ivp_solver in that previous PR (now closed) actually has a (piecewise) analytical solution. This new PR uses this improved solution, implemented in the existing `run_one_step` and `run_step_with_adaptive_time_step_solver` methods.


DavidLitwin:
I've also done a few other things here:

- Clarified some documentation, including fixing an outstanding error in an equation in the docstring
- Removed and/or changed some indexing that was redundant and possibly slowing things down
- Added a `where` statement to the division operations to eliminate almost all divide by zero errors.
- Cleaned up miscellaneous things in the component tests
Like the previous PR, the solution here doesn't handle cases where the regolith thickness is zero. This could probably be changed in the future if someone found that to be a problem. It's not currently standing in my way, and would likely require several additional calculations. 

Still to do is to improve component speed in a meaningful way. I'm still exploring some different options.

gregtucker:
@DavidLitwin are you ready for reviews now, or still tweaking?

DavidLitwin:
Hi @gregtucker, I think the version  in this PR is pretty much ready for review. I've checked that it still passes as the gdp tests. Speed is very similar to the original adaptive timestep solver method that this updates. 

At the same time, I'm working on a version of this branch over on the branch `dlitwin/gdp_analytical_cython` where the whole `while` loop that iterates through substeps is written in cython. This is a bit of a headache because in that loop I calculate gradients, flux divergence, and use one of the mappers. However, I think it may be worthwhile, given that all the assignments and communication with the grid that happens in that loop is unnecessary overhead -- that info is never accessed, unless a callback function is used. I'm going to keep tinkering with that branch, but it might be a while before I have anything that works nicely. If you have any suggestions, I would also love to talk about this!

gregtucker (dismissed):
Looks good to me

mcflugen (approved):
👍

kbarnhart:
nice work @DavidLitwin !!

DavidLitwin:
@mcflugen, @gregtucker, @kbarnhart, thank you! This has been working great for me over the last few weeks of using it. I'm excited to have it out, and I think it's a real improvement! I'll follow up later on speed improvements if they seem necessary.

PR #1306, opened 2021-06-27
Gt/make ttl depth dependent taylor diffuser

gregtucker:
This PR adds a tutorial for the DepthDependentTaylorDiffuser component. In making it, I discovered an error in the math: the `linear_diffusivity` parameter is actually NOT a diffusivity as the documentation indicates, but rather a transport velocity. The diffusivity is the product of this transport velocity and the characteristic depth scale. The PR therefore includes the following changes:

- The `linear_diffusivity` parameter is deprecated in favor of a new parameter, `soil_transport_velocity`. If a user passes `linear_diffusivity`, they will get a deprecation warning but the component will use the `linear_diffusivity` in place of `soil_transport_velocity`.
- The math in the docstring has been corrected.


gregtucker:
@BCampforts can you review this PR?

BCampforts (approved):
Checked the math and the plots in the depth_dependent_taylor_diffuser notebook. 
Everything seems to be sound. 
One warning on the use of the colormap (Block 19): 
"MatplotlibDeprecationWarning: You are modifying the state of a globally registered colormap. This has been deprecated since 3.3 and in 3.6, you will not be able to modify a registered colormap in-place. To remove this warning, you can make a copy of the colormap first. cmap = mpl.cm.get_cmap("pink").copy()"

BCampforts (approved):
Checked the math and the plots in the depth_dependent_taylor_diffuser notebook.
Everything seems to be sound.
One Matplotlib Deprecation Warning on the use of the colormap (Block 19).

PR #1311, opened 2021-07-06
Small correction to GDP

DavidLitwin:
Accidentally left off an absolute value in the Courant condition.


SiccarPoint (approved):
Thumbs up

Issue #1320, opened 2021-08-02
>1.6 million warnings is too many


SiccarPoint:
Most of this is np.int and np.float warnings. Easy fixes.

mcflugen:
Closed with #1323.

PR #1323, opened 2021-08-02
Siccar point/suppress warnings

SiccarPoint:
Suppresses a huge number of warnings related to unnecessary use of numpy number types, addressing #1320


mdpiper (approved):

mdpiper:
@SiccarPoint `black` is implacable.

Thank you for making these changes. I think we saw some of these in our ESPIn summer school in June, and people who were newer to Python (and hadn't learned about the difference between warnings and exceptions) were concerned. I, of course, ignored them, and thereby missed a teaching moment where I should've stopped and filled out an issue.

PR #1335, opened 2021-09-07
Add KinematicWaveRengers back to components

kbarnhart:
In 2016 KinematicWaveRengers got deleted. This PR brings it back. 

I did all the things necessary to bring it in LL2.0 compliance (field names, private variables, use of super(), etc).

I brought it and two other missing kinematic wave components into the documentation.

I did a few other things to clean up existing issues that I observed when I built the docs. 

The existing test passes. 

tag: @frengers

PR #1336, opened 2021-09-08
docs: add linkcheck to docs testing

### LINKED ISSUE #1334: Run documentation build check on PRs
@mcflugen it doesn't look like we are enforcing documentation checks on all PRs...

Maybe there is something I'm missing in the github action logic, but I don't see it in some of the more recent PRs.

I don't think there are too many issues that have accumulated (I've just addressed a few in a forthcoming PR I'm making associated with an independent thing)... but I do think it makes sense to enforce this (in part because I know what it takes to get the docs in line if they get out of hand).

mcflugen:
This pull request is intended to fix #1334. In the move from Travis to GitHub Actions for our CI, we turned off *linkcheck* when testing our documentation build. It may be we had a good reason for doing so but I've turned it back on here to see if we can get it to work again.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,22 +1,6 @@
 name: Documentation
 
-on:
-    push:
-        paths:
-            - "docs/**"
-            - "AUTHORS.rst"
-            - "CHANGES.rst"
-            - "CONTRIBUTING.rst"
-            - "LICENSE.rst"
-            - "README.rst"
-    pull_request:
-        paths:
-            - "docs/**"
-            - "AUTHORS.rst"
-            - "CHANGES.rst"
-            - "CONTRIBUTING.rst"
-            - "LICENSE.rst"
-            - "README.rst"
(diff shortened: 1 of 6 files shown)


kbarnhart:
@mcflugen excellent! FYI, many of these doc warnings as errors should have been addressed in #1335.

kbarnhart (approved):
@mcflugen this looks great!

mcflugen:
@kbarnhart Thanks for having a look at this.

As I mentioned in #1334, I wanted to change our CI so that if *only* documentation files (i.e. stuff in the *docs* folder, *README.rst*, etc.) are modified, then the regular unit tests and notebook tests are not triggered. I was able to make that happen but that meant those tests never "passed" and, because we added a branch protection rule that requires those checks to pass before merging, pull requests wouldn't be accepted. Anyway, although it might be a pain to wait for all the tests to pass after just fixing a typo in the readme, that's just the way it is for now. 🤷‍♂️

PR #1342, opened 2021-09-16
maint: add a CITATION.cff file

### LINKED ISSUE #1328: New Github citing system
Add a CITATION.cff to help people cite Landlab through Github, per:

https://www.software.ac.uk/news/research-software-now-easy-cite-github

mcflugen:
This pull request adds a citation file in [citation file format](https://citation-file-format.github.io/) for the latest *landlab* release (*v2.0.1*). I used the author list (including order) from what is on [Zenodo](https://doi.org/10.5281/zenodo.3776837). I also updated our DOI badge in the README file.

Closes #1328.


CITATION.cff
```diff
@@ -0,0 +1,18 @@
+# YAML 1.2
+---
+authors:
```
gregtucker, line 3: Only two authors seem to be listed, yet there are many more contributors. Is this a limitation of the .cff format?

gregtucker (commented):
@mcflugen great to see this CITATION.cff file -- that was on my to-do list but you beat me to it! See my question about # of contributors. I'm also wondering how we can acknowledge folks who have not necessarily been actively contributing to the code/docs base, but have contributed in other ways (e.g., to grants). Another question is whether we can include in the CITATION.cff one or more published articles, such as Hobley et al. (2017), Barnhart et al. (2020), or the pending Tucker et al. (2021) GMD?

CITATION.cff
mcflugen: Oops! I must have committed my changes before hitting the save button. Yes, there are many more contributors than two. I pushed a version that includes the entire author list as it appears on Zenodo.

mcflugen:
> I'm also wondering how we can acknowledge folks who have not necessarily been actively contributing to the code/docs base, but have contributed in other ways (e.g., to grants).

I think we can add anyone we like to the author list in the citation file. For now I just used the author list from Zenodo, which, I believe, is based on number of commits to the repository. 

> Another question is whether we can include in the CITATION.cff one or more published articles, such as Hobley et al. (2017), Barnhart et al. (2020), or the pending Tucker et al. (2021) GMD?

My understanding is the CITATION.cff file is being used by GitHub to just provide a citation to the source code. I suppose, however, we could add another citation file that contains other references. The citation file format is just *yaml* so I think we could store multiple references in a single file by separating them with the `---` document separator. I'm not sure what would happen if we included multiple documents in CITATION.cff.

mcflugen:
@gregtucker I've removed all version-specific metadata from the citation file (i.e. version number, release date, etc.) and added a DOI that *Zenodo* provides for any *landlab* version and will always point to the most recent version. This should ensure that anyone citing the *landlab* repository will be using the same citation.

This is what *eWaterCycle* did with their citation file and, since *eScience Center* contributed to the citation file format, I figured they would have figured out the best way to deal with this.

gregtucker:
Do you know how eWaterCycle handles the case when the authorship changes over time? For example, a CITATION.cff is created for version 1.0 (but it doesn't say anything about a version), with Abdul and Baker listed as authors. Then some time later, a new doi is generated for version 2.0 and now Abdul, Baker, and Catwoman are authors, and CITATION.cff is updated. Does that mean the first doi and second doi are different citations, even though they both refer to the software without a specific version number? Do we need to worry about people citing the older Abdul & Baker instead of the latest? I guess these are questions to be resolved by the community over the next few years... Would it make sense to include something like "Version 2.x" in the metadata, just to flag that it is anticipated that a different citation might apply to a future version 3.x?

mcflugen:
@gregtucker I'm not sure, but it appears they add authors to the citation file as contributions come in, and is what I was thinking we would do. This means, although the DOI stays the same, the actual citation will be slightly different since the list of authors will include the new contributors. Even before a new release, and certainly before *landlab* 3.x, the citation will change (but still point to the same DOI).

If someone cites the code, they are free to add information to the citation beyond what's in our citation file—e.g. the version of *landlab* they used. After thinking about this a little more, however, my feeling is now that it is better practice to describe the *landlab* version in the text (or supplemental docs) rather than through the reference.

If we keep the citation file up-to-date, I hope  we can reduce the chance of someone using an old reference (i.e. Abdul and Baker, rather that Abdul, Baker, and Catwoman) but that relies on people getting the citation from GitHub or the citation file rather than must copying and pasting it from another article that cites *landlab*.

kbarnhart:
@mcflugen forgive me for a naive question that is somewhat late to the game. 

It seems like this file is creating yet another citation record, when what makes more sense is to point towards an existing citation. At present we have a few different possible citations, depending on what version of landlab is being used. Namely citing either Hobley et al. (2017) or both Hobley and Barnhart et al. (2020). In addition there are citations that exist for each zenodo archive. 

However, our documentation makes pretty clear how to cite landlab, and what is in this .cff file is not at all in line with what we recommend. Why does the .cff not Why does the .cff not just provide those citations? It could point to the two papers and the "all-versions" Zenodo record?

It appears to be possible: https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#credit-redirection

kbarnhart:
I think the way to accomplish the goals would be to have three entries in the .cff

1. Hobley paper
2. Barnhart paper
3. Zenodo archive <-- this would be the entry that gets updated with new contributors (as you've been implementing in this PR)

gregtucker:
@kbarnhart I agree: seems reasonable to ask people to cite the foundational papers plus the actual source code as indicated in the CITATION.cff. Because the CITATION.cff file will contain the full list of code contributors at any given time, and the papers ideally include those who contributed in other "foundational" ways (e.g., getting the initial grants, shaping the concepts, guiding the development, etc.), citing both of these things allows everyone's contributions should be acknowledged.

The idea of keeping the same doi in perpetuity but allowing the author list to evolve over time has the virtue of simplicity for us as developers, because we need only keep the CITATION.cff file up to date. We also wouldn't have to worry about splitting citations (e.g., instead of 20 citations of version 2 and 30 citations of version 3, etc., you just have 50 citations to the code base generally).

But I'm wondering how would this interact with traditional journal citation practice of Author(s), Year, Title, etc. What would the publication date be if the doi does not refer to any specific version?

Also, much as I'd hate to make life easier for bean counters, should we worry about fairness if new contributors could in theory be added after 1,000 citations have already accumulated based on prior work that they did not contribute to?

I guess the broader question is what does it mean to cite something that is growing and evolving? One precedent in the world of books is to cite the particular edition, according to its year of publication. That model is sort of akin to assigning different doi's to different versions. But of course software versions generally come out much faster than new book additions.

PR #1343, opened 2021-09-20
docs: fix unexpected section in green ampt component

mcflugen:
This pull request fixes the docstring for *SoilInfiltrationGreenAmpt* by removing a non-standard heading that caused an error building the docs.


mcflugen:
The *coveralls* website is down, which is causing our tests to fail. I'll restart the tests once they are back up but it looks like the documentation is building just fine so we should be good.

gregtucker (approved):
Looks good, thanks!

Issue #1352, opened 2021-10-09
lateral_erosion component not resetting and division by zero error


josh-wolpert:
Hello,

I'm using the lateral_erosion component, and I've noticed pits developing after a node meets the condition for a lateral erosion event. I think self._dzlat needs to be reset each time run_one_step is called. Since self._dzlat doesn't reset, once a node experiences lateral erosion, it will continue to erode by lateral erosion at every subsequent timestep. Below is the relevant code from the component (lines 520-526):

                    if vol_lat[lat_node] >= voldiff:
                         self._dzlat[lat_node] = z[flowdirs[i]] - z[lat_node]  # -0.001
                        # after the lateral node is eroded, reset its volume eroded to
                        # zero
                        vol_lat[lat_node] = 0.0
        # combine vertical and lateral erosion.
        dz = dzdt + self._dzlat

Even when the condition for lateral erosion is not met (line 520), a node continues to erode by lateral erosion (line 526).

I've also noticed division by zero errors coming from the angle_finder method in node_finder.py. I think this happens when current nodes (cn) have the same x or y coordinates as their donor (dn) or receiver (rn) nodes.

I've made minor modifications that I'll submit for your consideration in a pull request.

Best,
Josh

gregtucker:
@josh-wolpert thank you for posting this issue and submitting a PR that fixes it!

PR #1353, opened 2021-10-09
Fix pits and division by zero in lateral_erosion component

josh-wolpert:
Hello,

I have a couple of minor edits to the lateral_erosion component for your consideration. These were first brought up in issue #1352.

The first edit's in lateral_erosion.py. At the start of the run_one_step methods, I reset self._dzlat to zeros. This prevents a node from eroding by lateral erosion at every time step after lateral erosion occurs once. Without this, I've noticed pits developing in my runs, which I believe are related to this attribute not resetting.

The second edit is in node_finder.py and involves the angle_finder method. I've made a minor modification to allow for current nodes (cn) to have the same x or y coordinates as their donor (dn) or receiver (rn) nodes without a division by zero error occurring.

Hopefully this is of use!
Josh


mcflugen:
@josh-wolpert Thanks for this! I've made a couple changes to your pull request.
* Use `np.fill` to zero-out the *_dzlat* array. This is probably a little faster because it eliminates a floating point operation and it properly handles *nan*s and *inf*s (a *nan* times 0.0 is still a *nan*).
* Use `np.isclose` for the floating point comparisons.
* Rewrote *angle_finder*. I think the old one had a bug where it returned 135° instead of 45° for some combination of nodes. The new one also eliminates the divide-by-zero you discovered, and it's more general in that it can take arrays of points (see the doctoring for an example).
* added unit tests for *angle_finder*

PR #1355, opened 2021-10-12
tests: check notebooks are clean and blackened

mcflugen:
This pull request adds a script to check that *landlab* notebooks are clean—no output, no execution counts. I've also added this check to the pre-commit file, along with a check to see if notebooks are formatted as per *black*. I haven't added these checks to our CI yet.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -164,20 +164,37 @@ of Earth-surface dynamics, Earth Surface Dynamics, 5, p 21-46,
 Are there any examples of using Landlab I can look at?
 ------------------------------------------------------
 
-The Landlab package contains a directory at ``landlab/notebooks`` which contains
-Jupyter notebooks describe core concepts and give examples of using components.
+The Landlab package contains a directory, ``landlab/notebooks``, with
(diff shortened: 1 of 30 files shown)


gregtucker (commented):
This looks good to me overall. It's great to test for clean notebooks, since the notebooks often have more complex test-like examples than the unit tests.

I'm not sure about enforcing black though. Consistency in code formatting is great, but isn't there a risk that enforcing it will dissuade people from contributing notebooks? I don't think black-ifying is as simple as running black on an .ipynb file, or at least when I tried it, it corrupted the notebook. I would propose we try to *use* black-style formatting as much as possible in the notebooks (as an example that others can follow), but that we don't *enforce* it in CI. There may also be cases where there's a good reason, e.g. for readability, to deviate from black style (for example, if you wanted to define an array that is 1D but actually represents 2D data, so you might display the numbers as if they were 2D).

gregtucker (approved):
Hooray for notebook testing and cleanup!

PR #1356, opened 2021-10-26
fix: clip active layer thickness to zero

mcflugen:
This is a small pull request that eliminates an `invalid value encountered in power` warning in the `NetworkSedimentTransporter`. The warning occurs when calculating the active layer thickness for small values of *taustar*. The offending expression is,
```python
(taustar - 0.0549) ** 0.56
```
The fix is to just clip the base to have a lower bound of zero, which results in an active layer thickness of zero for small *taustar*.


gregtucker (dismissed):
This looks like a sensible modification to guard against negative excess Shields stress and consequently taking a fractional root of a negative number.

pfeiffea:
This looks great to me, @mcflugen . Thanks for tidying things up!

PR #1357, opened 2021-10-26
feat: infer data types of shapefile fields

mcflugen:
Sometimes when reading a shapefile, the fields contained in that shapefile are mis-typed—usually arrays that are of type `object` when they should really be *float* or *int*, for example. This happens when a field contains the value `None` and then gets transformed into a *numpy* array, which is given a *dtype* of `object`. A simple example that demonstrates this behavior is,
```python
>>> x = np.array([None, 1, 2, 3])
>>> x.dtype
array([None, 1, 2, 3], dtype=object)
>>> y = x[1:]
>>> y
array([1, 2, 3], dtype=object)
```
Even after the `None` element has been removed, the array is still of type `object`.

This pull request adds a function that infers the data type of the array from an array's elements. The user can, as before, override this behavior using the *link_field_conversion* and *node_field_conversion* keywords.


mdpiper (approved):
The *hypothesis* package and the idea of property-based testing are both neat and new to me. I enjoyed learning about the slightly different approach they take for testing.

My one concern was that having *hypothesis* "run the test function over a wide range of matching data" would lead to slower tests; however, all the tests in **test_infer_dtype.py** took only 6 s to run on my ancient laptop.

Another consideration (concern is too strong a word) is that *hypothesis* adds another layer of abstraction to testing. For all the people (OK, people like me) who are late to reading about the decorators available with *pytest*, having additional decorators from *hypothesis* make tests harder to read. The solution is, of course, to rtfd. But we shouldn't expect Landlab contributors to write tests of such complexity.

gregtucker:
Maybe one solution to @mdpiper's point about complexity is to use *hypothesis* relatively sparingly, for Landlab core routines, but generally avoid using it for components (whose code is more likely to be read or modified by users)?

Issue #1358, opened 2021-11-03
Including field units with create_grid


alexmitchell:
I can't figure out how to include field units when creating a grid from a dictionary using landlab's `create_grid` function. Here is the minimum code to show the issue:
```python
from landlab import create_grid
params = {
        "RasterModelGrid": [
            {
                "shape": (5,7), "xy_spacing": (10,10),
            },
            {
                "fields": {
                    "node": {
                        "topographic__elevation": {
                            "constant": [{"value": 0.0}],
                        },
                    }
                },
            },
        ],
}

grid = create_grid(params)
print(grid.at_node.units)
```
The output will be:
`{'topographic__elevation': '?'}`

I want to add information about the units (e.g. `"units" : "m"`) somewhere within the `"topographic__elevation"` dictionary. However the `"topographic__elevation"` dictionary only accepts synthetic field functions and they do not accept units either.

I can manually set the units of a field after creating the grid, but it does not behave as expected:
```python
f_name = 'topographic__elevation'
print("Before:")
print(grid.at_node.units[f_name])
print(grid.field_units('node',f_name))

grid.at_node.units[f_name] = 'm'

print("After:")
print(grid.at_node.units[f_name])
print(grid.field_units('node',f_name))
```
Produces:
```
Before:
?
?
After:
m
?
```

Is there a "best way" to assign units to fields when creating grids from dictionaries or after a field already exists? If not, I think it might make the most sense to modify the `add_fields_from_dict` or `add_field_from_function` methods in landlab/grid/create.py to accept units. In other words, I would like the following to be a valid input dictionary:
```python
params = {
        "RasterModelGrid": [
            {
                "shape": (5,7), "xy_spacing": (10,10),
            },
            {
                "fields": {
                    "node": {
                        "topographic__elevation": {
                            "constant": [{"value": 0.0}],
                            "units" : 'm',
                        },
                    }
                },
            },
        ],
}
```

mcflugen:
@alexmitchell Thank you for this. You are correct, you cannot set a field's units when using the `create_grid` function. I believe this is just an oversight and it should be easy to add that functionality. Using your example, I think the dictionary would look something like the following,
```python
params = {
    "RasterModelGrid": [
        {
            "shape": (5, 7),
            "xy_spacing": (10, 10),
        },
        {
            "fields": {
                "node": {
                    "topographic__elevation": {
                        "constant": [{"value": 0.0, "units": "m"}]
                    },
                }
            },
        },
    ],
}
[...]

PR #1359, opened 2021-11-08
feat: add field units with create_grid (#1358)

mcflugen:
This pull request addresses #1358 by adding the ability to set a field's units when creating a grid using the *create_grid* function. I've also added the ability to change a field's units after it has been added to a grid.

A field's units can now be specified in the dictionary of grid parameters passed to *create_grid*. For example,
```python
>>> params = {
...         "RasterModelGrid": [
...             {
...                 "shape": (5, 7),
...                 "xy_spacing": (10, 10),
...             },
...             {
...                 "fields": {
...                     "node": {
...                         "topographic__elevation": {
...                             "constant": [{"value": 0.0}],
...                             "units": {"units": "meters"},
...                         },
...                     }
...                 },
...             },
...         ],
...     }
>>> grid = create_grid(params)
>>> grid.at_node.units["topographic__elevation"]
'meters'
```
I've implemented this by adding a new synthetic values function, *units*, which adds a new field to a grid if it doesn't already exist and then, instead of setting its values, sets its units.

You can also now change the units of a field after it has been created,
```python
>>> _ = grid.add_zeros("z", at="node")
>>> grid.at_node.units["z"]
'?'
>>> grid.at_node.set_units("z", "meters")
>>> grid.at_node.units["z"]
'meters'
>>> grid.field_units("node", "z")
'meters'
```

@kbarnhart If you have time, I would like to hear your thoughts on how I've implemented this. I've extended the idea of what a synthetic values function is to now just being a function that adds a field to a grid (if it doesn't already exist) and then alters that field in some way (that may not alter its values at all). Do you think this makes sense?

Some other options:
---
Option 1
```python
...                 "fields": {
...                     "node": {
...                         "topographic__elevation": {
...                             "constant": [{"value": 0.0, "units": "meters"}],
...                         },
...                     }
...                 },
```
In this case, every synthetic values function would take an additional keyword, *units*, that would be passed to the *add_field* function.

---
Option 2
```python
...                 "fields": {
...                     "node": {
...                         "topographic__elevation": {
...                             "constant": [{"value": 0.0}],
...                             "attrs": {"units": "meters"},
...                         },
...                     }
...                 },
```
In this case, there would be a new synthetic values function, *attrs*, that sets attributes of a field. This would act like the *units* function I've created in this pull request but could, in the future, be extended to set other field attributes. Right now, other than *units*, we don't keep track of any other field attributes but maybe in the future we could?


mcflugen:
@kbarnhart Would you be able to have a quick look at this?

kbarnhart:
@mcflugen sorry to have forgotten to give feedback on this. I like your plan on how to implement it (as compared with your other options). I would actively oppose option 1. I see the benefit of option 2 (and has parallels in netcdf construction). but I think it adds too many layers. 

Only thing that I suspect needs doing is documenting this in the create_grid docstring (noting whether units are optional, which I think they should be).

mcflugen:
@kbarnhart Thanks! I appreciate you taking the time to have a look at this! I'll update the docs to reflect these changes.

Issue #1239, opened 2020-09-14
feature inquiry: creating a NetworkModelGrid from a RasterModelGrid


shelbyahrendt:
I am interested in creating a NetworkModelGrid from a RasterModelGrid for two purposes: 

1) To make NetworkModelGrid creation for physical locations easier
2) To co-locate RasterModelGrid and NetworkModelGrid nodes for future coupling of the NetworkSedimentTransporter with other Landlab components

I noticed @kbarnhart opened issue #853 on a related topic and I am following-up on on the status of this idea for a tool and/or whether an effort is "in-the-works"? I have investigated implementing this conversion using the ChannelProfiler on a RasterModelGrid, with the idea that 'xy' node locations and 'nodes_at_link' attributes could be added to the existing data-structure output by the channel profiler and then used to instantiate a NetworkModelGrid. 

I've started a crude, 'ad hoc' version of this idea using the Hugo Watershed site from the OverlandFlow landlab tutorials [here](https://github.com/shelbyahrendt/landlab/blob/shelby_dev/notebooks/tutorials/network_sediment_transporter/create_network_from_raster.ipynb): However, I've had difficulty grappling with an efficient way to create proper 'nodes_at_link' references from the ChannelProfiler data structure for the NetworkModelGrid. In the [code ](https://landlab.readthedocs.io/en/master/_modules/landlab/io/shapefile/read_shapefile.html) creating a NetworkModelGrid from a shapefile, I noticed the ordering of nodes is from bottom left to top right, however, within the ChannelProfiler datastructure it seems intuitive to order nodes from watershed outlet to headwater channels since that is the ordering of the extracted channel segments. 

@kbarnhart and @mcflugen do you have any tips on this issue or could you point me towards existing 'in-the-works' code for such tools?

Issue #853, opened 2018-12-19
Network Model Grid mappers


kbarnhart:
In discussions related to Network Model Grid (mostly #627) we described wanting functions to map from a network model grid to a raster/hex/voronoi model grid. 

This is beyond the scope of the current NMG pull request (#848) so I'm making a unique issue.

PR #1360, opened 2021-11-10
feat: create network grid from raster grid functionality

shelbyahrendt:
This PR addresses issues #1239 by adding a function `create_network_from_raster` that will delineate a NetworkModelGrid from a given RasterModelGrid with a channel network. It also partially addresses #853 by co-locating NetworkModelGrid nodes with RasterModelGrid nodes and storing the corresponding ids of these nodes as attributes on the NetworkModelGrid for coupling of node properties between grids.

There are currently two ways to initiate the desired NetworkModelGrid node grid-spacing:

1. By **a user specified distance between nodes** (must be larger than the RasterModelGrid resolution)
2. By a **dynamically calculated node-spacing based upon some number of local channel widths** (channel width approximated by a drainage-area relationship) The goal of this setting is to somehow ensure that the NetworkSedimentTransporter will be stable on the grid that is generated--preliminary model testing by Muneer Ahammad suggests that node-spacing less than two local channel widths helps stability of NetworkSedimentTransporter.

The PR includes files: `landlab/landlab/grid/create_network.py` which contains the functionality and `landlab/notebooks/tutorials/network_sediment_transporter/create_networkgrid_from_rastergrid.ipynb` which demonstrates the function applied to an example DEM.

The `create_network_from_raster` function is still in progress and requires both refactoring and testing.


shelbyahrendt:
@mcflugen here is the initial PR. Let me know if you recommend any strategies for refactoring/improvements.

mcflugen:
The notebook tests are failing because our secret OpenTopography API key cannot be shared with forks, which this is.

mcflugen:
@shelbyahrendt Thanks for your help with this! 🎉 Since the only failing test is because we're working from a fork, I'm going to merge this. As we discussed, if you would like to add additional notebooks, we can do that in a separate pull request.

shelbyahrendt:
Great! Thanks

PR #1362, opened 2021-11-17
Bc/bedrock_landslider

BCampforts:
This PR  adds a new component, tests and a notebook tutorial to simulate deep-seated bedrock landsliding in Landlab. The BedrockLandslider component calculates the location and magnitude of episodic  bedrock landsliding.


landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,692 @@
...
+            slides = np.random.rand(p.size) < p
+
+            # Now, find the critical node, which is the receiver of i_slide
+            # Critical nodes must be unique (a given node can have more receivers...)
+            i_slide = np.unique(
+                self.grid.at_node["flow__receiver_node"][np.where(slides)]
+            )
+            # Remove boundary nodes
+            if not self.landslides_on_boundary_nodes:
+                i_slide = i_slide[(self.grid.node_is_boundary(i_slide) is False)]
```
mcflugen, line 446: I'm not sure this line does what you want it to do (it doesn't look like it's tested). I think what you want is the following?
```python
i_slide = i_slide[~self.grid.node_is_boundary(i_slide)]
```

landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,692 @@
...
+        else:
+            i_slide = np.array(self.critical_sliding_nodes)
+
+        # output variables
+        suspended_Sed = np.float64(0)
```
mcflugen, line 451: Why do you use *np.float64*? Why not just,
```python
suspended_Sed = 0.0
```

landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,692 @@
...
+            all_iP_el = topo[nb_up]
+            s_slide_all = (all_iP_el - cP_el) / distToIni_all
+
+            nb_up = nb_up[s_slide_all > self._angle_int_frict]
+            s_slide_all = s_slide_all[s_slide_all > self._angle_int_frict]
+
+            if s_slide_all.size > 0:
+                s_slide = max(s_slide_all)
+                storeV_bed = np.float64(0.0)
+                storeV_sed = np.float64(0.0)
```
mcflugen, line 489: As before, why not just,
```python
storeV_bed = 0.0
storeV_sed = 0.0
```

landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,692 @@
...
+                upstream = 0
+                uP = nb_up
+                if not self.landslides_on_boundary_nodes:
+                    uP = uP[(self.grid.node_is_boundary(uP) is False)]
```
mcflugen, line 493: I see that this line is tested but it seems like it should be,
```python
uP = uP[~self.grid.node_is_boundary(uP)]
```

landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,692 @@
...
+                # Fix sliding angle of particular LS
+                ang_sl = np.float64((self._angle_int_frict + s_slide) / 2)
```
mcflugen, line 495: 🤔 Maybe I'm missing something but, again, why not,
```python
ang_sl = (self._angle_int_frict + s_slide) / 2.0
```

landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,692 @@
...
+                stall = 0
+
+                while uP.size > 0 & (
+                    upstream <= self._max_pixelsize_landslide and stall < 1e4
+                ):
```
mcflugen, line 500: It's probably better to use a logical *and* here (rather than the bitwise operator, *&*)
```python
while uP.size > 0 and upstream <= self._max_pixelsize_landslide and stall < 1e4):
```

landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,692 @@
...
+                            (
+                                self.grid.active_adjacent_nodes_at_node[uP[0]],
+                                self.grid.diagonal_adjacent_nodes_at_node[uP[0]],
+                            )
+                        )
+                        nb = nb[nb != -1]
+                        nb_up = nb[topo[nb] > cP_el]
+                        uP = [*uP, *nb_up]
+
+                        temp, idx = np.unique(uP, "first")
```
mcflugen, line 537: I'm not sure why you are using the *"first"*. I think maybe you want,
```python
temp, idx = np.unique(uP, return_index=True)
```

landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,692 @@
...
+                        uP = np.array(uP)
+                        uP = uP[np.sort(idx)]
+                        if not self.landslides_on_boundary_nodes:
+                            uP = uP[(self.grid.node_is_boundary(uP) is False)]
```
mcflugen, line 541: As before,
```python
uP = uP[~self.grid.node_is_boundary(uP)]
```

landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,692 @@
...
+                    np.matlib.square(np.matlib.divide(slope, self._angle_int_frict)),
+                    0.999,
+                )
+            ),
+        )
+        Qs_out = np.zeros(z.shape)
+        dH_Hill = np.zeros(z.shape)
+        H_i_temp = np.array(z)
+        max_D = np.zeros(z.shape)
+        max_dH = np.ones(z.shape) + np.inf
```
mcflugen, line 633: If this is *inf*, can we just remove it?

mcflugen (changes requested):
Looks good so far. One general comment: if possible, it would be great if you could use [PEP8](https://www.python.org/dev/peps/pep-0008/) naming conventions for variables (i.e. descriptive names that are lowercase, with words separated by underscores).

landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,692 @@
...
+
+        if store_landslides_volume_sed:
+            self.landslides_volume_sed = []
+        else:
+            self.landslides_volume_sed = None
+
+        if store_landslides_volume_bed:
+            self.landslides_volume_bed = []
+        else:
+            self.landslides_volume_bed = None
```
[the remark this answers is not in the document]
BCampforts: For the landslides_size, landslides_volume, ... I want the users to be able to access them after the run. Should I write a getter for those?

landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,692 @@
...
+        bed = self.grid.at_node["bedrock__elevation"]
+        ss = self.grid.at_node["topographic__steepest_slope"]
+        soil_d = self.grid.at_node["soil__depth"]
+        ls_sed_in = self.grid.at_node["landslide_sediment_point_source"]
+        lb_bed_ero = self.grid.at_node["landslide__bed_erosion"]
+
+        # Reset LS Plains
+        lb_bed_ero[:] = np.zeros(self.grid.number_of_nodes)
+        # Reset landslide sediment point source field
+        ls_sed_in[:] = np.zeros(self.grid.number_of_nodes)
```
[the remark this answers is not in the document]
BCampforts: Done

landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,692 @@
...
+            H_el[H_el > MAX_HEIGHT_SLOPE] = MAX_HEIGHT_SLOPE
+
+            sc_rad = np.arctan(self._angle_int_frict)
+            Hc = (4 * self._cohesion_eff / (self._grav * self._rho_r)) * (
+                (np.sin(np.arctan(ss)) * np.cos(sc_rad))
+                / (1 - np.cos(np.arctan(ss) - sc_rad))
+            )
+            Hc[Hc == 0] = np.nan
+            # Spatial probability
+            p_S = H_el / Hc
```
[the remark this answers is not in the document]
BCampforts: Sweet! I was looking for a way to avoid 0 divisions

landlab/components/bedrock_landslider/bedrock_landslider.py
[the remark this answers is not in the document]
mcflugen: Yes, write a getter. The way you have it now a user would be able to set those attributes as well.
You could do something like the following,
```python
@property
def size_of_landslides(self):
    return self._landslides_size
```

landlab/components/bedrock_landslider/bedrock_landslider.py
BCampforts: Totally, test was only covering if no landslides occur on boundaries, not if the actual landslide still occurs

landlab/components/bedrock_landslider/bedrock_landslider.py
BCampforts: Nicely spotted!

landlab/components/bedrock_landslider/bedrock_landslider.py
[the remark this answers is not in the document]
BCampforts: Sounds good, done

landlab/components/bedrock_landslider/bedrock_landslider.py
BCampforts: Done

landlab/components/bedrock_landslider/bedrock_landslider.py
BCampforts: Done

landlab/components/bedrock_landslider/bedrock_landslider.py
BCampforts: Done

landlab/components/bedrock_landslider/bedrock_landslider.py
BCampforts: I see. Done

landlab/components/bedrock_landslider/bedrock_landslider.py
BCampforts: Yup, will do it too. Changed

landlab/components/bedrock_landslider/bedrock_landslider.py
BCampforts: Done

landlab/components/bedrock_landslider/bedrock_landslider.py
BCampforts: Good point, I had it there for some functionality I might support in the future, but not just yet. Removed it.

landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,669 @@
...
+        phi=0,
+        max_pixelsize_landslide=1e9,
+        seed=2021,
+        verbose_landslides=False,
+        landslides_on_boundary_nodes=True,
+        critical_sliding_nodes=None,
+        store_landslides_size=True,
+        store_landslides_volume=True,
+        store_landslides_volume_sed=False,
+        store_landslides_volume_bed=False,
```
mcflugen, line 242: What do you think about removing these options? Instead, leave it up to the user to save landslide information from time step to time step and have, for example, `BedrockLandslider.landslides_size` always return the landslide sizes for the current time step.

landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -0,0 +1,669 @@
...
+                f"Fraction of fines must be between 0 and 1 ({fraction_fines_LS})"
+            )
+
+        # Set seed
+        if seed is not None:
+            np.random.seed(seed)
+
+    # Getters for properties
+    @property
+    def landslides_size(self):
```
mcflugen, line 345: Add a docstring with just a one line description for each of these properties.

landlab/components/bedrock_landslider/bedrock_landslider.py
BCampforts: Not a bad suggestion. I had to rewrite some of the end-user scripts but this is more clean for sure!

landlab/components/bedrock_landslider/bedrock_landslider.py
BCampforts: Done

mcflugen (approved):
👍

kbarnhart:
@BCampforts @mcflugen. This looks amazing. Sorry to be late to the game with this comment. 

Can you add .rst files for these two new components and update the [docs/source/reference/components/index.rst](https://github.com/landlab/landlab/blob/master/docs/source/reference/components/index.rst) to include them. 

That way they will be more easily discoverable from the component description page in the docs. https://landlab.readthedocs.io/en/master/reference/components/index.html

Issue #1369, opened 2021-11-29
SPACE component(s): make qs_in public


gregtucker:
SPACE tracks two forms of sediment flux at each grid node: the outgoing sediment flux from the node to one of its downstream neighbors ("receivers"), and the incoming sediment flux from its upstream neighbors ("donors"). The outgoing sediment flux is publicly available through the `sediment__flux` field. However, the incoming sediment flux field is private, hidden in the component's (base class) variable `qs_in`. Because it is sometimes useful for external code to read the contents of this variable, it should be made public, either as a field or as a public array.

Also, the `sediment__flux` field's information string is misleading and should be updated.

gregtucker:
Addressed by #1370

PR #1370, opened 2021-11-29
make qs_in a field called sediment__influx

gregtucker:
Make a field for previously private array _qs_in

Addresses #1369


landlab/components/erosion_deposition/generalized_erosion_deposition.py
```diff
@@ -126,11 +134,9 @@ def __init__(
         self.initialize_output_fields()
 
         self._qs = grid.at_node["sediment__flux"]
+        self._qs_in = grid.at_node["sediment__influx"]
```
mcflugen, line 137: `self._qs_in` is unused. I recommend either exposing it as a property (with a descriptive name) or removing it.

mcflugen (changes requested):

landlab/components/erosion_deposition/erosion_deposition.py
```diff
@@ -130,6 +130,14 @@ class ErosionDeposition(_GeneralizedErosionDeposition):
             "doc": "Node array containing downstream-to-upstream ordered list of node IDs",
         },
         "sediment__flux": {
+            "dtype": float,
+            "intent": "out",
+            "optional": False,
+            "units": "m3/s",
+            "mapping": "node",
+            "doc": "Sediment flux (volume per unit time of sediment leaving each node)",
```
mcflugen, line 138: The description of the *sediment__flux* field is now inconsistent with other *landlab* components. The *LateralEroder* says that *sediment__flux* is, *"Sediment flux (volume per unit time of sediment __entering__ each node)"* but *Space* says that it is, *"Sediment flux (volume per unit time of sediment __leaving__ each node)"*.

I'm not sure which description is correct but we need to either update the descriptions or change some field names to make things consistent.

mcflugen (changes requested):

landlab/components/erosion_deposition/generalized_erosion_deposition.py
gregtucker: `self._qs_in` is defined here because it is used by derived classes.

mcflugen: In that case, how about making it public? That way, derived classes don't have to depend on the existence of a private variable. I think the safest way to do that would be through a property,
```python
@property
def sedflux_influx(self):
    return  self.grid.at_node["sediment__influx"]
```
Another option would be to drop the leading underscore and change the name to *qs_in* but then someone could accidentally change the pointer to the data.

landlab/components/erosion_deposition/erosion_deposition.py
gregtucker: The new description is correct for ErosionDeposition and SPACE (both versions), but the old one is correct for LateralEroder! So we have a naming inconsistency... Here's a possible solution:
- change doc for sediment__flux to "Sediment flux (volume per unit time)"
- use sediment__influx if the flux is specified as coming into an element from adjacent elements
- use sediment__outflux if it is specified as leaving the element
- for backward compatibility, retain sediment__flux and have it refer to one of the above, whichever it did originally

mcflugen: It seems, to me anyway, the best option would be to use two different names depending on whether sediment in flowing into or out of grid cells. The pain will be to maintain backward compatibility.

What do you think?

landlab/components/erosion_deposition/generalized_erosion_deposition.py
gregtucker: Agreed - changed as suggested

landlab/components/erosion_deposition/erosion_deposition.py
gregtucker: Agreed. Names changed to sediment__influx and sediment__outflux. Pointer to one of these, as appropriate, with name sediment__flux is retained for backward compatibility until version 3.0.0+.

PR #1371, opened 2021-12-01
Fix release filters

mcflugen:
Our release actions were not being triggered correctly. This pull request tries to address this. What we want is the following:

* Prerelease: triggered on a tag that looks like a version but ends with either an *a* or a *b* followed by numbers (should match e.g. *v4.3.0b0* and *v4.3.0b99*, but not *v4.3.0*, or *v4.3.0a*).
* Release: triggered on a tag that looks like a version but does end with the prerelease suffix (should match e.g. *v.4.3.0* but not *v4.3.0b0*).

Prereleases are sent to *TestPyPI*, and releases are sent to *PyPI*.


mcflugen:
It turns out there was a syntax error in *release.yml* that was preventing the workflow from running 🤦‍♂️. I think I did fix an error, though, where double-digit prereleases were not recognized.

PR #1377, opened 2021-12-10
New DimensionlessDischarge Component

slundell123:
This PR adds a new component and tests for a dimensionless discharge debris flow calculator. The DimensionlessDischarge component calculates the dimensionless discharge value, debris flow threshold value, and boolean for predicted debris flow (based on the two prior values) for stream segments.

@eculler @BCampforts


mcflugen:
@slundell123 thank you for this contribution! I'll have a closer look at it this weekend or next week.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
```diff
@@ -0,0 +1,196 @@
...
+        },
+        "dimensionless_discharge_above_threshold": {
+            "dtype": bool,
+            "intent": "out",
+            "optional": False,
+            "units": "none",
+            "mapping": "node",
+            "doc": "Dimensionless discharge value for a stream segment.",
+        },
+        "flux": {
```
mcflugen, line 114: We'll need a more descriptive name here. Is this field used in another existing component? If so, it should have the same name as that field. Otherwise, it's name should more descriptive name.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
```diff
@@ -0,0 +1,196 @@
...
+            "dtype": float,
+            "intent": "in",
+            "optional": False,
+            "units": "m/s",
+            "mapping": "node",
+            "doc": "flux in each stream segment",
+        },
+        "d50": {
```
mcflugen, line 122: Same as above. This needs a more descriptive name.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
```diff
@@ -0,0 +1,196 @@
...
+            "dtype": float,
+            "intent": "in",
+            "optional": False,
+            "units": "m",
+            "mapping": "node",
+            "doc": "soil grain size average in stream segment",
+        },
+        "dem_values": {
```
mcflugen, line 130: Is this just elevation? Could this name be *"topographic__elevation"*, which is used by lots of other *landlab* components?

landlab/components/dimensionless_discharge/dimensionless_discharge.py
```diff
@@ -0,0 +1,196 @@
...
+        super().__init__(grid)
+
+        # Store parameters
+        self._soil_density = soil_density
+        self._C = [C] * grid.number_of_nodes
+        self._N = [N] * grid.number_of_nodes
+        # change DEM values into slope of a stream segment
+        self._stream_slopes = grid.calc_slope_at_node(elevs="dem_values")
+        self.water_density = water_density
+        self.gravity = 9.8
```
mcflugen, line 167: How about adding *gravity* as a keyword and setting its default to a more accurate value? You can use *scipy.constants* for this.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+
+        super().__init__(grid)
+
+        # Store parameters
+        self._soil_density = soil_density
+        self._C = [C] * grid.number_of_nodes
+        self._N = [N] * grid.number_of_nodes
+        # change DEM values into slope of a stream segment
+        self._stream_slopes = grid.calc_slope_at_node(elevs="dem_values")
```
mcflugen, line 165: This is just used within this method so can be a local variable.
```python
stream_slopes = grid.calc_slope_at_node(elevs="dem_values")
```

landlab/components/dimensionless_discharge/dimensionless_discharge.py
```diff
@@ -0,0 +1,196 @@
...
+        self.water_density = water_density
```
mcflugen, line 166: This should be a private variable.
```python
self._water_density = water_density
```
If you intend for users to change the density of water after the component has been instantiated, you should set it as a property and add a setter. So something like,
```python
@property
def water_density(self):
    return self._water_density

@water_density.setter
def water_density(self, new_value):
    if water_density > 0.0:
        self._water_density = new_values
    else:
        raise ValueError(f"water density must be greater than zero ({new_value})")

File: landlab/components/dimensionless_discharge/dimensionless_discharge.py (hunk 5 above)
Comment (mcflugen) [reply landlab/components/dimensionless_discharge/dimensionless_discharge.py:165]: Actually, I imagine the slopes can change from time step to time step so maybe this calculation should go into the *run_one_step* method?

**Review by mcflugen — CHANGES_REQUESTED** [review]

**Comment by kbarnhart** [comment]
This is awesome! Super excited that you guys have created this. 

A big-picture comment.

It looks like D50 is an expected at-node field (though not documented as such). I think it may make sense to initialize both D50 and the sediment density value using the `return_array_at_node` function. This function allows an input to be specified as either a float, an array, or a string referring to an at-node field. You can specify the default as a field name (if you wish) and then indicate that that field is an optional input. 

`from landlab.utils.return_array import return_array_at_node`

Take a look at the input keyword argument "surface" in the FlowAccumulator to see an example. 

The reason for this recommendation is that it seems very reasonable that applications might want either of these values (D50 and sediment density) to be a constant or an array. The return_array_at_node function was made to make this easy.

File: landlab/components/dimensionless_discharge/dimensionless_discharge.py (hunk 1 above)
Comment (kbarnhart) [reply landlab/components/dimensionless_discharge/dimensionless_discharge.py:114]: Agreed. And it may make sense to have this be an existing field name, if possible. As defined in Tang et al, I think q should be thought of as discharge per unit width, as they define it as q=h*(u^2+v^2)^0.5 where h=water depth, and u and v are velocity vectors. 

It would be equivalent to "surface_water__discharge" divided by the relevant face width, I think. That might be a good default, as using it would permit connection to existing components. 

Either that or define a new field that reflects surface water discharge per unit width.

This page may be useful. https://landlab.readthedocs.io/en/master/user_guide/field_io.html

File: landlab/components/dimensionless_discharge/dimensionless_discharge.py (hunk 2 above)
Comment (kbarnhart) [reply landlab/components/dimensionless_discharge/dimensionless_discharge.py:122]: maybe "grain_size__median"... we should use something decent for this standard name since it is a standard value. What do you think @mcflugen

File: landlab/components/dimensionless_discharge/dimensionless_discharge.py (hunk 3 above)
Comment (kbarnhart) [reply landlab/components/dimensionless_discharge/dimensionless_discharge.py:130]: I think this is topographic elevation. 

I'd also note, that language here is a bit confusing. These are values at node, but the documentation string refers to "stream segment", the most analogous grid element in landlab to a segment is a link.

File: landlab/components/dimensionless_discharge/dimensionless_discharge.py [hunk 7]
```diff
@@ -0,0 +1,196 @@
...
+     against previous debris flow events.  The values will vary based on
+     local geology and soil type. In southern California values of C=12
+     and N=0.85 (upper limits, Tang et al, 2019) and C=4.29, N=0.78
+     (lower limits, Tang et al, 2019) have been used, while values of
+     C=0.195 and N=1.27 have been in used in the Italian Dolomites
+     (Gregoretti and Fontana, 2008). Default values are C=12 and N=0.85.
+
+     Parameters
+     ----------
+     soil_density : float list
```
kbarnhart, line 53: inconsistancy here between these parameters and those below. Is soil density a float or a float list. Recommend only writing this down in one place.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
```diff
@@ -0,0 +1,196 @@
...
+     to determine whether the dimensionless discharge calculated
+     exceeds thresholds for a sediment-laden (Upper Limit) or
+     water-producing only (Lower Limit) debris flow.  C and N are
+     empirically-derived by comparing dimensionless discharge estimates
+     against previous debris flow events.  The values will vary based on
+     local geology and soil type. In southern California values of C=12
+     and N=0.85 (upper limits, Tang et al, 2019) and C=4.29, N=0.78
+     (lower limits, Tang et al, 2019) have been used, while values of
+     C=0.195 and N=1.27 have been in used in the Italian Dolomites
+     (Gregoretti and Fontana, 2008). Default values are C=12 and N=0.85.
```
kbarnhart, line 49: This is a really nice overview of the component. Nice job!

landlab/components/dimensionless_discharge/dimensionless_discharge.py
```diff
@@ -0,0 +1,196 @@
...
+        },
+        "dem_values": {
+            "dtype": float,
+            "intent": "in",
+            "optional": False,
+            "units": "m",
+            "mapping": "node",
+            "doc": "elevation stream segment",
+        },
+    }
```
kbarnhart, line 138: For new fields, please update the following two files:
docs/source/user_guide/field_definitions.rst
docs/source/user_guide/field_io.rst

The python script at the below link may be helpful. 
docs/build_standard_names_list.py

kbarnhart (changes requested):

landlab/components/dimensionless_discharge/dimensionless_discharge.py
```diff
@@ -0,0 +1,196 @@
...
+        # set threshold values for each segment
+        _ = self.grid.add_zeros("node", "dimensionless_discharge")
+        _ = self.grid.add_zeros("node", "dimensionless_discharge_above_threshold")
+        self.grid.at_node["dimensionless_discharge_above_threshold"] = np.array(
+            [[False] * self.grid.number_of_nodes]
+        )
+        _ = self.grid.add_zeros("node", "dimensionless_discharge_threshold_value")
+        self.grid.at_node["dimensionless_discharge_threshold_value"] = self._C / (
+            self._stream_slopes ** self._N
+        )
```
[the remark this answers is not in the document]
kbarnhart: @mcflugen I think that C and N do need to be stored, since the threshold value should also be calculated in run one step (because it depends on the slopes, which should be calculated in run one step).

landlab/components/dimensionless_discharge/dimensionless_discharge.py
```diff
@@ -0,0 +1,196 @@
...
+            "mapping": "node",
+            "doc": "Dimensionless discharge value for a stream segment.",
+        },
+        "dimensionless_discharge_above_threshold": {
+            "dtype": bool,
+            "intent": "out",
+            "optional": False,
+            "units": "none",
+            "mapping": "node",
+            "doc": "Dimensionless discharge value for a stream segment.",
```
kbarnhart, line 112: Doc string here needs updating.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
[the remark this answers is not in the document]
mcflugen: @kbarnhart Yes, definitely. I wrote this comment before I realized that the slopes would have to be calculated within *run_one_step*.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
```diff
@@ -0,0 +1,196 @@
...
+            threshold equation; Empirically derived constant (See Tang
+            et al. 2019)
+        """
+
+        super().__init__(grid)
+
+        # Store parameters
+        self._soil_density = soil_density
+        self._C = [C] * grid.number_of_nodes
+        self._N = [N] * grid.number_of_nodes
```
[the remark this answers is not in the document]
mcflugen: If slopes will be calculated within *run_one_step*, then both *C* and *N* will need to be saved. Since *numpy* will broadcast both to the correct shape in the *dimensionless_discharge_threshold_value calculation*, you can just save the passed values.
```python
self._c = C
self._n = N
```

landlab/components/dimensionless_discharge/dimensionless_discharge.py
mcflugen: @kbarnhart We should. Maybe *sediment_grain__d50_diameter*? It's an existing name.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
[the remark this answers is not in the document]
mcflugen: The *dimensionless_discharge_threshold_value* is neither an input or an output field. Perhaps it should just be an attribute of the class?
```python
    self._dimensionless_discharge_threshold_value = C / self._stream_slopes ** N
```
Or, if this is being moved into *run_one_step*, it can just be a local variable.
```python
    dimensionless_discharge_threshold_value = self._c / stream_slopes ** self._n
```

landlab/components/dimensionless_discharge/dimensionless_discharge.py
kbarnhart: If that is an existing name, then yes, that's what should be used. Is that a CSDMS standard name?... I couldn't find a component that used it?

landlab/components/dimensionless_discharge/dimensionless_discharge.py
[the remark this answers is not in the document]
kbarnhart: I think its meant to be an output field...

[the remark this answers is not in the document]
mcflugen: Makes sense.

@slundell123 if that's the case, please add this field to the class's *_info* dictionary. I suggest also renaming it to just *dimensionless_discharge_threshold*.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
mcflugen: Yes, a CSDMS Standard Name (the existing name is actually, *channel_bottom_sediment_grain__d50_diameter*). There aren't any *landlab* components that use it.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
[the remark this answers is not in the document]
slundell123: Updated.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
```diff
@@ -0,0 +1,196 @@
...
+        self.water_density = water_density
+        self.gravity = 9.8
+
+        # set threshold values for each segment
+        _ = self.grid.add_zeros("node", "dimensionless_discharge")
+        _ = self.grid.add_zeros("node", "dimensionless_discharge_above_threshold")
+        self.grid.at_node["dimensionless_discharge_above_threshold"] = np.array(
+            [[False] * self.grid.number_of_nodes]
+        )
+        _ = self.grid.add_zeros("node", "dimensionless_discharge_threshold_value")
```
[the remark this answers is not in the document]
slundell123: Fixed.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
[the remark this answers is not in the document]
slundell123: I have added the field to the _info dictionary. It is used to determine if the dimensionless discharge value would exceed the threshold and  produce a debris flow, but it is possible that someone would like access to this value as an output.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
```diff
@@ -0,0 +1,196 @@
...
+            * (self.grid.at_node["d50"] ** 3)
+        )
+
+        self.grid.at_node["dimensionless_discharge_above_threshold"] = [
+            True
+            if self.grid.at_node["dimensionless_discharge"][i]
+            >= self.grid.at_node["dimensionless_discharge_threshold_value"][i]
+            else False
+            for i in range(self.grid.number_of_nodes)
+        ]
```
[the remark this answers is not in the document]
slundell123: Fixed.

tests/components/dimensionless_discharge/test_dimensionless_discharge.py
```diff
@@ -0,0 +1,62 @@
...
+from landlab import RasterModelGrid
+from landlab.components.dimensionless_discharge import DimensionlessDischarge
+
+
+def init_grid():
+    """initialize grid for testing"""
+    watershed_grid = RasterModelGrid((3, 3))
+    _ = watershed_grid.add_ones("node", "flux")
+    _ = watershed_grid.add_ones("node", "d50")
+    _ = watershed_grid.add_ones("node", "dem_values")
```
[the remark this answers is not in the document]
slundell123: Updated.

tests/components/dimensionless_discharge/test_dimensionless_discharge.py
```diff
@@ -0,0 +1,62 @@
...
+    watershed_grid.at_node["dem_values"] = np.array([[1.1, 2, 3, 4, 2, 3, 4, 5, 3]])
+    return watershed_grid
+
+
+def test_dimensionless_discharge_final_values():
+    """Testing for correct final dimensionless discharge values"""
+    watershed_grid = init_grid()
+    dd = DimensionlessDischarge(watershed_grid)
+    dd.run_one_step()
+    expected_values = [
```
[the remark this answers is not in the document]
slundell123: 0.55372743 is the expected value from the dimensionless discharge (q*) equation in the Tang et al paper when flux = 1 and d50 =1. This was hand-calculated to validate the equation is set up correctly. I will add a comment about this.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: @eculler Do you have an opinion on this (will forward to Ryan too)

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: @eculler Thoughts on this naming too?  channel_bottom_sediment_grain__d50_diameter sounds accurate to me.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: We have specifically been using a digital elevation model (dem) to produce our elevation values, but topographic__elevation likely captures the information that is needed better. I will make that change.

@kbarnhart I have been a bit confused on the best grid types to use for the different variables (I have little experience working with grids), so if you think a link would be better I can make that update. Stream segments are sections of the stream that we average over as we can not run this model over every single point along a river.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: Updated. I have set the value for gravity to 9.8 for my tests and that is what I hand-calculated my expected values with.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: You use the same streams segments for each iteration so I do not think this value would change. @eculler thoughts?

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: Made into a private variable.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: Removed duplicate listing of parameters.

slundell123:
@kbarnhart @mcflugen Thanks for all the feedback! I have commented on the reviews that I have addressed or started to work on. I will continue to comment as I go and will work on the ones I have not yet gotten to next week.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: Updated.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: Thanks!

slundell123: Thanks!

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: Updated.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: I ran that script (docs/build_standard_names_list.py), but it didn't seem to do anything. What is it meant to do?

landlab/components/dimensionless_discharge/dimensionless_discharge.py
mcflugen: Since these elevations are a *landlab* field, they can be shared with other *landlab* components that may update the elevations with each time step. This component could, for example, be used within a coupled set of components that form a landscape evolution model where the elevation of the landscape is updated with every time step.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: After thinking about thinks in the context of this problem I think that using nodes for this would still be best. We need a slope for each stream segment, so an elevation is given for each location of a stream segment (our nodes). I am not sure how you would convert links into elevations for each node.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: I added updating the elevations into the run_one_step function

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: I have changed the name flux to surface_water__discharge

landlab/components/dimensionless_discharge/dimensionless_discharge.py
kbarnhart: Sorry, should have been clearer. That script can be used to generate the majority of the two files that need updating. 

I've updated the script so that it runs automatically when the docs are built and creates the entirety of the two files. (flat @mcflugen). Will push that change to this branch shortly.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
slundell123: @kbarnhart Just had a conversation about this variable. Something like Surface_Water__Width_Normalized_Dischage or Surface_Water__Unit_Dischange would capture what this is a bit better. Is there a variable similar to these already (I couldn't find one)?

kbarnhart:
@slundell123 I've found that it is quite a bit easier for me to make some direct edits so I'll be pushing a few more commits with those changes for your consideration (will be pushed after I test them). If you don't like any of them, feel free to modify. The changes relate to the use of optional fields, the return_array constructor, and when the thresholds are calculated (this should also be in run one step, as slopes are updated in run one step).

landlab/components/dimensionless_discharge/dimensionless_discharge.py
kbarnhart: @mcflugen can you provide guidance here?

kbarnhart: I'm OK with surface_water__unit_discharge with a description that states that it is volumetric surface water discharge divided by channel width or grid cell width. Also tapping in @gregtucker as he may have thoughts on what it should be called. 

I think that once we have a good name for this field, its metadata is updated, (and the two field-related .rst files are updated to reflect it) that this component is looking good. For example, the metadata has the units as L/T, and I think they should be L^2/T

landlab/components/dimensionless_discharge/dimensionless_discharge.py
kbarnhart: I think what you have is fine.

landlab/components/dimensionless_discharge/dimensionless_discharge.py
kbarnhart: A similar issue is to do this for calculating the threshold value (as it can change with changing slopes). I've addressed this is some changes I've pushed for your consideration.

kbarnhart:
@slundell123 I've finished going through this again. Nice work!

Other than the name & documentation of discharge per unit width, I think this is ready to go. @mcflugen may have other comments. 

Let me know if you have any questions/comments about the changes I've pushed. Please consider them as suggestions. If you don't like something I did, please revise it away.

slundell123:
@kbarnhart your changes look good to me. I just updated the variable surface_water__discharge to surface_water__unit_discharge as discussed above. I'm not too sure how to change the field-related .rst files that you mentioned.

slundell123:
@mcflugen @kbarnhart Is there anything else I need to do on this so that it can be merged?

kbarnhart:
Will take a look in the next few days.

mcflugen (approved):
👍

PR #1396, opened 2022-03-03
Setup towncrier

mcflugen:
This pull request sets up *[towncrier](https://towncrier.readthedocs.io/en/actual-freaking-docs/)* for updating and organizing our changelog.

@gregtucker, @mdpiper could one or both of you have a look at this and let me know what you think?

I've set *landlab* up to use the *towncrier* utility to, hopefully, help maintain a better CHANGELOG that our users (and we) can look at to quickly see all of the great new changes. The basic idea is that when someone submits a pull request they must create one or more files in the *news/* folder (with the name `<issue/pr #>.<pr type>`; e.g. `3142.bugfix`). We can then use *towncrier* to gather these news fragments and add them to our changelog. I've added some documentation about [how to create news fragments](https://github.com/landlab/landlab/pull/1396/files#diff-afacdd599c6b35b3c037a41740e451eb355488d5ed1df11ccc8af3f2b34bb062) that highlight what a merged pull request has added to the repository.

I've defined several fragment types:
* `component`: adds a new component
* `notebook`: adds a tutorial notebook
* `feature`: adds a new feature
* `docs`: improves the documentation
* `bugfix`: fixes a bug
* `misc`: everything else

We're not limited to these and it's easy to add other categories if you think we need more.

I've updated the current CHANGES.rst so you can see what [it will look like](https://github.com/landlab/landlab/blob/mcflugen/setup-towncrier/CHANGES.rst) if we decide to go this route. For the most part the entries are basically just pull request messages and aren't really what we're after—entries should, ideally, be full sentences and more descriptive. I may go through the changelog and add better descriptions but I also might leave that for another time and another pull request.

### DIFF
--- a/docs/source/development/contribution/index.rst
+++ b/docs/source/development/contribution/index.rst
@@ -37,7 +37,7 @@ summarized here when finalized. If you have any questions regarding whether
 your potential JOSS submission is appropriate, the best thing to do is to
 make a pre-submission inquiry with JOSS.
 
-NEWS Entries
+News Entries
 ------------
 
 The ``CHANGES.rst`` file is managed using `towncrier`_ and all non trivial changes
--- a/docs/source/development/contribution/index.rst
+++ b/docs/source/development/contribution/index.rst
@@ -36,3 +36,31 @@ contributions to Landlab fit the scope of a JOSS submission. These will be
 summarized here when finalized. If you have any questions regarding whether
 your potential JOSS submission is appropriate, the best thing to do is to
 make a pre-submission inquiry with JOSS.
+
+NEWS Entries
+------------
+
+The ``CHANGES.rst`` file is managed using `towncrier`_ and all non trivial changes
+must be accompanied by a news entry.
+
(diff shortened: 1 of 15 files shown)


docs/source/development/contribution/index.rst
```diff
@@ -36,3 +36,31 @@ contributions to Landlab fit the scope of a JOSS submission. These will be
 summarized here when finalized. If you have any questions regarding whether
 your potential JOSS submission is appropriate, the best thing to do is to
 make a pre-submission inquiry with JOSS.
+
+NEWS Entries
```
mdpiper, line 40: Maybe "News" instead of "NEWS"?

mdpiper (dismissed):

mdpiper:
It'll be a little extra work for a contributor, but I think it'll be useful to track changes in a human-readable way, especially as Landlab continues to grow.

gregtucker:
I agree - thanks for adding this!

mcflugen:
Thanks @gregtucker and @mdpiper!

PR #1398, opened 2022-03-08
Add plot_layers function

mcflugen:
This pull request adds a new function, *plot_layers*, that creates a stratigraphic plot of sediment layers.


mcflugen:
@gregtucker I think this is ready for you to have a look at when you have the time.

gregtucker (dismissed):
Nice! And incorporated into a tutorial too!

PR #1397, opened 2022-03-08
Fixed reference in PriorityFloodFlowDirector

BCampforts:
The main change in this PR is to fix a broken reference in the PriorityFloodFlowDirector where the gradients of the hillslopes are being updated. A second fix is to resolve instabilities related to the use of very small H* values when using the "Space_Large_Scale_Eroder". Some other cosmetic changes and added control on location of ylabels of colorbars in imshowhs_grid function. @mcflugen this is a minor change compared to the previous PR, mainly fixing some minor issues.

PR #1399, opened 2022-03-09
Test landlab with Python 3.10

mcflugen:
This pull request builds and tests *landlab* and the tutorial notebooks using Python 3.8, 3.9 and now 3.10 (and drops 3.7).

I hope this will also fix the notebook test failures @BCampforts is seeing in #1397.

### DIFF
--- a/.github/workflows/test-notebooks.yml
+++ b/.github/workflows/test-notebooks.yml
@@ -31,7 +31,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.8", "3.9"]
+        python-version: ["3.8", "3.9", "3.10"]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


mcflugen:
Highlights!
* I've set up our main tests to test Python 3.8, 3.9, and 3.10 (added 3.10, dropped 3.7).
* We now only test the notebooks with Python 3.9. The tests take a long time to run but I think testing with just one version is probably ok. I encountered some installation issues with the notebook requirements with Python 3.10, which is why I chose 3.9 instead of 3.10.
* Test all the notebooks on Windows—even the ones that require *richdem*.
* Fixed a bug that incorrectly diagnosed if the *richdem* engine is working.

gregtucker (approved):
Looks good!

For notebooks on other Py versions, if this turns out to be helpful maybe we can come up with a way to run them only in advance of Minor and Major releases, rather than for every PR?

mcflugen:
Thanks @gregtucker!

> For notebooks on other Py versions, if this turns out to be helpful maybe we can come up with a way to run them only in advance of Minor and Major releases, rather than for every PR?

Agree. I think we should be doing something along these lines as it's probably not necessary to run these tests with every change that's pushed to GitHub—they consume a good bit of time and computational resources.

PR #1400, opened 2022-03-17
Add demo of NetworkModelGrid creator on OpenTopography DEM

shelbyahrendt:
This PR adds a [Jupyter Notebook Tutorial](https://github.com/shelbyahrendt/landlab/blob/add-network-from-raster/notebooks/tutorials/network_sediment_transporter/run_network_generator_OpenTopoDEM.ipynb) that demonstrates several options for creating a NetworkModelGrid on a DEM fetched from OpenTopography using the [Topography utility](https://bmi-topography.readthedocs.io/en/latest/). The notebook demonstrates the following:
1. Fetches the DEM
2. Re-projects the coordinate system (into one that makes more sense for defining a watershed area)
3. Trims the DEM to the largest watershed
4. Creates a NetworkModelGrid in three different ways


mcflugen:
> @mcflugen I'm not sure if I set-up this PR correctly--I meant it to only reflect the recent commits with the new tutorial, but it also looks like it includes a bunch of stuff from awhile ago!

I think the issue is that you continued to work on the same branch as the old pull request (*add-network-from-raster*) and so all those commits are listed here. However, since we merged that pull request with all those commits, only a few files are actually changed from the master branch.

Anyway, this looks great and I think it's ready to be merged.

mcflugen (approved):
Looks good @shelbyahrendt! 👍

PR #1404, opened 2022-03-30
Build docs Sphinx v4

mcflugen:
This pull request modifies the *landlab* docs so they will build with newer versions of *Sphinx*—up until now we had required *Sphinx* v2 but the latest is v4. I believe the main reason we were using the old version of *Sphinx* was because it ignored duplicated documentation (i.e. a class method or attribute was documented in two separate locations), which we inadvertently had in a few places.

The most common cause of this was a class whose docstring contained something like,
```python
class SomeClass:
    """
    ...

    Methods
    -------
    some_method

    ...

    """
```
but then the method itself had its own docstring,
```python
def some_method(self):
    """Docstring for some_method."""
```


kbarnhart:
@mcflugen can you turn on read the docs building for this branch so I can verify they work?

kbarnhart:
> I just realized that I haven't checked to see that they build on _read the docs_, though. That's something I need to do.

☝️ yes, exactly, this is what I wanted to look at. Thanks! And 🎉 YAY for this!

mcflugen:
@kbarnhart You can now view the docs on [readthedocs](https://landlab.readthedocs.io/en/mcflugen-build-docs-sphinx-v4/).

The main change you will notice right away is that I changed the theme. Our old theme used a custom layout (the now deleted *layout.html* file) that, for whatever reason, stopped working on *readthedocs* with the newer version of *Sphinx* (or maybe *jinja*, I don't really know). Instead of trying to debug the problem, I switched to a more standard layout that is easier to work with—and I actually like the look of the new theme better.

Anyway, please have a look and let me know what you think.

mcflugen:
@kbarnhart If you have a moment could you have a look at the new docs?

kbarnhart (approved):
Looks great @mcflugen . I looked at the readthedocs page. Love it. Thanks for keeping the docs in good health!

PR #1407, opened 2022-04-11
Gt/add papers

gregtucker:
Add several new Landlab-using references to the list of papers.


mcflugen (approved):
👍 so many references!

PR #1408, opened 2022-04-13
Add script to download landlab notebooks

mcflugen:
This pull request adds a short script that can be used to download a set of *landlab* notebooks.

If the user doesn't have this script locally (e.g. they have a version of *landlab* without this script—v2.4.1, for example) they could run something like the following,
```bash
 $ curl -L https://raw.githubusercontent.com/landlab/landlab/mcflugen/fetch-notebooks-script/notebooks.py | python - v2.4.1
```
A user can download a set of notebooks from a particular version of *landlab*,
```bash
$ python notebooks.py v2.4.1
```
or to get the latest version of the notebooks, simply leave out the version,
```bash
$ python notebooks.py
```
In all cases, just the *notebooks* folder from the *landlab* repository will be downloaded.

This script uses only packages from the Python standard library so there shouldn't be any bootstrapping issues if a user doesn't already have *landlab* (or any other 3rd party packages) installed.


mcflugen:
Still to do:
* Add documentation
* Fetch the dependencies needed for a set of notebooks

mcflugen:
Here is one possible workflow for a new user who doesn't have *landlab* installed on their computer.

First get the notebooks,
```
$ curl -L https://raw.githubusercontent.com/landlab/landlab/mcflugen/fetch-notebooks-script/notebooks.py | python - v2.4.1
fetching notebooks for landlab v2.4.1
https://github.com/landlab/landlab/archive/v2.4.1.tar.gz
notebooks have been extracted into landlab-2.4.1
landlab-2.4.1
```
and then install *landlab* and its dependencies,
```
$ conda create -n landlab-tutorials --file=landlab-2.4.1/requirements-notebooks.txt
$ conda activate landlab-tutorials
```

gregtucker (commented):
Bravo!

Curious about the usage: would someone download notebooks.py and run it? Or would there be shell script added to the user's path when LL is installed such that they can run the script via a command, without worrying about the path location of notebooks.py?

mcflugen:
Right now it's simply a script that a user would run. I did it this way so that a user who already has *landlab* installed (or doesn't even have it installed yet) could still use it without upgrading to the development version (or our next release).

mcflugen:
@BCampforts @gregtucker I've updated both the *landlab* installation instructions and the tutorial installation instructions.

You can view them here:
* [*landlab* installation](https://landlab.readthedocs.io/en/mcflugen-fetch-notebooks-script/install/index.html#install)
* [tutorials installation](https://landlab.readthedocs.io/en/mcflugen-fetch-notebooks-script/user_guide/tutorials.html)

@mdpiper you may want to have a look at the [*landlab* installation](https://landlab.readthedocs.io/en/mcflugen-fetch-notebooks-script/install/index.html#install) instructions and see how I use tabs for the *conda* and *pip* commands. I think this is something we could do for the *bmi* documentation where a reader could switch from one language to another when looking at the language bindings.

mdpiper:
@mcflugen The tabs look nice! I'll use this pattern when I next update docs.

PR #1418, opened 2022-05-12
Fix read_shapefile unit tests caused by closed BytesIO streams

mcflugen:
This pull request fixes some failing unit tests caused by a new version of *pyshp*. With *pyshp* version 2.3.0, *shapefile.Writer* closes open streams, which makes those streams inaccessible after the writer completes, causing tests that continue to use these streams to fail. This error is described in GeospatialPython/pyshp#244.

For now I've just limited *pyshp* versions to anything other than 2.3.0.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ matplotlib
 netcdf4
 numpy
 pyyaml
-pyshp ~= 2.3.0
+pyshp != 2.3.0
 scipy
 statsmodels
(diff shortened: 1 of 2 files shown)


karimbahgat:
Fixed in v2.3.1.

mcflugen:
> Fixed in v2.3.1.

:tada: Excellent! Thanks @karimbahgat!

Issue #1419, opened 2022-05-12
Fix broken links to JupyterHub in README


mdpiper:
The links in the README to run the Landlab tutorials on the CSDMS JupyterHub are broken (they point to the old URL for the Hub).

mdpiper:
Fixed with #1420.

PR #1420, opened 2022-05-12
Update links to the new jupyterhub on lab.openearthscape

mcflugen:
This pull request fixes the broken links to the openearthscape JupyterHub as described in #1419.

@mdpiper Does this fix the issue? Are these the appropriate links?


mdpiper:
Maybe swap "jupyter" for "lab"--people should only get to _lab_ through a class, whereas _jupyter_ would be the path for a casual explorer of LL.

mcflugen:
Ah, yes, of course. How's this?

mdpiper:
👍

Issue #1424, opened 2022-05-15
grid.imshow should use `at` keyword


mcflugen:
The `grid.imshow` function should use the `at` keyword like other _landlab_ functions. That is,
```python
grid.imshow("topographic__elevation", at="node")
```
rather than
```python
grid.imshow("node", "topographic__elevation")
```
Since `imshow_grid` already takes this form, this should be easy to implement by just updating the `RasterModelGridPlotter` mix-in in *raster.py*. The only tricky bit will be to maintain backward compatibility. Something along the lines of the following should work,
```python
class RasterModelGridPlotter:
    def imshow(self, *args, **kwds):
        if len(args) == 1:
            name = args[0]
        elif len(args) == 2:
            at, name = args
            raise DeprecationWarning(f"use grid.imshow({name!r}, at={at!r})")
            if "at" in kwds:
                raise ValueError("location was provided through the 'at' keyword as well as an argument")
        else:
            raise TypeError(f"imshow expected 1 or 2 arguments, got {len(args)}")

        imshow_grid(self, name, **kwds)
```

PR #1430, opened 2022-05-20
Add "at" keyword for imshow and other enhancements

mcflugen:
This pull request addresses #1424 but adding an *at* keyword to *imshow* when used as a *ModelGrid* method. There are also a couple other changes:
* *imshow* is now available to all *ModelGrids*
* *imshow* can now guess at a location to plot values if the *at* keyword is not given. This allows things like: `grid.imshow(values)` and `grid.imshow("topographic__elevation")`.
* deprecates the *values_at* keyword
* deprecates usages like `imshow("node", "topographic__elevation")`

### DIFF
--- a/landlab/grid/base.py
+++ b/landlab/grid/base.py
@@ -19,6 +19,7 @@ from ..core.utils import add_module_functions_to_class
 from ..field.graph_field import GraphFields
 from ..layers.eventlayers import EventLayersMixIn
 from ..layers.materiallayers import MaterialLayersMixIn
+from ..plot.imshow import ModelGridPlotterMixIn
 from ..utils.decorators import cache_result_in_object
 from . import grid_funcs as gfuncs
 from .decorators import (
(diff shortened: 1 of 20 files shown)


mcflugen:
@vaga3461, @ethan-pierce: I noticed you used the `imshow_grid` function a lot in your *landlab* clinic. Would one of you have time to give this pull request a look over and let me know what you think of the changes?

I added an `imshow` method to all of the grids (raster, hex, radial, etc.) so that you don't have to import `imshow_grid` anymore. So instead of,
```python
from landlab.plot import imshow_grid
imshow_grid(grid, ...)
```
you can just do,
```python
grid.imshow(...)
```
I've also made the syntax (I think) a little cleaner and more like other parts of *landlab*. Instead of,
```python
imshow_grid(grid, "node", "topographic__elevation")
```
you can do,
```python
grid.imshow("topographic__elevation", at="node")
```
If you don't provide an `at` location, `imshow_grid` will guess at a location.
```python
grid.imshow(values)  # values is a array of values, doesn't have to be a field
grid.imshow("topographic__elevation"). # looks for the named field first at nodes, and then at cells
```

vaga3461 (approved):
these changes look good to me, I like that you've made the syntax more similar to other functions. Thanks!

mcflugen:
Thanks @vaga3461! I'll update the usages of *imshow_grid* in the notebooks and then merge.

Issue #1435, opened 2022-06-24
Silly ubuntu test fails on -0.


gregtucker:
A couple of doctests are (sometimes?) failing because the test expects a value of 0. but returns a value of -0. Presumably this is a round-off error. Specifics are below.

Possible solutions:
- Use absolute value
- Use numpy.testing all_close or similar

Case 1: DepthSlopeProductErosion

```
landlab.components.detachment_ltd_erosion.generate_erosion_by_depth_slope.DepthSlopeProductErosion _
[gw1] linux -- Python 3.9.13 /usr/share/miniconda/envs/test/bin/python3.9
091     Using the set topography, now we will calculate slopes on all nodes.
092 
093     First calculating slopes on links
094 
095     >>> grid.at_link['water_surface__slope'] = grid.calc_grad_at_link('surface_water__depth')
096 
097     Now putting slopes on nodes
098 
099     >>> grid['node']['water_surface__slope'] = (grid['link']['water_surface__slope'][grid.links_at_node] * grid.active_link_dirs_at_node).max(axis=1) # doctest: +NORMALIZE_WHITESPACE
100     >>> grid.at_node['water_surface__slope']
Expected:
    array([ 0.,  1.,  1.,  1.,  0., -0.,  1.,  1.,  1.,  0., -0.,  1.,  1.,
            1.,  0., -0.,  1.,  1.,  1.,  0.,  0.,  0.,  0.,  0.,  0.])
Got:
    array([ 0.,  1.,  1.,  1.,  0., -0.,  1.,  1.,  1., -0., -0.,  1.,  1.,
            1., -0., -0.,  1.,  1.,  1., -0., -0.,  0.,  0.,  0., -0.])
```

Case 2: map_upwind_node_link_max_to_node

```
[gw1] linux -- Python 3.9.13 /usr/share/miniconda/envs/test/bin/python3.9
714     >>> from landlab.grid.mappers import map_upwind_node_link_max_to_node
715     >>> from landlab import RasterModelGrid
716 
717     >>> rmg = RasterModelGrid((3, 4))
718     >>> rmg.at_link['grad'] = np.array([-1., -2., -1.,
719     ...                                  0.,  0.,  0.,  0.,
720     ...                                 -1., -2., -1.,
721     ...                                  0.,  0.,  0.,  0.,
722     ...                                 -1., -2., -1.])
723     >>> map_upwind_node_link_max_to_node(rmg, 'grad')
Expected:
[...]

PR #1436, opened 2022-06-25
Gt/fix ubunto minus zero

gregtucker:
Address #1435 to avoid spurious doctest failures on ubunto due to expected values of 0. versus actual reported values of -0.


gregtucker:
Changing the failing doctest in mappers.py is intended to avoid having a zero value, and hence avoid the possibility of -0 on some platforms and 0. on others. 

However, I am confused by the use of the minus sign on `values_at_links`. The sign convention on `link_dirs_at_node` is such that incoming links are positive and outgoing links are negative (and non-existent links are zero). Thus, if you multiply a flux value defined on links (i.e., `value_at_links`) by `link_dirs_at_node`, the resulting values should be positive where flow is coming into a node (i.e., "upstream") and negative where flow is leaving a node. The minus sign applied in this function seems to make it do the opposite! I have not tried to change this, as the function is used by at least one component. But I am scratching my head. @mcflugen can you offer a second opinion? I might just be confused.

mcflugen:
@gregtucker I was actually thinking the same thing and was going to ask you about it!

mcflugen (dismissed):
👍

PR #1445, opened 2022-07-19
Exclude tests folder from installed packages

mcflugen:
This pull request addresses conda-forge/landlab-feedstock#48. I've modified *pyproject.toml* to only include *landlab* ~and exclude the *tests* folder. This is overkill as just including `landlab*` would work just fine—no need to also exclude `tests*`.~

### DIFF
--- a/landlab/components/gravel_river_transporter/gravel_river_transporter.py
+++ b/landlab/components/gravel_river_transporter/gravel_river_transporter.py
@@ -7,7 +7,7 @@ from scipy.sparse.linalg import spsolve
 def make_empty_matrix_and_rhs(grid):
     from scipy.sparse import csc_matrix
 
-    mat = csc_matrix((grid.number_of_core_nodes, grid.number_of_core_nodes),)
+    mat = csc_matrix((grid.number_of_core_nodes, grid.number_of_core_nodes))
     rhs = np.zeros(grid.number_of_core_nodes)
     return mat, rhs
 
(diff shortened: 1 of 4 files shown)


mcflugen:
@mdpiper You might want to have a look at this. I've made this mistake before and maybe you have too.

For new packages I now create a *src* folder that contains just the package(s), which should help with inadvertently including extra folders in the installation.

mdpiper:
@mcflugen Thanks for this idea. I used the following in `pyproject.toml` in *bmi-topography* and (I think) it did what I wanted:
```toml
[tool.setuptools.packages.find]
where = ["."]
exclude = [
  "bmi_topography.tests*",
  "bmi_topography.examples*",
]
```

mcflugen:
@mdpiper Actually, I don't think that's doing what you want. I believe it should be:
```toml
[tool.setuptools.packages.find]
where = ["."]
exclude = [
  "tests*",
  "examples*",
]
```
What you had would exclude folders called *tests* and *examples* within the *bmi_topography* package. I think.

Well, on closer inspection, you don't need to exclude anything since neither the *tests* or *examples* folder contain `__init__.py` (and so aren't packages).
```python
>>> from setuptools import find_packages
>>> find_packages(".")
['bmi_topography']
```

mdpiper:
@mcflugen Ah, OK--I'll fix this!

mdpiper:
I built a wheel & only the _landlab/_ directory is present (no _tests/_).

mdpiper (approved):
What you've proposed here works! It looks like the failing notebook is unrelated (due to something in _rasterio_).

PR #1444, opened 2022-07-15
fix: update Mesa syntax

gregtucker:
Apparently the name of the Mesa.space.MultiGrid class in Mesa ABM has changed the syntax of the `_remove_agent` method to `remove_agent`, and removed one of the arguments. This PR updates the tutorial that uses Mesa ABM to reflect this change.

PR #1442, opened 2022-07-11
Gt/test lake mapper fix

gregtucker:
This fix, suggested by Sebastien Lénard, resolves an issue with flow re-routing on an irregular Voronoi-Delaunay grid.

PR #1439, opened 2022-07-07
feat: create new GravelRiverTransporter component

gregtucker:
New component to model gravel transport and abrasion in a gridded network of river segments

Issue #1433, opened 2022-06-20
Create a field for surface water height in LinearDiffusionOverlandFlowRouter


yuvalsmilo:
It is also uses that as an array. It will be easier if it will be as a field.

gregtucker:
Resolved by #1434

PR #1446, opened 2022-07-29
Add CI check for news fragments

mcflugen:
This pull request:
* adds a GitHub Actions workflow that checks if a pull request contains a news fragment
* adds missing news fragments for some recent pull requests.

It looks like the following issues/pull requests are missing news fragments,
* [x] #1444
* [x] #1442
* [x] #1433
* [x] #1439

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,26 @@
+name: Check
+
+on:
+  pull_request:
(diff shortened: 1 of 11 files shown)


mcflugen:
@gregtucker When you have a chance, could you have a quick look at this? I've added some missing news fragments for pull requests that you authored.

Issue #1440, opened 2022-07-07
Doc development/install/install.html need small update of install instructions to combine requirements.txt files in one only instruction


sebastien-lenard:
The page describing the steps to install landlab as a developer has outdated instructions to install requirements (as per the CSDMS 2022 school course and discussion with Mark).

https://landlab.readthedocs.io/en/master/development/install/install.html

The following instructions:

$ conda install -n landlab_dev --file=requirements.txt -c conda-forge
$ conda install -n landlab_dev --file=requirements-testing.txt -c conda-forge
$ conda install -n landlab_dev --file=requirements-notebooks.txt -c conda-forge
$ conda install -n landlab_dev --file=requirements-dev.txt -c conda-forge

should be replaced by this instruction:
$ conda install -c conda-forge -c defaults --file=requirements.txt --file=requirements-dev.txt --file=requirements-notebooks.txt

to make sure to group the conda requirements so that conda behave as expected when resolving dependencies.

Issue #1447, opened 2022-07-31
The *run_network_generator_OpenTopoDEM* tutorial notebook is failing


mcflugen:
The *run_network_generator_OpenTopoDEM* notebook is failing with the latest version of *rasterio* (*v1.3*). I believe this is due to a regression in *rasterio* and should, ultimately, be fixed there (see rasterio/rasterio#2538 for my attempt at a fix).

Until that fix is made, the offending cell could be modified to be,
```python
x_UTM13, _ = rasterio.transform.xy(transform, [0] * dim_x_transf, np.arange(0, dim_x_transf))
_, y_UTM13 = rasterio.transform.xy(transform, np.arange(0, dim_y_transf), [0] * dim_y_transf)
```
Not pretty but it should work.

mcflugen:
@shelbyahrendt Is this the same issue you were seeing?

shelbyahrendt:
Yes! That was the same problem, but now that I'm looking back I realize there had been a cascade of issues in delineating a watershed later on. I had that cell block running again with a similar workaround:

`x_UTM13, _ = rasterio.transform.xy(transform, np.zeros(dim_x_transf), np.arange(0, dim_x_transf, 1))
_, y_UTM13 = rasterio.transform.xy(transform, np.arange(0, dim_y_transf, 1), np.zeros(dim_y_transf))`

(which looks like it produces the same result.) However, something subtle may have changed in the transformation, because when I try to isolate the largest watershed using the same process of running the FlowAccumulator and get_watershed_masks_with_area_threshold() using the maximum drainage area I no longer get a single watershed (snip below). The previous notebook version used to return a single watershed bounded on the south somewhere around the red line.

![image](https://user-images.githubusercontent.com/36043125/183114480-2ef59853-1a22-42df-958d-e19473b2e21e.png)

My though was that some small change in the coordinate transformation resulted in drainage connectivity between cells on the west or east of the domain (assuming the only changes since creating the notebook were to the rasterio function). Sure enough, there's this strange north-flowing line of cells on the east side of the domain.

![image](https://user-images.githubusercontent.com/36043125/183124977-fabd83a4-7e0f-4743-a325-3795100f1e98.png)

I then tried to strategically adjust the domain bounds that the Topography tool pulls initially to get rid of this region. Due to the shape of the watershed, this didn't work so well so I switched the input lat,lon to a domain centered on Boulder Creek:

[...]

PR #1448, opened 2022-08-04
Update developer install doc - issue #1440 :memo:

sebastien-lenard:
Update the developer install documentation to make sure that the creation of the environment takes into account all dependencies in the same call. This solves the issue #1440.

@mdpiper

### DIFF
--- a/docs/source/development/install/install.rst
+++ b/docs/source/development/install/install.rst
@@ -65,10 +65,8 @@ The next step is it create this environment, called *landlab_dev*.
 .. code-block:: bash
 
    $ conda create -n landlab_dev python
-   $ conda install -n landlab_dev --file=requirements.txt -c conda-forge
-   $ conda install -n landlab_dev --file=requirements-testing.txt -c conda-forge
-   $ conda install -n landlab_dev --file=requirements-notebooks.txt -c conda-forge
-   $ conda install -n landlab_dev --file=requirements-dev.txt -c conda-forge
+   $ conda create -n landlab_dev -c conda-forge --file=requirements.txt --file=requirements-dev.txt --file=requirements-testing.txt --file=requirements-notebooks.txt
+
 
 The conda environment described by ``environment-dev.yml`` contains the minimal
 set of dependencies necessary to run the Landlab tests and notebooks, and keep


mcflugen:
Thanks @sebastien-lenard! 🎉

Before merging this, I would like to figure out what exactly is going on here. It doesn't make sense to me that it should make a difference if the dependencies are installed in one step or multiple steps.

sebastien-lenard:
> 
Hi!
- I tried to launch the "pytest test_notebooks.py" on my dev environment (on the jupyter hub), it skips all notebooks. How is the pytest launched in the validation process?
- I checked the notebook run_network_generator_OpenTopoDEM.ipynb and it doesn't seem to have a dependency with the doc file of this pull.

mdpiper:
@mcflugen When working with external code (C, Fortran), I've reliably encountered problems when creating a conda environment in a single step versus adding packages over multiple steps.

mcflugen:
@sebastien-lenard The failure in that notebook is related to #1447 and has nothing to do with your this request.

mcflugen:
> @mcflugen When working with external code (C, Fortran), I've reliably encountered problems when creating a conda environment in a single step versus adding packages over multiple steps.

@mdpiper Do you know why this might be? This also complicates things. You're saying you see problems when installing things in a single step, which is what this pull request is suggesting users do.

Maybe we should suggest users use *mamba* instead of *conda*. I wonder if this is a bug in how *conda* resolves dependencies.

mdpiper:
@mcflugen Oh, I'm sorry, my wording was atrocious! I'll go back and edit the comment.

mdpiper:
@mcflugen I guess it's conda's dependency resolution process. It solves an environment for one set of requirements, but then encounters conflicts when attempting to update the environment with an additional set of requirements.

sebastien-lenard:
Hi, I'm still new to github, and just realized that I couldn't have two simultaneous pull requests. 

The new commit and pull "FramedVoronoiGrid creation issue #1450" is fully independent from:

"Update developer install doc - issue #1440".

I hope that it's not a problem that they are in the same pull request.

mcflugen:
@sebastien-lenard Having them both in the same pull request is a problem—but a fixable one.

There should be no limit to the number of pull requests you can have at one time. What was the error you were seeing that prevented you from generating a second pull request?

mcflugen:
@sebastien-lenard I think I see the problem.

When developing with *landlab* you should always create a new topic branch for each change/addition you are working on. For example, you could have created two new branches named something like, `update-installation-docs`, and `add-framed-voronoi-grid` and then worked on each of those branches separately and created separate pull requests .You've, instead, been working only on the *master* branch.

sebastien-lenard:
Hi,
yes, I understood that afterwards. The new pull request button was simply deactivated in my case.

I had not made new modifications since my last commit. I think I managed to revert the commit and push by doing this locally:

```
git reset HEAD^
git push -f
```

sebastien-lenard:
This pull request is replaced by #1452 which has been pulled on a dedicated branch.

Issue #1450, opened 2022-08-06
Improving VoronoiDelaunayGrid by new FramedVoronoiGrid with fixed boundary and minimal distance between nodes


sebastien-lenard:
The present implementation of VoronoiDelaunayGrid let the user construct the set of x-y coordinates of nodes that the grid receives as input. This has two drawbacks:
- perimeter (and boundary) nodes are not calculated properly. The issue was raised by #885.
- the distances between two nodes can be very small. Particularly when the set of coordinates is pulled randomly. This occasionally triggers unstability in the models that includes diffusion or river incision (depending on the timestep of iterations).

To bypass these drawbacks, I propose to implement a new type of grid, the FramedVoronoiGrid, following a structure of classes and methods similar to the HexModelGrid. Four new classes are associated: the grid itself, FramedVoronoiGrid, a dual graph, DualFramedVoronoiGrid, which is parent of the grid, FramedVoronoiGraph, which is parent of DualFramedVoronoiGrid, and HorizontalRectVoronoiGraph, a static class implementing the methods necessary to instantiate a FramedVoronoiGraph.

The FramedVoronoiGrid starts by constructing a rectangular network of nodes. The boundary nodes are then fixed. This means that they are not determined by the ConvexHull as in the VoronoiDelaunayGrid. The core nodes are then moved around their position, in such a way that the distance between nodes don't exceed a certain threshold. While loosing the full potential of a fully random unstructured grid, the FramedVoronoiGrid is remains irregular and may fulfill some needs not currently fulfilled by the VoronoiDelaunayGrid or the HexModelGrid.

sebastien-lenard:
Hi @mcflugen I merged the #1454 and let me know when I can view the changes on the edge node ids you implemented in your branch.

sebastien-lenard:
@mcflugen it seems I missed your last comment on the pull request:

```seed: int or tuple of int, optional.
seeds used to generate the random x and y moves. This parameter is unused
when random_seed = True.
```
[@sebastien-lenard](https://github.com/sebastien-lenard) Sorry, I still don't get why a user would want to have separate seeds for the x and y moves. The only reason I can think of is if they wanted the moves to be the same (i.e. not independent).

How about, instead, something like the following (seed is either an int or None),

```
rng = np.random.default_rng(seed=seed)
x_moves = rng.uniform(-max_move[0], max_move[0], shape)
y_moves = rng.uniform(-max_move[1], max_move[1], shape)

[...]

PR #1454, opened 2022-08-06
Create FramedVoronoiGrid issue #1450

sebastien-lenard:
Creates FramedVoronoiGrid with general idea developed in #1450.
New notebook presenting all classes of grids.

May impact other grids and graphs with these files modified:
```
landlab/__init__.py
docs/build_grid_func_docs.py
landlab/graph/__init__.py
landlab/grid/__init__.py
tests/grid/test_constructors.py
```
NB: a revert of commits has been done on this branch, because it was constructed on a commit made on the master branch  (#1148).


sebastien-lenard:
@mcflugen please help me, I'm confused about the black-jupyter error and don't know how to amend it (no doc here: https://landlab.readthedocs.io/en/master/search.html?q=black-jupyter)

mcflugen:
@sebastien-lenard It looks like you just need to run *black* on your notebook. If you want to try using *nox* you should be able to run `nox -s lint` (you may have to *pip*-install *nox* if you don't already have it). Otherwise, you should be able to run *black* on your notebook directly (`black <path-to-your-notebook>`).

sebastien-lenard:
@mcflugen How can I generate the documentation from the docs/source/reference/grid/text_for_framed_voronoi.py.txt file? the workflows requires framed_voronoi.rst, but it's indicated this .rst file is auto-generated ...

mcflugen:
> @mcflugen How can I generate the documentation from the docs/source/reference/grid/text_for_framed_voronoi.py.txt file? the workflows requires framed_voronoi.rst, but it's indicated this .rst file is auto-generated ...

It looks like you have this worked out, is that right? There is a script that generates *framed_voronoi.rst* from *text_for_framed_voronoi.py.txt*. Don't worry too much about this, though, as I'm working to overhaul how this is done.

sebastien-lenard:
> > @mcflugen How can I generate the documentation from the docs/source/reference/grid/text_for_framed_voronoi.py.txt file? the workflows requires framed_voronoi.rst, but it's indicated this .rst file is auto-generated ...
> 
> It looks like you have this worked out, is that right? There is a script that generates _framed_voronoi.rst_ from _text_for_framed_voronoi.py.txt_. Don't worry too much about this, though, as I'm working to overhaul how this is done.

Actually, I duplicate hex.rst into framed_voronoi.rst, but that creates new doc problems. The workflow really requires an authentic framed_voronoi.rst and not a copy.

mcflugen:
> Actually, I duplicate hex.rst into framed_voronoi.rst, but that creates new doc problems. The workflow really requires an authentic framed_voronoi.rst and not a copy.

*hex.rst* is auto-generated from *text_for_hex.py.txt*. Because you just copied the hex version, you now have duplicate locations to the hex documentation and so *sphinx* doesn't know where to point references to the hex grid.

sebastien-lenard:
> > Actually, I duplicate hex.rst into framed_voronoi.rst, but that creates new doc problems. The workflow really requires an authentic framed_voronoi.rst and not a copy.
> 
> _hex.rst_ is auto-generated from _text_for_hex.py.txt_. Because you just copied the hex version, you now have duplicate locations to the hex documentation and so _sphinx_ doesn't know where to point references to the hex grid.

Alright @mcflugen , I removed the framed_voronoi.rst. I let you see how to make the workflow ok with that.

sebastien-lenard:
Hi @mcflugen! Did you find a solution for the generation of the framed_voronoi.rst file?

I'd also like to get an advice for my unit tests within this pull request, because they probably aren't what's expected.

mcflugen:
@sebastien-lenard When I added the *nox* session to build the docs, I didn't have it run the scripts that generate the grid documentation pages. I did this because I'm in the process of removing those scripts in favor a cleaner method. The reason the docs are building outside of this pull request is because the grid rst files have been added to the repository. I think you can get the docs to build by adding your *framed_voronoi.rst* file to the repository.

If you run `make html` from the *docs* folder on your computer, the docs should build locally, is that right (if you look in *noxfile.py* you'll notice I use *sphinx-build* rather than *make*)? Sorry about that.

I'll have a look at your unit tests later this afternoon.

sebastien-lenard:
> If you run `make html` from the _docs_ folder on your computer, the docs should build locally, is that right (if you look in _noxfile.py_ you'll notice I use _sphinx-build_ rather than _make_)? Sorry about that.

That should, yes @mcflugen. But the problem is that sphinx is not installed on the Jupyter hub (my landlab dev platform). @mdpiper do you think that might be possible to install sphinx on the jupyter hub?

mdpiper:
> @mdpiper do you think that might be possible to install sphinx on the jupyter hub?

Yes, I can! Do you think a new kernel, "development", would be useful, @sebastien-lenard?

sebastien-lenard:
> > @mdpiper do you think that might be possible to install sphinx on the jupyter hub?
> 
> Yes, I can! Do you think a new kernel, "development", would be useful, @sebastien-lenard?

Yes, that would be definitely great to develop notebooks and more generally.

mcflugen:
> sphinx is not installed on the Jupyter hub

@sebastien-lenard can you create your own environment on the JupyterHub and install it there? I thought that was what you were doing, actually.

@mdpiper is the above possible or do users not have enough space to create environments in their home directories?

mdpiper:
@mcflugen, @sebastien-lenard I think it might be nice to have an environment (mapped to a kernel) with all needed development tools, so that users don't have to create environments in their home directories. This should save disk space in the long run.

sebastien-lenard:
@mdpiper  Here's the conda list in my jupyter landlab dev environment:

[sebastien_landlab_dev_conda_list_2022-08-19.txt](https://github.com/landlab/landlab/files/9384243/sebastien_landlab_dev_conda_list_2022-08-19.txt)

sebastien-lenard:
> @mdpiper Here's the conda list in my jupyter landlab dev environment:
> 
> [sebastien_landlab_dev_conda_list_2022-08-19.txt](https://github.com/landlab/landlab/files/9384243/sebastien_landlab_dev_conda_list_2022-08-19.txt)

@mdpiper I tested a bit the new Dev environment. That's great! :heart_eyes:

Remarks:
a/ Presently, I noticed that these two libs, which are used by the auto-documentation can be added
sphinx_jinja
sphinxcontrib-towncrier

b/ It's not important right now, but the cython doc says that cython can be called in Jupyter notebooks through the syntax:
```
%load_ext Cython
```
But this doesn't seem to work 
https://cython.readthedocs.io/en/latest/src/quickstart/build.html?highlight=notebook#using-the-jupyter-notebook

sebastien-lenard:
Hi @mcflugen!
Could you have a look to my unit tests before I merge this request?

Thank you

landlab/graph/framed_voronoi/framed_voronoi.py
```diff
@@ -0,0 +1,483 @@
...
+            around their position cannot be above this threshold:
+            (xy_spacing - xy_min_spacing) /2
+            If float, same minimal spacing for x and y.
+        random_seed: bool, optional.
+            If True, the moves of coordinates are completely random. False is used
+            when reproducibility of moves is needed. Move pseudo-randomness is
+            then controlled by the parameter seed.
+        seed: int or tuple of int, optional.
+            Seeds used to generate the random x and y moves. This parameter is unused
+            when random_seed = True.
```
mcflugen, line 85: As written, *seed* must be a *tuple* not an *int*. It seems to me though that *seed* should only be allowed to be an *int*. I don't see the usefulness of having a separate seed for the *x* and *y* moves.

landlab/graph/framed_voronoi/framed_voronoi.py
```diff
@@ -0,0 +1,483 @@
...
+            xy_random_generator[0].random(n_core_nodes) * max_move[0] * 2 - max_move[0],
+            xy_random_generator[1].random(n_core_nodes) * max_move[1] * 2 - max_move[1],
+        )
+
+        k = 0
+        for i in range(1, n_rows - 1):
+            for j in range(1, n_cols - 1):
+                x_of_node[i, j] += x_y_random_move[0][k]
+                y_of_node[i, j] += x_y_random_move[1][k]
+                k += 1
```
mcflugen, line 135: I think you can remove these loops with something like,
```python
x_of_node[1:-1, 1:-1] += xy_random_move[0]
y_of_node[1:-1, 1:-1] += xy_random_move[1]
```

landlab/graph/framed_voronoi/framed_voronoi.py
```diff
@@ -0,0 +1,483 @@
...
+        random_seed=True,
+        seed=(200, 500),
+    ):
+        """Create the graph.
+
+        Parameters
+        ----------
+        shape : int or tuple of int
+            For a rectangular layout, number of rows and columns of nodes.
+            If int, rows number = columns number = value
```
mcflugen, line 303: I don't think we use this pattern for any other grid (i.e. that *shape* can be an int or a tuple), do we? I think shape is always a *tuple*.

landlab/graph/framed_voronoi/framed_voronoi.py
```diff
@@ -0,0 +1,483 @@
...
+
+        >>> from landlab.graph import FramedVoronoiGraph
+        >>> graph = FramedVoronoiGraph((3, 2))
+        >>> graph.number_of_nodes
+        6
+        """
+        # 1. Check and format input arguments
+        #####################################
+        try:
+            shape_ = np.asfarray(np.broadcast_to(shape, 2))
```
mcflugen, line 346: Why do you convert *shape* to an array of floats?

mcflugen (commented):
@sebastien-lenard Just a few initial comments. I'll continue to review this tomorrow but, so far, looking good.

mcflugen (commented):
@sebastien-lenard I think this is looking good. The only other thing I think I would add is to see if you could remove some of the tests that rely on the random positioning of nodes.

sebastien-lenard:
> @sebastien-lenard I think this is looking good. The only other thing I think I would add is to see if you could remove some of the tests that rely on the random positioning of nodes.

Alright thank you a lot @mcflugen I'll do this.

landlab/grid/framed_voronoi.py
```diff
@@ -0,0 +1,159 @@
...
+        orientation : string, optional
+            'horizontal' only
+        node_layout : string, optional
+            'rect' only. The grid layout of nodes.
+        xy_min_spacing: float or tuple of float, optional
+            Final minimal spacing between nodes. Random moves of the core nodes
+            around their position cannot be above this threshold:
+            (xy_spacing - xy_min_spacing) /2
+            If float, same minimal spacing for x and y.
+        random_seed: bool, optional
```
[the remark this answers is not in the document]
sebastien-lenard: Done

landlab/grid/framed_voronoi.py
```diff
@@ -0,0 +1,159 @@
...
+
+    Inheritance diagram
+    x xxxxxxxxxxxxxxx x
+    x                  ModelGrid                                                                       DelaunayGraph
+    x FramedVoronoiGrid /                                                                               /
+    x                  |                        FramedVoronoiGraph (Layout: HorizontalRectVoronoiGraph)
+    x                  DualFramedVoronoiGraph /
+    x                                         \
+    x                                          DualGraph
+    x                                                   ~ use of static Graph.sort()
```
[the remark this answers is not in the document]
sebastien-lenard: Done (as a """ """ string separate from the docstring

landlab/graph/framed_voronoi/framed_voronoi.py
```diff
@@ -0,0 +1,483 @@
...
+        seed: int or tuple of int, optional
+            Seeds used to generate the random x and y moves. This parameter is unused
+            when random_seed = True.
+
+        Returns
+        -------
+        DualFramedVoronoiGraph
+            A newly-created graph.
+
+         Examples
```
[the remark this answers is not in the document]
sebastien-lenard: Done

landlab/graph/framed_voronoi/framed_voronoi.py
```diff
@@ -0,0 +1,483 @@
...
+    def corner_nodes(shape):
+        """
+        Parameters
+        ----------
+        shape: tuple of int
+            Number of rows and number of columns
+
+        Returns
+        -------
+        ndarray of Int
```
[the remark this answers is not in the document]
sebastien-lenard: amended

landlab/graph/framed_voronoi/framed_voronoi.py
sebastien-lenard: i simplified by requiring a tuple of int only (or None).

x moves should be independent from y moves. 

Implementation pull a random move depending on the node id. If only one seed is used, this may imply to pull a random move on the node id for x and on the node id + number of nodes for y, which is less clearer in my view than just have independent seeds for x and y and make the pull on the node id similarly for the x and y move

(sorry if I expressed it in a not so clear way :flushed:)

landlab/graph/framed_voronoi/framed_voronoi.py
sebastien-lenard: now requires tuple of int only

landlab/graph/framed_voronoi/framed_voronoi.py
sebastien-lenard: It was a mistake. corrected

landlab/graph/framed_voronoi/framed_voronoi.py
sebastien-lenard: Done. Actually you have to reshape x_y_random_move so that works.

landlab/grid/framed_voronoi.py
```diff
@@ -0,0 +1,159 @@
...
+            If int, rows number = columns number = value
+        xy_spacing : float or tuple of float, optional
+            Node spacing along x and y coordinates. If float, same spacing at x and y.
+        xy_of_lower_left : tuple, optional
+            Minimum x-of-node and y-of-node values. Depending on the grid
+            no node may be present at this coordinate. Default is (0., 0.).
+        orientation : string, optional
+            'horizontal' only
+        node_layout : string, optional
+            'rect' only. The grid layout of nodes.
```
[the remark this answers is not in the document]
sebastien-lenard: removed.

tests/grid/test_constructors.py
```diff
@@ -199,6 +200,141 @@ def test_network_from_file():
...
+            11.0,
+            21.0,
+            31.0,
+            41.0,
+            51.0,
+            1.0,
+            39.38915671,
+            11.73417072,
+            28.64950826,
+            21.81959986,
```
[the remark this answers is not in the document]
sebastien-lenard: Tests on interior points have been removed.

sebastien-lenard (commented):
All changes required by @mcflugen's review have been implemented (except the seed for random x-y moves kept as a tuple of int).

landlab/graph/framed_voronoi/framed_voronoi.py
```diff
@@ -0,0 +1,474 @@
...
+            y_of_node[1:-1, 1:-1] += np.reshape(
+                x_y_random_move[1], (n_rows - 2, n_cols - 2)
+            )
+        # Control the node id attribution for left and right edge. For instance, for a 3x3 grid,
+        # make sure that node 3 is at the left of the 2nd row and node 5 at the right.
+        # For this, for each core row, set y of the leftmost node as the minimal y of the row
+        # and set y of the rightmost node as the maximal y of the row
+        for i in range(1, n_rows - 1):
+            y_of_node[i, 0] -= max_move[1] + 1.0e-3
+            y_of_node[i, n_cols - 1] += max_move[1] + 1.0e-3
```
mcflugen, line 143: @sebastien-lenard I believe you do this only so you can easily get the perimeter nodes, is that correct? It may not matter for what you are using this for but it seems like it may introduce some bias to the node layout that may matter in some situations.

I've "fixed" this (i.e. all edge nodes are not moved from their initial position) on my branch and was thinking I would create a pull request with those changes after this pull request is merged. That way you can review what I've done before incorporating it.

landlab/graph/framed_voronoi/framed_voronoi.py
mcflugen: @sebastien-lenard Sorry, I still don't get why a user would want to have separate seeds for the *x* and *y* moves. The only reason I can think of is if they wanted the moves to be the same (i.e. *not* independent).

How about, instead, something like the following (*seed* is either an `int` or `None`),
```python
rng = np.random.default_rng(seed=seed)
x_moves = rng.uniform(-max_move[0], max_move[0], shape)
y_moves = rng.uniform(-max_move[1], max_move[1], shape)

x_of_node[1:-1, 1:-1] += x_moves[1:-1, 1:-1]
y_of_node[1:-1, 1:-1] += y_moves[1:-1, 1:-1]
```

landlab/graph/framed_voronoi/framed_voronoi.py
sebastien-lenard: Not only I want the list of perimeter nodes, but I also want to know what is the node id at the right of the grid and what is the node id at the left, at a certain "row").

PR #1455, opened 2022-08-06
Minor update developer install doc try 2 #1440

sebastien-lenard:
Update the developer install documentation to make sure that the creation of the environment takes into account all dependencies in the same call. Solve #1440.

----------------------------------------------------------------------------------------------------------

Sorry for the little mess that was with this pull request :/ but everything should be neat right now.

This pull request replaces #1448, #1452, #1453. #1448 had a commit on the master branch rather than a dedicated branch. #1452 was a try for a commit on the sl/1440_developer_install_doc branch. #1453 was a try to revert the commit on the master branch.

Finally, I delete all the commits on the branchs sl/1440_developer_install_doc branch and sl/1450_framed_voronoi_grid which were based on the initial commit on the master branch. I also delete the commit on the master branch. This was done by:

1- git checkout branch_x (for the 2 branches and the master)
2- git reset HEAD^ and git push -f for each commit until back to the initial version.


sebastien-lenard:
Hi @mdpiper may you review and validate this? Thank you!

docs/source/development/install/install.rst
```diff
@@ -65,10 +65,8 @@ The next step is it create this environment, called *landlab_dev*.
 .. code-block:: bash
 
    $ conda create -n landlab_dev python
-   $ conda install -n landlab_dev --file=requirements.txt -c conda-forge
-   $ conda install -n landlab_dev --file=requirements-testing.txt -c conda-forge
-   $ conda install -n landlab_dev --file=requirements-notebooks.txt -c conda-forge
-   $ conda install -n landlab_dev --file=requirements-dev.txt -c conda-forge
+   $ conda create -n landlab_dev -c conda-forge --file=requirements.txt --file=requirements-dev.txt --file=requirements-testing.txt --file=requirements-notebooks.txt
```
mdpiper, line 68: If you remove the "=" after each `--file` flag, then shell completion works.

sebastien-lenard: @mdpiper should I let only the first --file this way?
```
conda create -n landlab_dev -c conda-forge --file requirements.txt requirements-dev.txt requirements-testing.txt requirements-notebooks.txt
```

mdpiper: No, unfortunately each requirements file requires its own `--file` flag. (I've tried this before and it failed.)

sebastien-lenard: Alright @mdpiper I commit the changes :smiley:

mdpiper: Thanks @sebastien-lenard!

mdpiper (dismissed):

sebastien-lenard:
> @mdpiper, @sebastien-lenard Are we sure this is the only solution? I'm not completely satisfied with this as it requires new users to install a lot of extra requirements, which they may not need—the notebook requirements, in particular.
> 
> @sebastien-lenard I can't seem to find the original issue that described this particular problem. Do you know if there is one?

Hi @mcflugen the issue is #1440. The method which is currently described in the doc makes crash the install (and occasionally the jupyter server session for the user) because requirements are not solved simultaneously. 

Do you remember who wrote all these requirements, basically?

mcflugen:
> the issue is https://github.com/landlab/landlab/issues/1440.

I was aware of that issue but I thought there may have been a previous issue that described the steps necessary to reproduce the error and a description of the error. I'm unable to reproduce the error, at least as I understand the problem. It's possible that the problem has gone away with an update of *conda* or one of *landlab* requirements.

For example, the following works just fine for me,
```bash
$ conda create -n landlab-dev
$ conda activate landlab-dev
$ conda install python
$ conda install --file requirements.txt
$ conda install --file requirements-dev.txt
$ conda install --file requirements-testing.txt
$ conda install --file requirements-notebooks.txt
```

Can you reproduce the problem?

> Do you remember who wrote all these requirements, basically?

The *landlab* team mostly. I'm sure I wrote many of them. For the notebook requirements, it was mainly the notebook contributors.

sebastien-lenard:
@mcflugen ```
> $ conda install --file requirements.txt
> $ conda install --file requirements-dev.txt
> $ conda install --file requirements-testing.txt
> $ conda install --file requirements-notebooks.txt
> ```
In the doc, there's no explicit indication that one of these instructions is optional. I assume most new users will try the four requirements.

So if you still think it's necessary to split, we can propose:
If you want to dev only, install this way:
> $ conda install --file requirements.txt --file requirements-dev.txt --file requirements-testing.txt

If you want to dev + use the notebooks, install this way:
> $ conda install --file requirements.txt --file requirements-dev.txt --file requirements-testing.txt --file requirements-notebooks.txt

But I don't think many people will dev without simultaneously looking a bit the notebooks.

Issue #1459, opened 2022-08-09
POC: unit tests with pytest on cython def, cdef and cpdef functions / Module xxx has no attribute '__pyx_capi__'


sebastien-lenard:
**1. Is there any unit test on the cython code?**

With research trying to running models with grids of several 100,000 nodes over more than 1,000 iterations, I guess that the cython part of Landlab will become more and more crucial. However, I'm not sure if there's any unit test on cython code, which is ... risky?

**2. Implementation for simple case: **def** functions within cython code.**
Testing a **def** function written a cython file .pyx is straightforward. You only need one test file which will be called by pytest. This test file import the cython as a module and use it as it is used in normal python scripts. e.g.
```
[...]
import landlab.components.flow_director.cfuncs as _cfuncs
[...]
_cfuncs.adjust_flow_receivers(src_nodes, dst_nodes, z, link_slope, g.active_links,
    receiver, receiver_link, steepest_slope)
assert [...]
```
Note that the correct working of pytest for such a test file during pull request needs to be validated.

**3. Implementation for a more difficult case: **cdef**/**cpdef** functions within a cython code.** 
Let's call the cython file **cython_1.pyx** and the function **cdef void travel(void)**
This was definitely not straightforward. My first tries led to the undocumented and horrific error *Module xxx has no attribute '__pyx_capi__'*. I didn't find any post online which tangibly helped me to correct this problem.

However, I fell on a somewhat hidden webpage that helps me a lot:
https://stuff.mit.edu/afs/sipb/project/sage/packages/cython-0.9.6.8.b/Doc/sharing.html

Three problems are to be overcome:
- you can't directly call a **cdef** function from a python script, because that function is compiled as a C function. To call the **cdef** function, you need an intermediary cython file I'll name **test_cython_1.pyx**, where you implement a **def** function which will call the **cdef** function of **cython_1.pyx**. You need to import **cython_1** and to do this, you write:
```
cimport cython_1;  import cython_1
def test_travel():
[...]
cython_1.travel()
assert[...]
```

- To call a **cdef** or **cpdef** function from another cython file, you need to declare the function in a **.pxd** file. So here, you need to write a **cython_1.pxd** file this way:
```
cimport cython
cdef void travel(void)
```
Note that you have to replace the void by the actual signature of the function return and input parameter (ints, doubles, etc...)

- Since a cython file cannot be called by pytest, you have to create a classic python script which will call the **def** function. Let's call it **test_travel.py**. It'll contain:
```
[...]

PR #1464, opened 2022-08-09
Create pytest on the cython func of FlowDirectorSteepest #1459 

sebastien-lenard:
To address the possibility that unit tests also cover cython code #1459 point 2, this small delivery of a unit test file for the function def adjust_flow_receivers which underlies the FlowDirectorSteepest.run_one_step().

NB: Retry of pull request #1460 which was built with a commit initially for another pull request #1454


sebastien-lenard:
Hi @mcflugen ! This request is ready for review, if you have time. Thanks!

mcflugen:
@sebastien-lenard I made a couple tweaks, the main one was to move the test into a function. If the test isn't in a function, then it will get run when *pytest* is looking for tests rather than when it runs the tests.

Would you be able to rewrite the test without using random input data? Because you're using random input, it's not clear to me that the *adjust_flow_receivers* function is working correctly. A better test would be one where you know what the correct output should be before you run the test.

sebastien-lenard:
> @sebastien-lenard I made a couple tweaks, the main one was to move the test into a function. If the test isn't in a function, then it will get run when _pytest_ is looking for tests rather than when it runs the tests.
> 
> Would you be able to rewrite the test without using random input data? Because you're using random input, it's not clear to me that the _adjust_flow_receivers_ function is working correctly. A better test would be one where you know what the correct output should be before you run the test.

Thank you @mcflugen.

The topography is actually not random since the seed is fixed to 500. Then, the random_generator.random(nodes_n)) always produce the same array. Would you still prefer I replace the syntax?

mcflugen:
> Would you still prefer I replace the syntax?

I would. Even though the test is repeatable, it's not clear to me that *adjust_flow_receivers* is generating the correct output. I may be missing something but it appears that you created some random input, ran it through *adjust_flow_receivers*, and then copied that output into you test. Instead, I would like to see a test that uses input for which you know the output before running the tested function.

sebastien-lenard:
@mcflugen I updated the test file, could you validate it? thanks :)

mcflugen:
@sebastien-lenard I've add three more unit tests for *adjust_flow_receivers*: two for a dipping plane on a raster grid, and one for a dipping plane on a hex grid.

sebastien-lenard:
Thank you @mcflugen, I tried your new version and it works. May you validate this request or does it remain something to do?

sebastien-lenard:
Hi @mcflugen, May you validate this request or does it remain something to do?
Thank you!

sebastien-lenard:
Hi @mcflugen! A new problem about coverage is now raised and seems unrelated to this request which files are tests:
```
Coveralls - false — Coverage decreased (-0.0%) to 85.934%
```
I'll perform the merge 9/2 except if you have any remarks.
Thanks!

mcflugen:
@sebastien-lenard I've changed the threshold for the decrease in percent coverage (was 0.0%, now 0.1%) so, hopefully, we won't see this error again.

PR #1468, opened 2022-08-10
Update setup.py for enhanced compilation of cython files #1467 - to be cancelled

sebastien-lenard:
Contains a new implementation of the compilation of cython files according #1467. The compilation can be parameterized at the global and file levels with a setup_cython.yml and a setup_XXX.yml conf files.

This request includes a modification on landlab/components/flow_director/cfuncs to prevent numpy warning during compilations and to show the possibilities of the file level conf file.

Issue #1467, opened 2022-08-10
New implementation of the compilation of the cython files with cythonize


sebastien-lenard:
Since #869, landlab install collects the cython files and compiles the cython extensions in executables. This is done with the help of the **setup.py** script. Files are compiled in c-executables by including the numpy library and without the "new" cython function "cythonize()". The setup script works well but remains limited, whatever its use in standard or dev mode. 

I propose a major update of this script, with the double aim to ease its parametering at the global level and to specify some compilation instructions at the cython file level. For this, a new setup_cython.yml configuration file will be designed. The template of the setup_xxx.yml configuration file at the cython file level will be also supplied.

This implementation adds new functionalities at the **global** (cython) level:

- to force the recompilation of all files, whatever the previous compilation date and edition date of the cython files.
- to restrict the directories where to look the extensions: e.g. to landlab and test directories and sub.
- to exclude some directories according to the text within them: e.g. .ipynb_checkpoints
- to exclude some files according to the text within their filename
- to parameter the build of the extensions with cythonize arguments, which include:
- - the compilation of the extensions using Python 3 (Python 2 is the default today)

And at the **file** level:
- to include c or cpp file codes
- to compile in c++ (not working when include c++ libraries without the cythonize function)
- to add extra c/c++ compilation arguments, such as -fopenmp which opens the avenue to parallelization on several computer cores in Landlab components.

sebastien-lenard:
@mdpiper, using extra-compilation options for the c/c++ compiler (as described in the specs of this issue) is a bit tricky. There are two famous compilers, gcc and clang, and their arguments can be slightly different. For instance, gcc can accept -fopenmp (to enable multithreading). But clang rejects this argument alone. One solution is to specify the complementary argument -Xpreprocessor, which yields the string of arguments "-Xpreprocessor -fopenmp". This string works both for gcc and clang.

Additional remark: on the Jupyterhub, gcc is the compiler, while on the github action platform, clang is the compiler. It might be interesting in the future to have clang rather than gcc on the Jupyterhub.

mdpiper:
@sebastien-lenard I agree: this will be tricky, there's no way around it. I'm biased--I like to use the [conda compilers.](https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html) But not everyone uses conda, so this isn't a solution.

[...]

PR #1471, opened 2022-08-15
Update setup.py for enhanced compilation of cython files #1467 Version 2

sebastien-lenard:
Spinoff of pull request #1468.

Contains a new implementation of the compilation of cython files according https://github.com/landlab/landlab/issues/1467. Features:
- The compilation can be parameterized at the  file levels with # instructions at the top of the .pyx file. (This presently doesn't work).
- The compilation can include other directories (and subs) than landlab. Here it includes the directory tests.
- The compilation is done with Python 3 and include docstrings

This request includes a use case with .pyx files and test .pyx and py files. (These files are required for the development of the FlowRouter which will follow.

This spinoff follows discussions with @mdpiper  and @mcflugen that highlight it may be preferable to keep the setup.py file as small as possible, without calling yaml library.


mcflugen:
@sebastien-lenard Thanks for this. I made a few minor changes to *setup.py*.

sebastien-lenard:
@mcflugen I still have a problem with your version of setup.py, which includes hidden directories and makes break my compilation on jupyterhub. This is why I added the line:
```
_dirs[:] = [d for d in _dirs if not d[0] == "."]
```

mcflugen:
> @mcflugen I still have a problem with your version of setup.py, which includes hidden directories and makes break my compilation on jupyterhub.

Interesting. I have not seen that before. Where do the hidden folders containing *.pyx* files come from on the JupyterHub?

sebastien-lenard:
> > @mcflugen I still have a problem with your version of setup.py, which includes hidden directories and makes break my compilation on jupyterhub.
> 
> Interesting. I have not seen that before. Where do the hidden folders containing _.pyx_ files come from on the JupyterHub?

@mcflugen each time there's a compile error, this generate a .ipynb_checkpoints directory (in each subdirectory where the compilation failed. I'm not sure the directory is generated when compilation succeeds.

mcflugen:
> @mcflugen each time there's a compile error, this generate a .ipynb_checkpoints directory (in each subdirectory where the compilation failed. I'm not sure the directory is generated when compilation succeeds.

Are you compiling the extensions from within a Jupyter notebook rather than from a terminal?

sebastien-lenard:
> > @mcflugen each time there's a compile error, this generate a .ipynb_checkpoints directory (in each subdirectory where the compilation failed. I'm not sure the directory is generated when compilation succeeds.
> 
> Are you compiling the extensions from within a Jupyter notebook rather than from a terminal?

@mcflugen I compile via a console, under a copy of the csdms environment, either using python setup.py build_ext --inplace or pip install -e .

sebastien-lenard:
@mcflugen clang compilation on macos fails (Test / Run the tests (macos-latest, 3.10) (pull_request) ), apparently because compilation is under an older version of c++ and c++11. I can try to force it through compilation arguments, but maybe there's an explanation for this difference between windows/ubuntu and macos?

sebastien-lenard (commented):
All changes concern new files, except setup.py. This file has been edited by @mcflugen and reviewed + tested locally by me. Everything seems correct.

Issue #1461, opened 2022-08-09
Flow directions and depression filling for all type of grids: specs for a new, more scalable, FlowRouter component


sebastien-lenard:
When run over tens of thousands of years, landscape evolution can create lasting depressions on topography. In nature, these depressions can be breached in a relatively short time (a few hundreds to thousands of years). But without a specific bit of code, these depressions can stay much longer in a landscape evolution model, due to the relatively slow behavior of the stream power law. This is why we usually reconstruct the flow by overcoming these depressions, to make sure that all points of the landscape are drained to a base-level outlet and make the model run as expected.

Presently, we have three options:

**1. Run the PriorityFloodFlowRouter on RasterModelGrids.**

The run first calls the richdem python package to obtain a depression-free topographic surface. Then it calculates flow directions and accumulations based on this surface, either with single or multiple flow directions. The algorithms used for these calculations are similar to the ones used by the FlowDirectors and FlowAccumulator components. The richdem python package wraps a C++ library that implements a priority flood flow algorithm to direct flow over depressions, starting from base-level. This algorithm is assumed to be the most efficient algorithm to calculate flow directions with depressions (Barnes et al., 2014; https://richdem.readthedocs.io/en/latest/intro.html). The PriorityFloodFlowRouter delegates part of its execution (almost all loops on arrays) to cython code. Cython helps to translate and precompile loops into C code. This makes the run of the PriorityFloodFlowRouter faster than if it was only coded in Python.

Note that the use of an external package maintained only by a small community might be slightly risky given the very frequent usage of flow direction calculations in landscape evolution models in Landlab.

**2. Run the DepressionFinderAndRouter with a FlowDirector on all grids.**

The run first calls a FlowDirector to associate receivers and donors in a single (steepest slope) or multiple flow direction way. Then the DepressionFinderAndRouter maps all depressions and solve their outlet and recalculates flow directions down to a base-level outlet. The algorithm used by the FlowDirectorSteepest is very efficient. The algorithms used in DepressionFinderAndRouter are not efficient for landscapes evolving with some depressions. Additionally, the DepressionFinderAndRouter is fully coded in Python, which make it very very slow. The component doesn't have a specific notebook for its usage.

**3. Run the LakeMapperBarnes with a FlowDirector on all grids (?)**

[...]

PR #1474, opened 2022-08-17
Add nuke and requirements sessions to the noxfile

mcflugen:
This pull request adds two new sessions to the *landlab* *nox* file: *requirements* and *nuke*. The *requirements* session creates the various requirements files from the requirements listed in *pyproject.toml* (by running the *requirements.py* script). The *nuke* session is intended to *really* clean up a working repository by running the *clean* session followed by removing *jupyter* notebook checkpoint directories.

Examples:
```
$ nox -s requirements  # recreate requirements.txt, requirements-docs.txt, etc.
$ nox -s nuke  # really clean the entire repository
$ nox -s nuke -- landlab tests  # only run on the landlab and tests folders
```


sebastien-lenard:
@mcflugen I'm a bit confused. Is there any doc on nox and how to use it within the landlab dev environment?

mcflugen:
@sebastien-lenard I think the best place to look for using *nox* is probably just their [main documentation](https://nox.thea.codes).

For your *landlab* development, you may (or may not) find the following useful:
* Run all of the tests in an isolated environment,
```bash
$ nox -s tests
```
* Check for lint,
```bash
$ nox -s lint
```
* Remove auto-generated files (`*.pyo`, `__pycache__`, `*.so`, etc),
```bash
$ nox -s clean
```
* Clean as above but also remove notebook checkpoints, running only in the *landlab* folder,
```bash
$ nox -s nuke -- landlab
```

sebastien-lenard:
Alright @mcflugen. So if I have a problem with notebook checkpoints in my landlab dev, I can run 
```
$ nox -s nuke -- landlab
```
and then 
```
python setup.py build_ext --inplace
```
without having problem with these hidden .ipynb files ? #1461 #1467

But does 
```
$ nox -s nuke -- landlab
```
also removes all compiled files? because I don't want to recompile everything each time I want to compile one file to test it ...

mcflugen:
> also removes all compiled files?

It does because that's sometimes something we want to do. However, we can certainly add a new session or modify the existing ones to do different tasks.

mcflugen:
@sebastien-lenard I'll update this pull request with some reorganization of the cleaning sessions to include your use case.

sebastien-lenard:
Hi @mcflugen 
these notebook entrypoints are also created in the docs (on the Jupyter hub), I don't know why ...

- docs/source/reference/.ipynb_checkpoints/
- docs/source/development/install/.ipynb_checkpoints/
- docs/source/.ipynb_checkpoints/

mcflugen:
I've included some additional changes in the pull request,
* removed more lint, this time mostly in *pyx* files
* changed our ci to use *nox*
* added two new linters: one that runs *pyroma* and one that runs *check-manifest* (which was inspired from @mdpiper's recent pull request #1534)

@mdpiper Would you mind having a look at this when you get a chance?

mdpiper (approved):

mdpiper:
👍

Issue #1484, opened 2022-09-06
Bug in imshowhs_grid


loroberge:
@BCampforts imshowhs_grid occasionally inverts the plot in both the east-west and north-south directions. For example, instead of plotting the point 0,0 at the bottom left it will plot in the top right. This problem seems to occur only when there is a drape for landslide erosion and/or deposition but there is no data to plot on the drape.

mcflugen:
@loroberge Could please provide a [minimal, reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) for your issue?

loroberge:
Below is as minimal an example as I could come up with. I use a 12x12 grid with an outlet at Node ID 0 (i.e. at coordinates 0,0).

I've annotated this in the code, but adding here as well for clarity: I first plot a DEM with imshow_grid. I then run the BedrockLandslider twice. The seed is set such that no landslides occur in the first timestep and then 2 landslides occur in the second timestep. After each timestep, I use imshowhs_grid to overlay the "landslide__erosion" and "landslide__deposition" fields on top of a hillshade. When no landslides occur, the two fields are filled with zeros and the plot has an inverted y-axis. When landslides occur, the two fields have non-zero values at some locations and the axes are plotted correctly.

Another issue that I did not see before: the hillshade base layer does not plot data from the rows and columns adjacent to the boundary nodes. However, the two drape layers do plot the correct data. 

```python
# %% Import libraries, generate model grid and fields, instantiate components

from landlab import RasterModelGrid, imshow_grid, imshowhs_grid
from landlab.components import PriorityFloodFlowRouter, BedrockLandslider
import numpy as np
import matplotlib as mpl

cmap_terrain = mpl.cm.get_cmap("terrain").copy()
cmap_erosion = mpl.cm.get_cmap("Wistia").copy()
cmap_deposition = mpl.cm.get_cmap("winter_r").copy()
    
mg = RasterModelGrid((12, 12), 25)
mg.axis_units = ('m', 'm')
mg.set_status_at_node_on_edges(right=4,
                               top=4,
                               left=4,
                               bottom=4)
mg.status_at_node[0] = mg.BC_NODE_IS_FIXED_VALUE

topo = np.array(
        [0.00000000e+00, 7.79918792e-02, 4.38409231e-02, 7.23465178e-02, 9.77989512e-02, 5.38495870e-02,
[...]

PR #1492, opened 2022-09-14
Bc/imshowhs

BCampforts:
### Description
Fixed some bugs in the *imshowhs_grid* plotting function, also addressing issue #1484 
Fix 1: Before, *imshowhs_grid* inverted the plot in north-south directions, when a no-data drape was provided. This issue is solved.
Fix 2: Axis tick marks were slightly off. This issue is fixed.
Fix 3: The hillshade base layer was not plotting data from the rows and columns adjacent to the boundary nodes. This issue is solved.
Fix 4: Boundary nodes were not indicated if requested. This issue is now solved. For now, boundary nodes can be displayed in white (``color_for_closed = 'white' ``) or not at all (``color_for_closed = None ``)

### Checklist - did you ...

- [x] All tests have passed?


BCampforts:
Hi @mcflugen, I do not manage to build the docs for this. Not sure how to resolve it (see checks; Documentation)

landlab/plot/imshowhs.py
```diff
@@ -454,6 +456,8 @@ def _imshowhs_grid_values(
     cmap = plt.get_cmap(cmap)
 
     if color_for_closed is not None:
+        if color_for_closed != "white":
```
mcflugen, line 459: Why limit *color_for_closed* to only *"white"* or *None*? It seems like an [matplotlib color](https://matplotlib.org/stable/tutorials/colors/colors.html#sphx-glr-tutorials-colors-colors-py) should work.

BCampforts: Good point, to plot the hillshade with an extra layer, I first create an RGB array and next change the transparency before plotting using pyplot's imshow. Didn't know about a quick fix to make this working with masks. 
See: 
`rgb[:, :, 3][values.mask] = 0`
` ima = ax1.imshow(rgb, extent=extent, **kwds)`

mcflugen:
> Hi @mcflugen, I do not manage to build the docs for this. Not sure how to resolve it (see checks; Documentation)

The problem was that your news fragment file was not properly-formatted reStructuredText. I think I've fixed it now, though. In addition, since you fixed multiple bugs, you should include multiple news fragments (one for each bug fix). I've done this for you but you should probably read them over to make sure they make sense.

I just realized that each of the news fragments refer to #1484. If some of those bug fixes don't relate to that issue, we should change their file name to point to their issue/pull request.

landlab/plot/imshowhs.py
mcflugen: I think [to_rgba](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.to_rgba.html?highlight=to_rgba) should do what you want.
```python
>>> from matplotlib.colors import to_rgba
>>> to_rgba("w")
(1.0, 1.0, 1.0, 1)
>>> to_rgba("w", alpha=0.0)
(1.0, 1.0, 1.0, 0.0)
>>> to_rgba("c")
(0.0, 0.0, 0.0, 1)
>>> to_rgba("k")
(0.0, 0.75, 0.75, 1)
```

BCampforts: Yup, that's it!

mcflugen:
@BCampforts I made some edits to the *docstrings* of the *imshowhs* functions. Could you please have a look?

I just copied the *docstring* for *imshowhs_grid*  to *imshowhs_grid_at_node*, is that ok? That seems to be what you had done. If they are the same, that indicates to me there should probably be just one function.

mcflugen (approved):

PR #1501, opened 2022-10-13
Gt/generic ls sed xporter

gregtucker:
### Description

This PR introduces a new, bare-bones landscape evolution component called `AreaSlopeTransporter`. 

The `AreaSlopeTransporter` is a landscape evolution component that models the time rate of change of elevation at a set of grid nodes, each of which has a defined contributing drainage area $A$ (field `drainage_area`) and a local steepest-descent slope gradient, $S$, defined from the node itself toward one of its neighboring nodes. The drainage area and slope can be computed with a drainage-routing component such as `FlowAccumulator` or `PriorityFloodFlowAccumulator`. The component is designed to function as an integral part of a transport-limited landscape evolution model in the spirit of Willgoose et al. (1991a,b,c,d).

Once $A$ and $S$ have been calculated, the volumetric outgoing sediment flux at each grid node is calculated using 

$Q_{out} = k_s A^m S^n$

The component's default parameters are $m=1.4$, $n=2.1$, and $k_s = 0.0055$ (with time units of years). These defaults are based on Willgoose et al. (1991a), who used the Einstein-Brown sediment transport formula as a basis, but here modified to include an empirical sub-grid-cell channel width. The derivation is provided in an accompanying notebook (*einstein-brown.ipynb*). A tutorial notebook (*transport-limited-LEM-example.ipynb*) shows how to use the component.

The time rate of change of elevation is calculated over a grid cell at node $i$ with surface area $\Lambda_i$, such that mass continuity gives the time rate of elevation ($z_i$) change (in the absence of other processes) as:

$\frac{dz_i}{dt} = \frac{Q_{in} - Q_{out}}{\Lambda_i}$

The sediment influx at each node is calculated as the sum of outflux of all upstream neighbor nodes that flow to it.

Note that this is a no-frills component, at least as of this initial version. It uses only drainage area (not discharge, even if calculated). It uses a very basic forward-Euler solution algorithm, and so it requires very small time steps. There is no adaptive solver. It is designed mainly for illustrative purposes: to show how a basic transport-limited area-slope LEM behaves. As with all Landlab components, the user should understand the theory and be familiar with the background literature!

*This Landlab component is dedicated to the memory of two remarkable hydrologists: Garry Willgoose and Ignacio Rodriguez-Iturbe.*

### Checklist

- [X] All tests have passed?
- [X] Formatted code with black?
- [X] Removed lint reported by flake8?

### DIFF
--- a/landlab/components/__init__.py
+++ b/landlab/components/__init__.py
@@ -1,3 +1,4 @@
+from .area_slope_transporter import AreaSlopeTransporter
 from .bedrock_landslider import BedrockLandslider
 from .carbonate import CarbonateProducer
 from .chi_index import ChiFinder
@@ -69,6 +70,7 @@ from .vegetation_dynamics import Vegetation
 from .weathering import ExponentialWeatherer, ExponentialWeathererIntegrated
 
 COMPONENTS = [
+    AreaSlopeTransporter,
     BedrockLandslider,
     CarbonateProducer,
     ChannelProfiler,
--- a/landlab/components/area_slope_transporter/__init__.py
+++ b/landlab/components/area_slope_transporter/__init__.py
@@ -1,0 +1,3 @@
+from .area_slope_transporter import AreaSlopeTransporter
+
+__all__ = ["AreaSlopeTransporter"]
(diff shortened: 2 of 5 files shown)


gregtucker:
@BCampforts @mcflugen could you have a look at this please?

mcflugen:
Sure. No problem.

mcflugen (approved):
👍

PR #1523, opened 2023-01-17
Remove richdem as a requirement

mcflugen:
### Description

This pull request removes *richdem* as a requirement for installing *landlab*. Previously, it was only required for certain platforms and Python versions for which it was available but now we leave it up to the user to install it separately if they need it (i.e. if they want to use the `PriorityFloodFlowRouter`).

### Checklist - did you ...

- [x] All tests have passed?


mcflugen:
@BCampforts Could you please have a look at this?

mdpiper:
TIL about `partial()` and *rich*.

mdpiper (approved):

landlab/components/priority_flood_flow_router/priority_flood_flow_router.py
```diff
@@ -375,6 +353,13 @@ def __init__(
         keyword arguments, tests the argument of runoff, and
         initializes new fields.
         """
+        if (richdem := self.load_richdem()) is None:
+            raise RuntimeError(
+                "PriorityFloodFlowRouter requires richdem but richdem is not installed"
+            )
```
BCampforts: Maybe we can help the user here by pointing out how to install? Not sure if we would recommend the conda version if this is not maintained though...

BCampforts (commented):
Looking good. One short comment on the error message when using PriorirtyFloodFlowRouter without RichDEM

landlab/components/priority_flood_flow_router/priority_flood_flow_router.py
mcflugen: It looks like you're looking at an old version. I've added a better error message but I guess it's not showing up. Maybe I forgot to push it to GitHub. 🤦

mcflugen: @BCampforts I'm not sure what's going on exactly but I think you must have commented on an old commit. This is code is not present in the latest version.

landlab/components/priority_flood_flow_router/priority_flood_flow_router.py
```diff
@@ -536,6 +517,20 @@ def __init__(
...
+        try:
+            import _richdem  # noqa: F401
+            import richdem
+        except ModuleNotFoundError as exc:
+            raise ModuleNotFoundError(
+                "PriorityFloodFlowRouter requires richdem but richdem is not installed."
+                "You can install richdem either from source, "
+                "(https://github.com/r-barnes/richdem), or through conda "
+                "(conda install richdem -c conda-forge) or pip (pip install richdem)."
+            ) from exc
```
mcflugen, line 531: @BCampforts Here is the error message that a user would see if they try to run `PriorityFloodFlowRouter` with *richdem*.

BCampforts: That looks better indeed!

BCampforts (approved):

Issue #1573, opened 2023-03-23
Link orientation


gregtucker:
`RasterModelGrid` provides arrays with the IDs of "horizontal" (`x`-directed) and vertical (`y`-directed) links. It would be useful if `HexModelGrid` could provide a similar capability. Because hex grids themselves have two options for grid orientation (one in which columns are staggered, and one in which rows are staggered), there are effectively six possible orientations. Aspects of such a functionality:

- Only create new arrays when needed (to save memory)
- Provide either a bool array or an array of IDs for any caller-specified orientation
- Handle any of the 6 possible orientations, which could be referred to based on the direction in which the link points: e, ene, nne, n, nnw, ese

gregtucker:
Resolved by #1574

Issue #1571, opened 2023-03-22
Numerical support for Total Variation Diminishing (TVD) advection scheme


gregtucker:
It would be useful to have a component that implements a TVD numerical scheme for advection, and various support functions for such a scheme.

gregtucker:
Addressed by #1582 and #1632

PR #1574, opened 2023-03-23
Link orientation: add link_has_orientation method to hex grid

gregtucker:
### Description

Add method to query orientation of grid links in `HexModelGrid`. Addresses #1573 and partly #1571. 

### Checklist - did you ...

- [x] All tests have passed?

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -46,7 +46,7 @@ executing==1.2.0
     # via stack-data
 fastjsonschema==2.16.2
     # via nbformat
-furo==2022.12.7
+furo==2023.3.27
     # via -r docs/requirements.in
 idna==3.4
     # via requests
@@ -54,7 +54,7 @@ imagesize==1.4.1
     # via sphinx
 incremental==22.10.0
     # via towncrier
-ipython==8.11.0
+ipython==8.12.0
     # via -r docs/requirements.in
 jedi==0.18.2
     # via ipython
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -132,6 +132,9 @@ pygments_dark_style = "monokai"
 autoclass_content = "both"
 
 
+# The default highlight language
+highlight_language = "none"
+
 # -- Options for HTML output ---------------------------------------------------
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
--- a/landlab/field/graph_field.py
+++ b/landlab/field/graph_field.py
@@ -956,8 +956,8 @@ class GraphFields:
         See Also
         --------
         numpy.empty : See for a description of optional keywords.
-        GraphFields.ones : Equivalent method that initializes the data to 1.
-        GraphFields.zeros : Equivalent method that initializes the data to 0.
+        :meth:`~.ones` : Equivalent method that initializes the data to 1.
+        :meth:`~.zeros` : Equivalent method that initializes the data to 0.
 
         Examples
         --------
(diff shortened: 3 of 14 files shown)


mcflugen:
@gregtucker Instead of `link_has_orientation` what do you think of a `HexModelGrid` (and also a `RasterModelGrid`, I suppose) having an attribute called `orientation_of_link` (the naming follows the pattern of other quantities like `length_of_link`, `angle_of_link`, etc.) that is an array of `unit8` values (much like `status_at_link`) that indicates the orientation?

The orientation codes could be defined much like `LinkStatus`. So, perhaps,
```python
from enum import IntEnum, unique

@unique
class LinkOrientation(IntEnum):
    """Define the link orientations"""

    E = 0
    ENE = 1
    NNE = 2
    N = 3
    NNW = 4
    ESE = 5
```

If you then wanted to get the links that have some orientation, you could do,
```python
grid.orientation_of_link == LinkOrientation.ENE
(grid.orientation_of_link == LinkOrientation.ENE) | (orientation_of_link == LinkOrientation.E)
```

Another option,
```python
from enum import IntFlag

class LinkOrientation(IntFlag):
    """Define the link orientations"""

    E = 1
    ENE = 2
    NNE = 4
    N = 8
    NNW = 16
    ESE = 32
```
You could then select links based on their orientation with things like the following,
```python
grid.orientation_of_link == LinkOrientation.ENE
grid.orientation_of_link & (LinkOrientation.ENE | LinkOrientation.E)  # not an array of bool
(grid.orientation_of_link & (LinkOrientation.ENE | LinkOrientation.E)).astype(bool)  # array of bool
```

gregtucker:
Ah, see now this is why we do code reviews ... good idea @mcflugen. The only downside I can see is the perpetual use of memory once `orientation_of_link` is created (as opposed to temporary creation of one or more bool arrays). But I suppose if that were a concern for a memory-scarce application one could always de-allocate it again with `orientation_of_link = 0`. Stay tuned...

mcflugen:
@gregtucker I made a few small changes.

* Fixed the documentation build error and, in doing so, also ended up fixing some cross-references in the documentation unrelated to this pull request.
* Moved `LinkOrientation` into its own module. This is what actually fixed the documentation build error (I don't know why) but it seemed to make sense anyway because other model grids may want to use this class in the future.
* Added more tests for the `orientation_of_link` method.
* Removed some repeated code by modifying `orientation_of_link` to use the existing `angle_of_link` method.
* Added the `cached_property` decorator to `orientation_of_link`.

gregtucker:
Thanks @mcflugen, this all looks good to me. I didn't know about the `@cached_property` decorator, how does it work?

I think the docstring description of `active_link_dirs_at_node` may be misleading, but would like your opinion before I try to modify it. The link directions are not actually the flux directions per se; rather, it's the combination of the link direction and the sign of a flux or velocity variable that determine direction: in the link direction when sign is positive, and the anti-link direction when sign is negative. This sometimes confuses Landlab users. If you agree with this distinction, perhaps we can update the description to something like "indicates whether a link points toward (+1) or away from (-1) the node".

mcflugen:
> I didn't know about the @cached_property decorator, how does it work?

It's a little mysterious but, basically, it turns the method into a property (like `@property` does) but only runs the body of the method the first time the method's called and then caches the result inside of the class and just returns that on subsequent calls.

One thing that is different from `@property`, which I just realized, is that the cached result is stored as an attribute with the same name as the method but that attribute is just a regular attribute and is not read-only.

That is, someone could do something like,
```python
>>> grid = HexModelGrid((3, 3))
>>> grid.orientation_of_link
array([ 1,  1, 16,  4, 16,  4, 16,  4,  1,  1,  1,  4, 16,  4, 16,  4, 16,
        1,  1], dtype=uint8)
>>> grid.orientation_of_link = "frog"
>>> grid.orientation_of_link
'frog'
```

> I think the docstring description of active_link_dirs_at_node may be misleading

I agree. The docstring should be updated to better describe what is being returned.

mcflugen:
@gregtucker I've updated the docstring for `active_link_dirs_at_node` to, I think, better reflect what it does.

I'm going to go ahead and merge this pull request as it's holding up your other one (#1577). If you don't like what I've done with the documentation, we can fix it over there.

PR #1582, opened 2023-04-03
AdvectionSolverTVD: component to model advection using 2nd-order nonlinear TVD scheme

gregtucker:
### Description

This PR adds a new component: `AdvectionSolverTVD`. The component models advection of a user-provided at-node input field at a velocity given by a user-provide at-link input field. The component uses a second-order nonlinear Total Variation Diminishing (TVD) numerical method.

### Checklist - did you ...

- [x] All tests have passed?


gregtucker:
One of the tricks with this component is that it applies advection to a user-specified field, rather than requiring a particular input field. But at least one of the automated tests seems to require that there be no required parameters other than `grid`; i.e., there must be default values for each parameter. The solution so far is to create a generic field if the user doesn't supply one.

mcflugen:
@gregtucker I refactored `find_upwind_link_at_link` to remove some `if` statements and make it a little simpler. More importantly, though, I changed its behavior to match that of `parallel_links_at_link` and not treat inactive links any differently that active links. Now it just finds the upwind link of every link and it's left up to the user to set the value of inactive links (which I've done in the `AdvectionSolverTVD` methods). I hope that's ok.

mcflugen (dismissed):
👍 @gregtucker, if you're ok with the changes I made to `find_upwind_link_at_link` and `upwind_to_local_grad_ratio` (and their tests), I think this looks good to me.

gregtucker:
@mcflugen those changes look good, thanks for injecting the advanced numpy wizardry! I think we're ready to merge. @BCampforts check this out, you inspired it...

PR #1627, opened 2023-05-31
Refactor overland flow component

mcflugen:
### Description

This pull request is an attempt to refactor the `OverlandFlow` and to try to speed it up.

### Checklist - did you ...

- [x] All tests have passed?


mcflugen:
@gregtucker, @yuvalsmilo I believe this is ready for review now.

All the tests are passing but, @yuvalsmilo, if you could please try these changes out on your test case, that would be great. I'm seeing some pretty large speed improvements on you example but it would be good if you could look at it to make sure that things are still working as you expect.

mcflugen:
There is one remaining *doctest* that's failing. I believe the discrepancy is the result of a ***bug fix***, not a new bug.

In the old version of *OverlandFlow*, have a look at the code block that adjusts discharge for stability and for supercritical flow. This is the code block contained within the large ``if self._steep_slopes`` block. Two checks are made: one for supercritical flow and one for stability. Discharge is then adjusted so that it is everywhere below the threshold for supercritical flow and then adjusted to something a little less than that needed for stability. But what if that threshold for stability is larger than that for supercritical flow? In such a case the discharge would be reset to be supercritical.

In my modified version, the discharge is first adjusted to be below supercritical and then the check for stability is performed (on the adjusted discharges) and modified if necessary.

In the failing test, the discharge required for stability is greater than that for supercritical flow.

mcflugen:
@gantian127 If you have time, would you also be able to test this version of *OverlandFlow* on your use cases?

gantian127:
@mcflugen I will test this version of OverlandFlow on my use case this week. Thanks!

gregtucker (dismissed):
Wow, congrats on a major overhaul! I just skimmed but based on that it looks good.

yuvalsmilo:
> @gregtucker, @yuvalsmilo I believe this is ready for review now.
> 
> All the tests are passing but, @yuvalsmilo, if you could please try these changes out on your test case, that would be great. I'm seeing some pretty large speed improvements on you example but it would be good if you could look at it to make sure that things are still working as you expect.

@mcflugen, I tested it in all of the configurations that I use and it works great. Thanks a lot!

mcflugen:
This seems to be an issue where small time steps cause an instability in discharge. I suspect this may be the result of how parallel links are averaged in the *q*-centered numerical scheme (eq 39 of the *de Almeida* paper). Since for small time steps the new *q* (eq 41) is dominated by this averaging term, it makes sense that this could be a problem. In addition, if I run a simulation with *theta*=1.0 (i.e. no averaging of parallel links) the discharge is nice and smooth, without spikes.

My thought is that this averaged *q* is not being calculated correctly on boundary nodes. With both the older version and my this newer version, if one of a link's parallel links is either inactive or nonexistent, the *q* of that parallel link is assumed to be zero. I'm not sure what the correct way to handle these boundaries is but if I set *theta* to 1 on just these links (i.e. don't consider the parallel links of links along the boundary), I get a smoothly varying discharge curve. Another option would be to set `q_{i-3/2} == q_{i+1/2}`.

I believe this is the same issue reported by @sophie-eisenring in #1966.

gantian127:
@mcflugen I tested this pull request with two use cases. Below are a brief summary of the results. I can share the result files and the testing code if you would like to look at them. Thanks!

**Test 1: paper use case**
The results for the original and refactor code are quite similar when theta is set as 1 or 0.8. 
**original results**
model_hx_1
model_hx_0 8

**refactor code results**
model_hx_1
model_hx_0 8
**Test 2: landscape use case**
The peak discharge value is smaller for the refactor code (original_code = 44.06 cms, refactor_code=40.46 cms). The timing of the peak discharge is earlier for the refactor code. The max surface water depth for the whole model simulation in the study site is smaller for the refactor code (original_code = 1.6m, refactor_code=1.2m). The recession limb in the time series is extended longer for the refactor code. 
**original results**

flow_12000

**refactor code results**
flow_12000

gregtucker:
Neat results @gantian127. The simple test case looks really similar; is it possible to tell which is closest to the expected solution from Bates, deAlmeida et al.? For the catchment test, that's a pretty big difference. Any insight into why they are so different?

PR #1632, opened 2023-06-06
Gt/advection multi field

gregtucker:
### Description

Add the ability to advect more than one field. Some applications need this. The motivating use case in this instance is the case where AdvectionSolverTVD is used to advect the hangingwall of an extensional fault. But the hangingwall has two materials: rock and sediment. Each needs to be advected. Using two instances of the component would be inefficient. The solution proposed here is to allow the user to submit a list of fields instead of a single field. Under the hood, the component keeps a list of fields (arrays) that get advected, and sweeps through all of them when update() or its synonym run_one_step() is called. If there's only one field, that's fine too of course - in that case, it is just a list of one.

### Checklist

- [x] All tests have passed?


gregtucker:
Hi @BCampforts @ethan-pierce would either of you have time for a quick review of this, given that you're both fans of TVD advection? And/or can you think of someone else in the community who could review? (I'm trying not to overburden @mcflugen ...)

Issue #1638, opened 2023-06-13
SpaceLargeScaleEroder overwrites "sediment__influx" field with zeros


loroberge:
The extension `calc_sequential_ero_depo.pyx` overwrites the _"sediment__influx"_ value with a zero as it iterates downstream through each node. This happens on line 96. The result is a _"sediment__influx"_ field containing only zeros except at the outlet (the last node in the iteration).

`SpaceLargeScaleEroder` already zeros the entire field prior to calling `_sequential_ero_depo` (line 528), so zeroing during the downstream iteration is redundant.

Removing line 96 of `calc_sequential_ero_depo.pyx` solves the problem and results in a correctly populated _"sediment__influx"_ field. This doesn't seem to break any tests. 

@BCampforts @gregtucker 

Example:
```
import numpy as np
from landlab import RasterModelGrid
from landlab.components import PriorityFloodFlowRouter, SpaceLargeScaleEroder

nrows = 3
ncols = 5
outlet_node = ncols

mg = RasterModelGrid((nrows, ncols))
mg.status_at_node[mg.boundary_nodes] = mg.BC_NODE_IS_CLOSED
mg.status_at_node[outlet_node] = mg.BC_NODE_IS_FIXED_VALUE

_ = mg.add_ones('soil__depth', at='node')
_ = mg.add_zeros('bedrock__elevation', at='node')
_ = mg.add_zeros('topographic__elevation', at='node')
mg.at_node['topographic__elevation'][:] = (mg.at_node["bedrock__elevation"]
                                           + mg.at_node["soil__depth"])
core_ids = np.append(outlet_node, mg.core_nodes)

fr = PriorityFloodFlowRouter(mg)
fr.run_one_step()

sp = SpaceLargeScaleEroder(mg)

for i in range(100):
    mg.at_node['bedrock__elevation'][mg.core_nodes] += 0.01
    mg.at_node['topographic__elevation'][:] = (mg.at_node["bedrock__elevation"]
                                               + mg.at_node["soil__depth"])
    fr.run_one_step()
    sp.run_one_step(dt=10)
```
Sediment outflux field shows sediment is moving out of each node.
```
mg.at_node['sediment__outflux'][core_ids]
```

Sediment influx should be same as outflux but shifted downstream by one node. Instead it is all zeros until the outlet.
```
mg.at_node['sediment__influx'][core_ids]
```

PR #1640, opened 2023-06-13
Keep sediment__influx values in SpaceLargeScaleEroder

gregtucker:
### Description

The `SpaceLargeScaleEroder` component has a `sediment__influx` field that tracks the volume rate of sediment entering each grid node. However, the values on core nodes are zeroed out in the main loop. This PR provides a potential solution.

As far as I can tell, the zeroing out is done to facilitate calculating the volume of sediment leaving the river(s) in each time step, where "leaving" means entering an active boundary node. By zeroing out every sediment-influx value BUT those on boundary nodes, the return of this summed calculation can be done by just summing up all the sediment influx values. This PR proposes an alternative approach: only sum sediment influx values on boundary nodes. Values on closed boundary nodes should always be zero, so in effect this amounts to adding up the values on open boundary (i.e., fixed value) boundary nodes. Doing it this way means it is no longer necessary to zero out the sediment influx value at each core node as it is processed, so the influx field is preserved for other components to use, or for further analysis and/or plotting.

This addresses issue #1638.

### Checklist

- [X] Add new / update outdated documentation?
- [x] All tests have passed?


gregtucker:
The test fail seems unrelated to this PR. Could @BCampforts take a look when you get a chance? @loroberge needs this for his "concentration tracking" component.

BCampforts:
> The test fail seems unrelated to this PR. Could @BCampforts take a look when you get a chance? @loroberge needs this for his "concentration tracking" component.

It's on my radar @gregtucker, will have a look sometime this week.

mcflugen:
@BCampforts Do you have any thoughts on this pull request? If it's good to go, I would like to include it with the forthcoming 2.7 release (tomorrow?).

mcflugen (approved):
👍

PR #1649, opened 2023-06-23
SpaceLargeScaleEroder: Add getter for several input parameters.

loroberge:
### Description

This PR addresses Issue #1652 . It adds getters for each of the following input parameters to _SpaceLargeScaleEroder_: `F_f`, `phi`, `v_s`, `m_sp`, `n_sp`. This is done so they can be easily obtained by other components or code that rely on the space instance. 

For example, I am working on a sediment property tracking component that must be run after _SpaceLargeScaleEroder_ and uses `F_f`, `phi`, and `v_s`. 

### Checklist - did you ...

- [x] All tests have passed?


mcflugen (changes requested):
Looks good. Thanks @loroberge! My only comment is that it would be nice if the public versions of these attributes had more descriptive names (and were all lowercase).

landlab/components/space/space_large_scale_eroder.py
```diff
@@ -420,6 +420,31 @@ def K_sed(self):
     def K_sed(self, new_val):
         self._K_sed = return_array_at_node(self._grid, new_val)
 
+    @property
+    def F_f(self):
```
mcflugen, line 424: `suspendable_fraction` instead of `F_f`?

landlab/components/space/space_large_scale_eroder.py
```diff
@@ -420,6 +420,31 @@ def K_sed(self):
...
+        """Fraction of permanently suspendable fines in bedrock [-]."""
+        return self._F_f
+
+    @property
+    def phi(self):
+        """Sediment porosity [-]."""
+        return self._phi
+
+    @property
+    def v_s(self):
```
mcflugen, line 434: `settling_velocity` instead of `v_s`?

landlab/components/space/space_large_scale_eroder.py
```diff
@@ -420,6 +420,31 @@ def K_sed(self):
...
+        """Effective settling velocity for chosen grain size metric [L/T]."""
+        return self._v_s
+
+    @property
+    def m_sp(self):
```
mcflugen, line 439: `drainage_area_exp`?

landlab/components/space/space_large_scale_eroder.py
```diff
@@ -420,6 +420,31 @@ def K_sed(self):
...
+        """Drainage area exponent (units vary)."""
+        return self._m_sp
+
+    @property
+    def n_sp(self):
```
mcflugen, line 444: `slope_exp`?

mcflugen (changes requested):
I added some (maybe not so great) suggestions for names but please feel free to use other names if you come up with something you think is better.

Actually, I take that back. If we're not changing the input keywords, then maybe it's most clear if these names match the names of those keywords. I think dropping the `_sp` suffix on some of them might still be a good idea (assuming that `sp` stands for `SPACE`).

tests/components/space/test_space_large_scale_eroder.py
```diff
@@ -206,6 +206,36 @@ def test_soil_field_already_on_grid():
...
         verbose=True,
     )
+    testing.assert_array_equal(
+        0.0,
+        sp.F_f,
+        err_msg="Parameter value issue",
+        verbose=True,
+    )
+    testing.assert_array_equal(
+        0.3,
```
mcflugen, line 216: You set *phi* to 0.0, not 0.3, when you create `sp`.

mcflugen (approved):

PR #1650, opened 2023-06-23
BedrockLandslider: Add getters for several input parameters.

loroberge:
### Description
This PR addresses Issue #1651 . It adds getters for the following input parameters to _BedrockLandslider_: `fraction_fines_LS`, `phi`. This is done so they can be easily obtained by other components or code that rely on the space instance.

For example, I am working on a sediment property tracking component that must be run after _BedrockLandslider_ and uses `fraction_fines_LS` and `phi`.

### Checklist - did you ...

- [x] All tests have passed?


landlab/components/bedrock_landslider/bedrock_landslider.py
```diff
@@ -342,6 +342,21 @@ def __init__(
             np.random.seed(seed)
 
     # Getters for properties
+    @property
+    def fraction_fines_LS(self):
```
mcflugen, line 346: @loroberge What do you think about renaming this property to just `fraction_fines`? Names should be all lowercase and I'm not sure what *LS* stands for. If it stands for landslider, then that's probably redundant since it's an attribute of the landslider component.

news/1651.misc
```diff
@@ -0,0 +1 @@
+Added getters for several BedrockLandslider input parameters.
```
mcflugen, line 1: Just for fun, instead of `BedrockLandslider`, write `` :class:`~.BedrockLandslider` `` so that it will link to its documentation page.

mcflugen (changes requested):
👍

news/1651.misc
loroberge: Great tip! Done.

landlab/components/bedrock_landslider/bedrock_landslider.py
loroberge: That sounds good. I believe this name was to differentiate it from a similar parameter used in `Space` and `SpaceLargeScaleEroder` (it is called `F_f` in those). You're right that it seems redundant.
@BCampforts do you have a preference for the property name?

loroberge: I'm now looking at PR #1649  with similar requests and I may have misunderstood. Do you want this only done for the public property? Here I've changed all instances of `fraction_fines_LS`....

mcflugen: I was just thinking of the public methods.

loroberge: ok. I've fixed it to only change the public method.

mcflugen (approved):

PR #1657, opened 2023-07-06
Add RiverBedDynamics Component

angelmons:
This commit introduces a new component, RiverBedDynamics, designed to calculate sediment transport and river bed evolution.

We also added the possibility to chose a time step in Overland Flow in the deAlmeida model

### Description

A new component, River Bed Dynamics, models the evolution of a gravel bed river. This component can utilize various bedload models, and it is capable of tracking changes in the substrate over time
### Checklist - did you ...

- [X] All tests have passed?
- [X] Formatted code with black?
- [X] Removed lint reported by flake8?
- [X] Sucessful documentation built? (if documentation added or modified)

### DIFF
--- a/.github/workflows/test-notebooks.yml
+++ b/.github/workflows/test-notebooks.yml
@@ -50,4 +50,4 @@ jobs:
           OPENTOPOGRAPHY_API_KEY: ${{ secrets.OPENTOPOGRAPHY_API_KEY }}
           MPLBACKEND: "Agg"
         run: |
-          nox --verbose -s test-notebooks -- -m "${{ matrix.markers }}"
+          nox --verbose -s test-notebooks --force-pythons="${{ matrix.python-version }}" -- -m "${{ matrix.markers }}"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,7 +62,7 @@ jobs:
         env:
           HYPOTHESIS_PROFILE: "ci"
         run: |
-          nox -s test -- -m "${{ matrix.pytest-marker }}"
+          nox -s test --force-pythons="${{ matrix.python-version }}" -- -m "${{ matrix.pytest-marker }}"
 
       - name: Coveralls
         if: matrix.os == 'ubuntu-latest'
--- a/news/1754.bugfix
+++ b/news/1754.bugfix
@@ -1,0 +1,3 @@
+
+Fixed a bug that caused an incorrect Python version to be used in *Landlab*'s
+continuous integration tests.
--- a/noxfile.py
+++ b/noxfile.py
@@ -7,9 +7,10 @@ import nox
 
 PROJECT = "landlab"
 ROOT = pathlib.Path(__file__).parent
+PYTHON_VERSION = "3.11"
 
 
-@nox.session(venv_backend="mamba")
+@nox.session(python=PYTHON_VERSION, venv_backend="mamba")
 def test(session: nox.Session) -> None:
     """Run the tests."""
     os.environ["WITH_OPENMP"] = "1"
@@ -38,7 +39,7 @@ def test(session: nox.Session) -> None:
         session.run("coverage", "report", "--ignore-errors", "--show-missing")
 
 
-@nox.session(name="test-notebooks", venv_backend="mamba")
+@nox.session(name="test-notebooks", python=PYTHON_VERSION, venv_backend="mamba")
 def test_notebooks(session: nox.Session) -> None:
     """Run the notebooks."""
     args = [
@@ -123,7 +124,6 @@ def build_index(session: nox.Session) -> None:
     session.log(f"generated index at {index_file!s}")
 
 
-# @nox.session(name="build-docs", venv_backend="mamba")
 @nox.session(name="build-docs")
 def build_docs(session: nox.Session) -> None:
     """Build the docs."""
@@ -168,7 +168,7 @@ def locks(session: nox.Session) -> None:
     # session.run("conda-lock", "lock", "--mamba", "--kind=lock")
 
 
-@nox.session(name="sync-requirements", python="3.11", venv_backend="conda")
+@nox.session(name="sync-requirements", python=PYTHON_VERSION, venv_backend="conda")
 def sync_requirements(session: nox.Session) -> None:
     """Sync requirements.in with pyproject.toml."""
     with open("requirements.in", "w") as fp:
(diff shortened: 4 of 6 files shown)


angelmons:
Hello @gregtucker , @mcflugen , and @nicgaspar 

I just submitted a new pull request. This one is a cleaner version, and it passed all the tests on my computer. In terms of the River Bed Dynamics algorithm, nothing has changed. Also, I followed the advice of @mcflugen and removed almost all the previous changes made to OverlandFlow. Now, only a user-defined time step option has been included, which is required by River Bed Dynamics.

gregtucker:
Looks great! I went ahead and launched the CI tests.

angelmons:
Hello @gregtucker 

I have been checking the test results and it has been saying:
"Check notebooks Expected — Waiting for status to be reported"
"Run the tests Expected — Waiting for status to be reported"

It has been like this for 2 days, would I do something about it?

Angel

mcflugen:
@angelmons I'm not exactly sure why those checks are showing up here but they will never start so no need to wait for them.

mcflugen:
> Now, only a user-defined time step option has been included, which is required by River Bed Dynamics.

I believe the ``dt`` keyword of the ``overland_flow`` method was intended to provide this functionality.

angelmons:
Hi @mcflugen ,
Thanks for the replies.
i) If the test will never start, what should I do? is that a problem? Both test, notebook and tests, were running without problems in my computer.
ii) I tried to use dt in overland flow but it was overwritten by the one calculated by calc_time_step. Specifically, I would set it as dt = 5 [s], then run_one_step in overlandFlow and got back dt = 16.1 [s]. The option of a user-defined fixed dt solved that problem.

mcflugen:
> If the test will never start, what should I do? is that a problem?

Ignore it. I don't think it's a problem since the other tests are running just fine.

> Both test, notebook and tests, were running without problems in my computer.

This is surprising since the tests are failing for me as well. From the looks of it, the ``gsd`` keyword is an integer by default (i.e. ``0``) but later in the code you use it as though it is an array (e.g. ``self._gsd[:, 0]``).

> I tried to use dt in overland flow but it was overwritten by the one calculated by calc_time_step. Specifically, I would set it as dt = 5 [s], then run_one_step in overlandFlow and got back dt = 16.1 [s]. The option of a user-defined fixed dt solved that problem.

In that case, that sounds like a bug to me. If possible, I would like this functionality to be provided through the ``dt`` keyword to the ``overland_flow`` method rather than as a keyword to ``__init__``.

It looks to me like what may be happening is that if you run the ``overland_flow`` method and provide a ``dt`` (i.e. ``of.overland_flow(dt=5.0)``), it runs for the given duration but the ``dt`` attribute is set to the calculated time step. Could that be the case?

angelmons:
> This is surprising since the tests are failing for me as well. From the looks of it, the `gsd` keyword is an integer by default (i.e. `0`) but later in the code you use it as though it is an array (e.g. `self._gsd[:, 0]`).

I had the opposite error before, it was saying that gsd can't be an array and I solved it by setting gsd to 0. That is strange. 


> > I tried to use dt in overland flow but it was overwritten by the one calculated by calc_time_step. Specifically, I would set it as dt = 5 [s], then run_one_step in overlandFlow and got back dt = 16.1 [s]. The option of a user-defined fixed dt solved that problem.
> 
> In that case, that sounds like a bug to me. If possible, I would like this functionality to be provided through the `dt` keyword to the `overland_flow` method rather than as a keyword to `__init__`.
> 
> It looks to me like what may be happening is that if you run the `overland_flow` method and provide a `dt` (i.e. `of.overland_flow(dt=5.0)`), it runs for the given duration but the `dt` attribute is set to the calculated time step. Could that be the case?

Yes, that is what I think, I will check again and let you know. Thanks for your help.

Angel

angelmons:
Hi @gregtucker, @mcflugen, and @nicgaspar,

Thank you for your previous comments. I've addressed all the issues you pointed out. @mcflugen, your suggestion was particularly helpful – setting the 'dt' during the call to OverlandFlow, rather than during instantiation, indeed sets the desired 'dt'. As a result, I no longer need to make any modifications to OverlandFlow, which is fantastic news. 

In addition, I've corrected a few errors in the tests that emerged when I ran them on my Linux machine. These were related to the gsd variable that @mcflugen pointed out. 

Tomorrow I'm going to close this pull request and open a new, clean one.

Once again, thank you for your insightful comments.

angelmons:
Hello @gregtucker, @mcflugen , and @nicgaspar 

I updated all the files. Would it be possible to run the tests?

In my computer I got this:
nox > Session test was successful.

I hope that it gives the same answer here.

Thanks

Angel

angelmons:
Hello @gregtucker, @nicgaspar, and @mcflugen,

It seems that when I addressed the issue with the 'gsd' variable for one test, it caused a failure in another test. I believe I'll need to find a different way to define it so that it works for all the tests.

Most tests were successful this time, so I'm optimistic about resolving this issue soon.

Angel

angelmons:
Hello @gregtucker and @mcflugen 

Could you please authorize to run the workflow. I hope this time everything goes ok. Thanks

Angel

angelmons:
Hello @gregtucker, @mcflugen , and @nicgaspar 

I have been checking the status of these tests, and there have been no changes during the last few days. What are the next steps, or is there anything I need to fix from your end?

Thanks

mcflugen:
@angelmons I just updated your branch with the *master* branch are reran the tests. I'll start my review of the code today. If it's ok with you, I may push some minor changes to your branch while I go through the code.

angelmons:
Thanks @mcflugen, sounds good. If you need anything from my end, please let me know.

angelmons:
Hello @gregtucker , @mcflugen , and @nicgaspar 

I hope you're doing well. I wanted to touch base regarding the status of my recent commit. If there's anything I need to adjust or update, please let me know. I noticed there haven't been any updates on it in the past three weeks, and I'm keen to assist in any way I can to move things forward.

Thank you for your time and guidance!

Angel

mcflugen:
@angelmons So sorry for the delay. I'm still going through this but one question I have concerns the number of fields your component uses. The fields you attach to the grid should only be ones you think your component might share with other components. I could be wrong but, to me, it appears that some of these fields are really arrays that would only be used within your component. If that's the case, I would change those arrays to be private attributes of you component rather than attaching them to the grid as fields.

Issue #885, opened 2019-01-23
Voroni "boundary" does not include all boundary nodes


kbarnhart:
A similar but related comment to Issue #880.

I recently wanted to make a Voronoi grid, close some boundaries, and run Stream Power on it. Naively I did not ask @gregtucker or @nicgaspar about how this was done in CHILD. Instead I just generated random points and went on my merry way. 

I expected that if I set all the "boundary nodes" to CLOSED, then that might not close off all the boundaries... But I expected that if I set all the nodes adjacent to boundary nodes to CLOSED, then this should work. 

However, it doesn't because the nodes ID'd as boundary by Landlab are not correct (this is where this issue overlaps with #880). Fixing this is not critical... But probably a thing to do by LL2.0.

A MWE:

```python
import matplotlib.pylab as plt
import numpy as np

from landlab import VoronoiDelaunayGrid, imshow_grid, CORE_NODE, BAD_INDEX_VALUE
from landlab.components import (
    FastscapeEroder,
    FlowAccumulator,
)

np.random.seed(42)

dx = 100

nr = 20
nc = 20

x = dx*nr*np.random.rand(nr*nc)
y = dx*nc*np.random.rand(nr*nc)

grid = VoronoiDelaunayGrid(x=x, y=y)
    
z = grid.add_zeros("node", "topographic__elevation")
z[:] += np.random.rand(grid.size("node"))
    
boundary_neighbors = grid.node_has_boundary_neighbor(grid.nodes)

grid.set_closed_nodes(boundary_neighbors)
grid.set_closed_nodes(grid.boundary_nodes)
imshow_grid(grid, z, cmap='viridis', color_for_closed='red', show_elements=True)


plt.figure()
imshow_grid(grid, z, cmap='viridis', color_for_closed='red')
plt.plot(grid.x_of_node[grid.boundary_nodes], grid.y_of_node[grid.boundary_nodes], 'k.')
plt.show()
``` 
<img width="341" alt="screen shot 2019-01-22 at 7 48 53 pm" src="https://user-images.githubusercontent.com/6589339/51579591-c61cce80-1e7e-11e9-931e-61965e8ecea2.png">

My expectation would be that red CLOSED cells would extend all the way around the grid. 

Note that making test grids that are too small (e.g. ~10x10) results in getting a result that looks right, but does not generalize to larger grids.

PR #1661, opened 2023-07-13
Improve mesh generation for unstructured grids

ethan-pierce:
**Work in progress!**
This pull request is intended to address a few major problems with the mesh generation used by the VoronoiDelaunayGrid. These include:
* Boundary points are rarely preserved (see #1450 and #885)
* Grid elements (patches and cells) are often ill-conditioned for numerical methods. (Angles of some elements are too small, or the difference in area between elements is too large.)
* Given a concave polygon, elements often stretch across the border of the shape to fill the convex hull, even when such behavior is not desired.
* The underlying scipy functions often fail for realistic geometries (e.g., shapefiles of real-world catchments), and rarely tell the user anything useful about why they failed.
* There is no easy way to refine the mesh based on an a priori understanding of the problem to be solved.

Rather than reinventing the wheel, this pull request argues that we should use Jonathan Shewchuk's Triangle software to generate Delaunay triangulations. Triangle is a community standard in computational geometry; see its documentation [here](https://www.cs.cmu.edu/~quake/triangle.html) for more information.

Since a few other grid types currently depend on the VoronoiDelaunayGrid, I propose to add a new grid type so as to avoid breaking any functionality. We can call it TriangleMeshGrid or something similar. It should be able to take a set of user-defined points or a path to a shapefile and produce a well-conditioned Delaunay triangulation over that region of interest. Geometrically, it won't be any different from the VoronoiDelaunayGrid, as we'll continue to use the Delaunay graph for nodes, links, and patches and the Voronoi graph for corners, faces, and cells. The primary advantage is that we'll be able to consistently generate *conforming* and/or *constrained* Delaunay triangulations, which are better for most numerical applications.

Of some interest, perhaps, this is also a prerequisite step for finite element methods, semi-adaptive mesh refinement, and a couple of domain-specific use cases in glacial modeling.

To keep things manageable, this PR implements a wrapper around Triangle and translates the output into a graph structure that can later be used to create a Landlab grid.

Checklist:
- [x] Implement a Python wrapper around Triangle's CLI
- [x] Translate Triangle's output into a Graph structure
- [x] Write some basic tests
- [x] Improve performance
- [x] Format and clean lint


mcflugen:
@ethan-pierce I would like it if the ``__init__`` method of ``TiangleGraph`` more closely matched that of the ``DelaunayGraph``. Instead of,
```python
def __init__(self, delaunay: dict, voronoi: dict):
```
what do you think of,
```python
def __init__(self, node_y_and_x):
```
and whatever keywords you think would be helpful.

ethan-pierce:
@mcflugen Good suggestion, I can refactor to bring TriangleGraph closer to the existing standard. How should we deal with situations where users choose not to provide node_y_and_x, but construct the mesh from e.g., a shapefile instead? (I think that will be a common use case here.)

mcflugen:
> How should we deal with situations where users choose not to provide node_y_and_x, but construct the mesh from e.g., a shapefile instead?

Through classmethods, I think, in the same way as you've done with ``TriangleMesh``. That is,
```python
@classmethod
def from_shapefile(cls, path_to_shapefile):
    ...
```
Would that work?

ethan-pierce:
This code smells a bit funny to me, but it matches the API for other graph classes. Let me know if there's a better way to do this without (ab)using composition.

mcflugen:
I was hoping for something like,
```python
@classmethod
def from_shapefile(cls, path_to_shapefile, **kwds):
    x_of_node, y_of_node = read_shapefile(path_to_shapefile)
    return cls((x_of_node, y_of_node), **kwds)
```
Do you think something like that could work?

I'm not crazy about ``__init__`` taking a required argument and then not using it (i.e. if ``input_file`` is provided).

ethan-pierce:
@mcflugen Fixed the issue with the constructors, so the __init__ interface should now be consistent. And, you were correct to guess that there are performance issues with moderately sized meshes (> 10k nodes). Working on that problem next.

landlab/graph/triangle/triangle_graph.py
mcflugen, line 1: @ethan-pierce I think we can get rid of this module now, correct? ``DualTriGraph`` should be a replacement for ``TriangleGraph``.

tests/graph/triangle/test_triangle_graph.py
```diff
@@ -0,0 +1,161 @@
...
+    assert_array_equal(
+        graph.corners_at_cell[:3],
+        [
+            [70, 79, 2, 77, 71, 68, 62, -1, -1],
+            [23, 65, 63, 27, 22, 28, -1, -1, -1],
+            [88, 115, 136, 101, 138, 89, 95, -1, -1],
+        ],
+    )
+
+    # assert_array_equal(graph.n_corners_at_cell[:3], [9, 9, 9])
```
mcflugen, line 50: ``DualTriGraph`` doesn't have a ``n_corners_at_cell`` (I don't think any of the graphs do). Do we need it? It looks like what was there before wasn't working correctly, anyway (the tested cells should have 7, 6, and 7 corners, not all 9).

tests/graph/triangle/test_triangle_graph.py
```diff
@@ -0,0 +1,161 @@
...
+    assert graph.number_of_links == 232
+    assert graph.number_of_patches == 144
+    assert graph.number_of_corners == graph.number_of_patches
+    assert graph.number_of_faces == 200
+    assert graph.number_of_cells == 57
+
+    assert np.all(graph.area_of_cell > 0.0)
+    assert np.all(graph.area_of_patch > 0.0)
+    assert np.all(graph.length_of_link > 0.0)
+    assert np.all(graph.length_of_face > 0.0)
```
mcflugen, line 29: There are still some faces that have zero length. I believe this test is still failing.

mcflugen:
@ethan-pierce The error on Windows is a memory violation error—the mysterious 3221225477 return code. This could be the result of, for example, dereferencing an uninitialized pointer, or accessing an array outside of its bounds.

The lack of output may indicate that the error is happening while reading the input file. However, oftentimes output is buffered and that buffer may not be flushed before the program crashes.

To get to the bottom of this I think we'll need to build *Triangle* on a Windows machine and see if we can reproduce the error. Maybe it can be solved with a compiler flag but, otherwise, we may need to create a patch.

PR #1679, opened 2023-07-25
Linear diffuser updates

isamarcortes:
### Description

This PR contains a work in progress code for a smaller version of the linear diffuser component. Will continue to work on this and updating regularly!


gregtucker:
Good idea!

Consider naming it something like "DiffusionSolverFiniteDifference" to match the generic advection component "AdvectionSolverTVD".

Issue #1704, opened 2023-08-22
Landlab dev install PackagesNotFound


pfeiffea:
Hey team. I'm struggling doing a developer install (following instructions [here](https://landlab.readthedocs.io/en/master/install/index.html#)) on a new machine. My branch is up to date with landlab/master, conda is up to date. I set up a new landlab_dev environment. 

Issue 1 (resolved, but do instructions need an update?): 

`conda install --file=requirements.in` gives me the following error

```
PackagesNotFoundError: The following packages are not available from current channels:

  - bmipy
  - rich-click
 ```

Based on [1486](https://github.com/landlab/landlab/issues/1486), I tried running `conda install landlab -c conda-forge` first, and this solved the problem. _Should installing regular landlab be a step written in to the developer install documentation?_ 

Issue 2: 
`pip install -e .` gives me the following 

```
Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... error
  error: subprocess-exited-with-error

  × Getting requirements to build editable did not run successfully.
  │ exit code: 1
  ╰─> [94 lines of output]

      Error compiling Cython file:
      ------------------------------------------------------------
      ...
          function be evaluated with x=1 to determine if it any erosion occured.

          """
          cdef double f

          f = x - 1.0 + (alpha * (x ** n)) - beta
                                           ^
      ------------------------------------------------------------

      landlab\components\stream_power\cfuncs.pyx:360:37: Cannot assign type 'npy_double complex' to 'double'
```
...
```
      Cython.Compiler.Errors.CompileError: landlab\components\stream_power\cfuncs.pyx
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build editable did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
```

I remember that I've had cython/compiler errors before related to Windows machines and C++ compilers, but now I can't find those instructions (referenced in [1513](https://github.com/landlab/landlab/issues/1513), but link is broken) on the updated documentation pages. 

Any suggestions?

mcflugen:
@pfeiffea I suspect you just need to add `-c conda-forge` to the install command.

pfeiffea:
Hey @mcflugen - perfect timing, I was just updating my issue. Yes, `-c conda-forge` was definitely the solution to issue #1.

[...]

PR #1706, opened 2023-08-23
Update install commands to include channels

mcflugen:
### Description

This pull request updates the *conda*/*mamba* installation commands in the docs to explicitly include channels. Since some of *Landlab*'s dependencies are not available in the default *conda* channel, it's important to tell *conda* to look to *conda-forge* for packages (see #1704). If someone doesn't have *conda-forge* in their *.condarc* file, packages from *conda-forge* won't be seen and users will see a ``PackagesNotFoundError``.

As there can be incompatibilities with packages from the default channel and from *conda-forge*, it's good practice to ensure that all packages come from *conda-forge*. Thus, the install command now looks like,
```
conda install --file=requirements.in -c nodefaults -c conda-forge --override-channels
```
This tells conda to look at *conda-forge*, ignore the default channel, and override any channel settings a user may have in their config file.

### Checklist - did you ...

- [x] All tests have passed?

### DIFF
--- a/docs/source/install/developer_install.rst
+++ b/docs/source/install/developer_install.rst
@@ -34,7 +34,14 @@ or download a `zip file <https://github.com/landlab/landlab/archive/refs/heads/m
    curl -OL https://github.com/landlab/landlab/archive/refs/heads/master.zip
 
 Once you have a copy of the source code, you can install it into your current
-Python environment,
+Python environment by first installing *Landlab* dependencies and then building
+and installing *Landlab*.
+
+
+Install dependencies
+````````````````````
+
+*Landlab*'s dependencies are listed in *requirements.in*.
 
 .. tab:: mamba
 
@@ -42,7 +49,6 @@ Python environment,
 
      cd landlab
      mamba install --file=requirements.in -c nodefaults -c conda-forge --override-channels
-     pip install -e .
 
 .. tab:: conda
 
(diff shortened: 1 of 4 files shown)


mcflugen:
@pfeiffea when you have time, would you please give this pull request a look over and see if it makes sense to you? And that it works!

pfeiffea (approved):
This looks great. Just ran successfully on my machine. 

Thanks @mcflugen

mcflugen:
@pfeiffea In #1700 you also had an issue with your C++ compiler. When building Python extensions (as we do when building *Landlab*), one must have a C++ compiler installed. This is true of any operating system but it seems to be most commonly an issue for Windows users (as you experienced). Since this is somewhat outside of a *Landlab* issue, and since it can be a moving target, we don't have anything about it on our docs. It would be nice, however, if we could point users to some online resources that describe how to install and set up their compilers. The best that I've found is buried in the [*conda-forge* docs](https://conda-forge.org/docs/maintainer/knowledge_base.html#notes-on-native-code) and on the [Python wiki page for Windows compilers](https://wiki.python.org/moin/WindowsCompilers). Would those have been helpful to you? Did you find anything online you found helpful that maybe we could point users to?

pfeiffea:
@mcflugen I ended up using the python wiki page you linked above, and the error that popped up in the command prompt sent me to the correct C++ compiler install page. That was totally sufficient. 

I think a simple mention in the install instructions that (especially Windows) users may need to install updated C++ compilers would be a simple, but substantial help. I've run into this every time I've set up a new computer (or re-set up a computer after wiping). I seem to remember there was mention of this step on the old docs page, so I wondered if it was no longer a necessary step when I didn't see it mentioned this time.

mcflugen:
@pfeiffea I've added text about compilers that also includes a couple links to resources for Windows users.

I've tried to balance saying enough to be helpful but not saying too much, which could make the process seem difficult, include information that could become dated, or repeat instructions that are already elsewhere on the internet.

@gantian127, @gregtucker Any thoughts?

gantian127:
@mcflugen I did a test for the source install of landlab on windows system. I mainly tested the conda commands and it worked. Only one comment is for the git clone command in the documentation. When using the link, it will show error info (fatal: unable to connect to github.com). It may be better to change the link as https://github.com/landlab/landlab.git.  Thanks!

mcflugen:
Thanks @gantian127!

I've changed the installation page to include commands to clone the repository through ssh (with the correct address!) and https.

Issue #1694, opened 2023-08-20
DepthDependentTaylorDiffuser should support hex grids


gregtucker:
`DepthDependentTaylorDiffuser` relies on the grid having an attribute `dx`, representing grid-cell spacing. Raster grids possess this attribute, but hex grids use `spacing` instead. The component should be smart enough to handle either case.

mcflugen:
@gregtucker A possible solution would be to use the `length_of_link` attribute, which is available on all grids. One issue with that, though, is that it is an array of length *n_links* so we would have to use, for example, something like `np.min(grid.length_of_link)`.

gregtucker:
@mcflugen that sounds right to me. I guess there's a choice to be made between calculating the minimum link length ahead of time, during `__init__`, or calculating it each time the Courant condition is applied. The former increases performance but could fail if one had a deforming or AMR ("active mesh refinement") grid; the latter is slower and allows deforming/AMR grids. I guess I lean toward the first, as I'm unaware of anyone using this or other LL components on deforming or AMR grids.

mcflugen:
@gregtucker I agree. This is something to keep in mind but, until *Landlab* allows grids to evolve after they have been created, not something we need to support yet.

gregtucker:
Addressed by PR #1707

PR #1707, opened 2023-08-23
Gt/use link length in taylor diffusers

gregtucker:
### Description

In the two 'Taylor Diffuser' components, the `RasterModelGrid` node spacing attribute `.dx` is used to estimate the upper limit on time-step size based on the Courant condition. But other grid types do not possess this attribute. This PR replaces the use of `.dx` with a calculation of the shortest link in the grid. This is calculated and stored during `__init__` so there remains potential for problems if a dynamic mesh were used. But at least this allows the components to work with other grid types, such as `HexModelGrid`.

Addresses #1694 

### Checklist

- [x] All tests have passed?


gregtucker:
The notebook fail is unrelated. Tests have otherwise passed. Promoting this to "ready" and pulling lastest master.

Issue #1637, opened 2023-06-12
Documentation needs some cleanup


gregtucker:
A recent/ongoing review of documentation points to several issues, listed below.

(note: this is a "live" issue, to be edited as the review progresses)

Larger issues:

* [ ] There are currently two different landing pages, one on landlab.github.io and one on landlab.readthedocs.io. The relationship between these pages should be resolved. Simplify landlab.github.io, or redirect to rtd.
* [x] The "precompiled install" link from the github.io landing page actually points to dev install.
* [x] ...and the dev install link points to a *different* set of dev install instructions.
* [ ] The API Reference link from github.io lands on the main RTD page, which doesn't identify itself as API Reference.
* [ ] The "Support" link from github.io lands on main RTD page, nothing obvious to do with support.
* [ ] Same with "Cite".
* [ ] We should have a functional, auto-generated API Reference section.
* [x] We should remove old extraneous stuff, like things having to do with Py 2.
* [ ] Tutorials section should have subsection on how to run, another with a list and explanation/catalog (e.g., script that identifies which tutorials pertain to which components).
* [ ] Add (more) links to videos.

Minor fixes:

* [ ] Links to code examples from github.io splash page: consider pointing to EarthscapeHub instead of Binder
* [ ] Same for the "ten minutes to Landlab" tutorial
* [ ] Same for Tutorials page
* [ ] The "ten minutes" intro should explain why you would choose jupyter vs. lab instance of hub.
* [ ] Consider having "papers and presentations" more easily accessible/findable.
* [ ] On github.io if you click Install, then page back up, further clicks on Install have no effect.
* [ ] Consider using the CSDMS version of the Landlab logo.
* [ ] "Software dev practices" is duplicated in the TOC.
* [ ] Consistent dev install between Install and Dev Guide sections.
* [ ] "Additional resources" under dev install needs to be cleaned up, most can go away.
* [ ] OverlandFlow users guide should go either in .py or a notebook.
* [ ] Same with CellLabCTS guide.
* [ ] Add a link to FramedVoronoiGrid.
* [ ] Clean up blank "ongoing contributions" and "desired contributions" under Dev Guide -> Guidelines for Contributing Code.
* [ ] Some nice eye candy for the landing page would be good.
* [ ] "Intro to Python" can go away.

Aspirational:

- It would be great to have a prominent image gallery with accompanying code examples!
- Similarly, it would be great to have a "literature gallery" with figures from papers and links to references.
[...]

PR #1730, opened 2023-09-06
Docs clean-up and add a notebook gallery

mcflugen:
### Description

This pull request continues the docs clean-up and also adds a gallery for both the tutorial notebooks and teaching notebooks.

Based on #1637, I've removed some old sections that were out-of-date or could be better found elsewhere on the internet and moved some stuff around,
* Intro to Python and numpy
* How to transition from *Landlab* v0 to v1 and v2
* Removed the examples section
* Removed the FAQ section
* Made the Papers and Presentations, and "citing Landlab" more prominent by adding links to them on the main page.

To create the notebook galleries, I moved the *notebooks/tutorials* and *notebook/teaching* folders into the *docs/source*. I had to do this because *nbsphinx*, which creates the galleries, can't deal with links or folders outside of the docs folder. I've replaced the old folder locations with symbolic links so people can still get to the notebooks as they did before.

The docs generated by this branch can be viewed on [readthedocs](https://landlab.readthedocs.io/en/mcflugen-clean-up-docs-issue-1637/).

### Checklist - did you ...

- [x] All tests have passed?


mcflugen:
@gregtucker, @mdpiper Could you have a look at this? I think the notebook galleries are turning out well. With the notebooks embedded in the docs, it makes it significantly easier for people to browse through our collection. In addition, visitors can use the search bar to look for all notebooks that contain, for example, a particular component they're interested in, which is something that we've had requests for.

A couple notes on the gallery:
* Because we save the notebooks without output, which I think is the correct thing to do, the notebooks on *readthedocs* also don't contain output. If we wanted to include the output, we could run all of the notebooks before building the rest of the docs. This is easy to do (it's just a matter of setting a flag in our *conf.py* file) but it would mean it would take a looooong time to build the docs.
* For the most part, the notebook thumbnails are just the *Landlab* favicon. The default is to use the last generated image in the notebook but, since we don't save any output, there is not image to use.

gregtucker:
This is fabulous! I agree it would be ideal to have the notebooks run before building the docs. Is it possible to do this step (the running of notebooks for doc-building) *only* when updating the online docs version (i.e., whenever we do a release), and *not* when we are just running tests, say, on a draft PR?

mdpiper:
Would it be possible to organize the notebooks on a single page like the Matplotlib Gallery? I'm biased, because I love the Gallery (although they changed the name to [Examples](https://matplotlib.org/stable/gallery/index.html), which is about as bad a marketing choice as renaming HBO to Max). Also, the pages with only one or two entries, like *Advection*, look bare.

Here's what I'm thinking. On the left navbar, under *Tutorials*, have a single entry, *Gallery*, replacing the individual pages *Advection*, *Agent Based Modeling*, etc. The Gallery page could then collect all of these individual notebooks under subheadings. Further, maybe include a brief description under each subheading (e.g., what is "Carbonates"?).

I'd be happy to help with this, if you'd like.

PR #1735, opened 2023-09-13
Add transparency option to imshow

keckje:
### Description

Added transparency (alpha) option to imshow

### Checklist 

no update to documentation because alpha is one of the **kwds

ran black on updated imshow.py and cleaned lint after checking with flake8


mcflugen:
Looks good @keckje. Thank you! I also added the option for transparency with non-raster grids.

mcflugen (approved):
👍

PR #1738, opened 2023-09-14
Fix parallel_links_at_link doctest

mcflugen:
### Description

This pull request fixes two failing doctests related to *parallel_links_at_link* and *orientation_of_link*.

The doctest for *parallel_links_at_link* assumed that links on the perimeter of a grid had no parallel links (i.e. their parallel links were returned as `[-1, -1]`) but the code didn't make this assumption. I've updated the doctest to now enforce that all links (on a structured grid of quadrilaterals, anyway) have at least one parallel link.

The doctest for *orientation_of_link*, which was added with *parallel_links_at_link*, was failing because the returned array is of type `uint8`, but the doctest was written for it being a plain old `int`. I've updated the doctest to include `dtype=uint8`.

### Checklist - did you ...

- [x] All tests have passed?

### DIFF
--- a/landlab/graph/hex/hex.py
+++ b/landlab/graph/hex/hex.py
@@ -763,10 +763,10 @@ class HexGraphExtras:
         >>> import numpy as np
         >>> grid = HexModelGrid((3, 2))
         >>> grid.orientation_of_link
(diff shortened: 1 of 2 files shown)


mcflugen:
@gregtucker Does this look good to you? These are the failing doctests we talked about this morning.

gregtucker (approved):
+1

PR #1743, opened 2023-09-20
OverlandFlow: add support for spatially non-uniform rainfall

DavidLitwin:
### Description

I'm toying with routing the local runoff produced by GroundwaterDupuitPercolator with OverlandFlow to produce more reasonable hydrographs. However, at the moment OverlandFlow only allows the "rainfall_rate" to be a float. I looked through the code, and found no reason why it shouldn't be allowed to be an array of floats at node, except the setter was checking for positivity in a way that was only compatable with float. I just changed the setter (using return_array_at_node now) and updated the docstring. 

### Checklist

- [x] All tests have passed?


landlab/components/overland_flow/generate_overland_flow_deAlmeida.py
```diff
@@ -368,7 +369,7 @@ def dt(self):
 
     @dt.setter
     def dt(self, dt):
-        assert dt > 0
+        assert dt > 0, "timestep dt must be positive"
```
mcflugen, line 372: I would rather remove the *assert* here. Instead, how about:
```python
if dt <= 0:
    raise ValueError("time step must be positive")
```
*assert* statements should really only be used in debugging since they are only run when the Python ``__debug__`` variable is set to ``True``.

landlab/components/overland_flow/generate_overland_flow_deAlmeida.py
```diff
@@ -380,11 +381,14 @@ def rainfall_intensity(self):
...
-    def rainfall_intensity(self, rainfall_intensity):
-        if rainfall_intensity >= 0:
-            self._rainfall_intensity = rainfall_intensity
+    def rainfall_intensity(self, new_val):
+        message = "Rainfall intensity must be positive"
+        if isinstance(new_val, float):
+            assert new_val >= 0, message
         else:
-            raise ValueError("Rainfall intensity must be positive")
+            assert (new_val >= 0).all(), message
```
mcflugen, line 389: Same as the above regarding the *assert*. Instead:
```python
if np.any(new_val < 0.0):
    raise ValueError("Rainfall intensity must be positive")
```
This should handle both scalar and array intensities.

landlab/components/overland_flow/generate_overland_flow_deAlmeida.py
```diff
@@ -380,11 +381,14 @@ def rainfall_intensity(self):
...
+
+        self._rainfall_intensity = return_array_at_node(self._grid, new_val)
```
mcflugen, line 391: I'm not sure you need to use ``return_array_at_node`` here. Can you not just have,
```python
self._rainfall_intensity = new_val
```

mcflugen (changes requested):
Looks good, just a few comments. Thanks @DavidLitwin!

mcflugen:
@DavidLitwin This is great! Thank you!

In addition to the above changes, could you also please add a test or two that ensure rainfall intensities of arrays really do work and they give the same results as scalar values?

DavidLitwin:
Thanks @mcflugen! I should know better about `assert` - it was used in the dt setter before, so now they are both changed. Added a news item as well.

DavidLitwin:
Hi @mcflugen, not sure what's going on with the remaining checks, but I think I'm all ready here.

mcflugen (approved):
👍

PR #1754, opened 2023-10-06
Fix CI builds for Python 3.12

mcflugen:
### Description

This pull request fixes our builds that began failing when *conda* added Python 3.12 as the default version. Note that I've done this by disabling Python 3.12 builds. I'm not sure one can build *Landlab* within a Python 3.12 environment yet. That, I hope, will come shortly with a subsequent pull request.

It appears, at least for the time being, some of *Landlab*'s dependencies might not be available for Python 3.12, which is now the default Python for *conda*. Because of this, some of our CI builds were failing while trying to create a testing environment. To fix this, I've specified Python versions (3.11, for now) in all of our *nox* sessions that use the *mamba* backend. Then, for the ci builds, I use the ``--force-pythons`` option to force *nox* to create a new environment with the Python version we want to test against.

I'm not certain but it appears that previously to this pull request, all of our tests were using the default version of Python and not the version our tests were supposed to be targeting. Anyway, with this pull request, we are now testing against the versions of Python we think we are.

### Checklist - did you ...

- [x] All tests have passed?

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -9,7 +9,7 @@ PROJECT = "landlab"
 ROOT = pathlib.Path(__file__).parent
 
 
-@nox.session(venv_backend="mamba")
+@nox.session(python="3.11", venv_backend="mamba")
 def test(session: nox.Session) -> None:
     """Run the tests."""
     os.environ["WITH_OPENMP"] = "1"
@@ -38,7 +38,7 @@ def test(session: nox.Session) -> None:
         session.run("coverage", "report", "--ignore-errors", "--show-missing")
 
 
-@nox.session(name="test-notebooks", venv_backend="mamba")
+@nox.session(name="test-notebooks", python="3.11", venv_backend="mamba")
 def test_notebooks(session: nox.Session) -> None:
     """Run the notebooks."""
     args = [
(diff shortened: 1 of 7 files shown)


mdpiper:
@mcflugen Thank you for doing this. I think this explains what @gantian127 and I were seeing in the CI builds earlier this week.

mcflugen:
> I think this explains what @gantian127 and I were seeing in the CI builds earlier this week.

Yup, the timing sounds about right. I hope this works for you as well. I think when we define *nox* sessions that use the *conda*/*mamba* backend, we should always specify a *Python* version to install. If we don't, I think the default gets installed, regardless of the version that's installed in the calling environment.

When you have a moment, could you give this a quick review?

noxfile.py
```diff
@@ -9,7 +9,7 @@
 ROOT = pathlib.Path(__file__).parent
 
 
-@nox.session(venv_backend="mamba")
+@nox.session(python="3.11", venv_backend="mamba")
```
mdpiper, line 12: This is at your discretion, but would it be good to define a global `DEFAULT_PYTHON="3.11"` above, then reference it here and on line 41? I'm thinking this may make it easier to change in the future.

mdpiper (dismissed):

noxfile.py
mcflugen: I agree. In fact, I've done exactly that locally but hadn't pushed the change to GitHub yet.

Issue #1901, opened 2024-03-22
SPACE deviates from analytical solution for mixed bedrock-alluvial river in a portion of the parameter space


cmshobe:
@BCampforts @gregtucker relevant also to Susannah's current work

SPACE large scale eroder does not match the bedrock-alluvial analytical solutions for the portion of the parameter space that causes "fast" sediment dynamics (high V, high H*, high Ks and Kr). Works fine for DL and TL end-member cases, but there is a substantial portion of the parameter space it doesn't like for the mixed case. 

Screenshot shows that slopes become higher than predicted in downstream regions, while sediment thickness is lower than predicted in those areas. Below the screenshot I paste the built-in test with a parameter set that causes it to fail.

<img width="765" alt="Screenshot 2024-03-22 at 15 22 54" src="https://github.com/landlab/landlab/assets/9465696/6c314647-e4c6-4d3b-940d-8ab829c0454b">

```python
import matplotlib.pyplot as plt
import numpy as np
from landlab import RasterModelGrid
from landlab.components import PriorityFloodFlowRouter, DepressionFinderAndRouter, FlowAccumulator, SpaceLargeScaleEroder
from numpy import testing

# set up a 5x5 grid with one open outlet node and low initial elevations.
nr = 5
nc = 5
mg = RasterModelGrid((nr, nc), xy_spacing=10.0)

z = mg.add_zeros("topographic__elevation", at="node")
br = mg.add_zeros("bedrock__elevation", at="node")
soil = mg.add_zeros("soil__depth", at="node")

np.random.seed(seed=5000)
mg["node"]["topographic__elevation"] += (
    mg.node_y / 100000 + mg.node_x / 100000 + np.random.rand(len(mg.node_y)) / 10000
)

mg.set_watershed_boundary_condition_outlet_id(
    0, mg["node"]["topographic__elevation"], -9999.0
)
soil[:] += 0.0  # initial condition of no soil depth.
br[:] = z[:]
z[:] += soil[:]

# Create a D8 flow handler
fa = FlowAccumulator(
    mg, flow_director="D8", depression_finder="DepressionFinderAndRouter"
)

# Parameter values for detachment-limited test
K_br = 0.005
K_sed = 0.01
U = 0.0001
dt = 10.0
F_f = 0.  
m_sp = 0.5
n_sp = 1.0
v_s = 1.0
H_star = 1.0

# Instantiate the SpaceLargeScaleEroder component...
sp = SpaceLargeScaleEroder(
[...]

PR #1907, opened 2024-04-16
Bugfix for issue #1901: deviation from analytical solution for mixed bedrock-alluvial river

BCampforts:
### Description

This PR is resolving a bug as reported in issue #1901 where SpaceLargeScaleEroder deviates from analytical solution for mixed bedrock-alluvial river in a portion of the parameter space. It seems that the error arises from the part in the code where the 'formation of potholes' is prevented. Yet this interferes with the proper calculation of bedrock incision and hence, this section has been removed in this bugfix. 

### Checklist - did you ...

- [x] Add a [news fragment](https://landlab.readthedocs.io/en/master/development/contribution/index.html#news-entries) file entry if necessary?
- [x] Add / update tests if necessary?
- [x] Add new / update outdated documentation?
- [x] All tests have passed?
- [x] Formatted code with black?
- [x] Removed lint reported by flake8?
- [x] Sucessful documentation built? (if documentation added or modified)

PR #1908, opened 2024-04-19
Adding more tests to Bcampforts space large scale issue

cmshobe:
This PR adds further testing to the fix made by @BCampforts to issue #1901, in particular to the high vs portion of the parameter space where the problem was occurring. @mcflugen sorry for duplicating the PR... I envisioned first merging into Benjamin's PR but I'm getting a little rusty.


BCampforts:
Additional test is now added to PR #1907

Issue #1909, opened 2024-04-19
Landlab dev Compilation errors with bothering folders .ipynb_checkpoints that automatically creates (jupyterhub)


sebastien-lenard:
Hi!
I have some compilation error when running pip install -e . on the current master branch of landlab pulled down from github (as a developer as in https://landlab.readthedocs.io/en/master/install/index.html). The error raised is:

`landlab/components/bedrock_landslider/.ipynb_checkpoints/cfuncs-checkpoint.pyx:1:0: 'landlab.components.bedrock_landslider..ipynb_checkpoints.cfuncs-checkpoint' is not a valid module name`

I didn't created these checkpoints files. And checked github master, they do not appear either.
 I will temporarily bypass the problem by removing the .ipynb_checkpoints folders, but it would be great if those folders were not included in the compilation.

mdpiper:
This may be related: the built-in editor in JupyterLab creates `.ipynb_checkpoints` directories.

sebastien-lenard:
Thank you @mcflugen, yes, i think your proposition to make explicit the modules which are compiled will reduce the number of errors.

I didn't investigate that much, but I feel that .ipynb_checkpoints at created even without starting a notebook, but just having it open in a tab of Jupyter.

PR #1915, opened 2024-04-23
List all extensions in setup.py

mcflugen:
### Description

This pull request adds a list of all the extensions to be built to ``setup.py``. The downside of doing this is that this list must be manually updated when adding/removing extension modules but the upside is that it is more explicit and prevents unwanted extensions being found as in #1909.

### Checklist - did you ...

- [x] All tests have passed?

### DIFF
--- a/news/1915.misc
+++ b/news/1915.misc
@@ -1,4 +1,4 @@
 
 Added a list of Landlab's extensions to setup.py that must be
 maintained manually. This replaces the old, and somewhat buggy,
-conducting a recursive glob for pyx files.
+method of conducting a recursive glob for pyx files.
--- a/news/1915.misc
+++ b/news/1915.misc
@@ -1,0 +1,4 @@
+
+Added a list of Landlab's extensions to setup.py that must be
+maintained manually. This replaces the old, and somewhat buggy,
+conducting a recursive glob for pyx files.
--- a/setup.py
+++ b/setup.py
@@ -2,12 +2,15 @@
 
 import os
 
-import numpy
+import numpy as np
 from Cython.Build import cythonize
 from setuptools import Extension
 from setuptools import setup
 
-extensions = (
+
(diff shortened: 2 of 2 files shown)


mcflugen:
@sebastien-lenard would you please have a quick look at this and let me know what you think?

A couple things:
1. All of *Landlab*'s extensions are now listed in ``setup.py``. Although maybe a bit more difficult to maintain, I think it's easier to understand and prevents extensions being built that shouldn't be built (as was the case in #1909).
2. There is an extension in the tests folder (``test_breach_c.pyx``), which looks to be yours, that I'm not sure about. I don't see why these tests need to be written as *cython*. It seems like they could more simply be written as Python and test the *priority_flow_routing* module directly.

sebastien-lenard (dismissed):
Tested and worked on Jupyter hub.

sebastien-lenard:
Hi @mcflugen !

_All of Landlab's extensions are now listed in setup.py. Although maybe a bit more difficult to maintain, I think it's easier to understand and prevents extensions being built that shouldn't be built (as was the case in https://github.com/landlab/landlab/issues/1909)._
Yes. 

_There is an extension in the tests folder (test_breach_c.pyx), which looks to be yours, that I'm not sure about. I don't see why these tests need to be written as cython. It seems like they could more simply be written as Python and test the priority_flow_routing module directly._
I can't directly call from a python script fully cythonized functions (),  in particular the ones having a nogil attributes (if I remember well), I need to pass through this intermediary cython script.

mcflugen:
> I can't directly call from a python script fully cythonized functions (), in particular the ones having a nogil attributes (if I remember well), I need to pass through this intermediary cython script.

Thanks for that context. I think we can probably simplify this so that we can test those functions without having to compile additional cython code. If so, we'll do that in a separate pull request.

PR #1920, opened 2024-04-25
Fixed bug in calculation of storage volumes

pfeiffea:
In main NST component, stored volume was calculated as sediment (rock + pore space) volume, but later assumed to be just rock volume. This bug resulted in systematically slightly-too-large aggradation when pulses are added. Thanks for catching, Ashanie! 

Changes: 
- minor edit to one line of code to fix this bug
- removed a few old, commented out scraps

Issue described for posterity at [https://github.com/landlab/landlab/issues/1921](https://github.com/landlab/landlab/issues/1921)


pfeiffea:
Hi @mcflugen this is a very minor (in the code) bug fix.

Issue #1930, opened 2024-05-04
Provide effective sediment erosion and deposition rates from SpaceLargeScaleEroder


BCampforts:
In order for SpaceLargeScaleEroder to be compatible with future concentrations trackers (@loroberge, @nicgaspar), we need access to effective sediment and deposition rates as calculated within the 'calc_sequential_ero_depo.pyx' cython method.

PR #1931, opened 2024-05-04
Provide effective sediment erosion and deposition rates from SpaceLargeScaleEroder

BCampforts:
### Description
This PR addresses Issue #1930 and aims to make SpaceLargeScaleEroder compatible with future concentrations trackers (@loroberge, @nicgaspar). We provide effective sediment and deposition arrays as return arguments of the run one step function in the SpaceLargeScaleEroder. Alternatively, they could be provided as fields but since these arrays only need the be accessed for this specific goal (tracking concentrations) I opted for return arguments instead. In addition to some minor changes in the calc_sequential_ero_depo back calculating effective sediment erosion and deposition, return arguments are added in the SpaceLargeScaleEroder and two additional tests are added.  

### Checklist - did you ...

- [x] All tests have passed?


BCampforts:
@loroberge, I propose to keep this PR open until the concentration tracking is going. Just in case we need more things to resolve.

loroberge:
@BCampforts , you mention the following about the erosion and deposition arrays:

> Alternatively, they could be provided as fields but since these arrays only need the be accessed for this specific goal (tracking concentrations) I opted for return arguments instead.

After some thought, I believe it would be better to provide them as fields. This makes it easier to couple the ConcentrationTracker component by automatically accessing the fields once instantiated. It also makes it easy for users to understand where sediment is moving to/from (which could come in useful for SpaceLargeScaleEroder users and will certainly be important for ConcentrationTracker users).

BCampforts:
> @BCampforts , you mention the following about the erosion and deposition arrays:
> 
> > Alternatively, they could be provided as fields but since these arrays only need the be accessed for this specific goal (tracking concentrations) I opted for return arguments instead.
> 
> After some thought, I believe it would be better to provide them as fields. This makes it easier to couple the ConcentrationTracker component by automatically accessing the fields once instantiated. It also makes it easy for users to understand where sediment is moving to/from (which could come in useful for SpaceLargeScaleEroder users and will certainly be important for ConcentrationTracker users).

@mcflugen what is your take on this? I am not sure what the best practice is in this case. Creating extra fields seems to be the more elegant solution but they will only be needed in specific cases (like with the concentration trackers) in other cases, it might create redundancy. Also not sure what is best performance and memory wise: adding extra fields or providing return arguments which might or might not (in most of the cases) be used.

loroberge:
> > @BCampforts , you mention the following about the erosion and deposition arrays:
> > > Alternatively, they could be provided as fields but since these arrays only need the be accessed for this specific goal (tracking concentrations) I opted for return arguments instead.
> > 
> > 
> > After some thought, I believe it would be better to provide them as fields. This makes it easier to couple the ConcentrationTracker component by automatically accessing the fields once instantiated. It also makes it easy for users to understand where sediment is moving to/from (which could come in useful for SpaceLargeScaleEroder users and will certainly be important for ConcentrationTracker users).
> 
> @mcflugen what is your take on this? I am not sure what the best practice is in this case. Creating extra fields seems to be the more elegant solution but they will only be needed in specific cases (like with the concentration trackers) in other cases, it might create redundancy. Also not sure what is best performance and memory wise: adding extra fields or providing return arguments which might or might not (in most of the cases) be used.

@mcflugen I'm hoping to crank out the ConcentrationTracker stuff in the next few weeks. As @BCampforts said, the most elegant solution is to create fields. This would also be the most user-friendly. What do you think?

gguryan:
I tried checking out the branch from this PR and whenever I try to import any Landlab component (including SpaceLargeScaleEroder), I get this error even though I'm not trying to import anything from flexure.

`ModuleNotFoundError: No module named 'landlab.components.flexure.ext.flexure1d'`

In the interim I manually copied the changes from this PR into my own branch, but when I run the example code in the docstring and save the return arguments as variables (vol_SSY_riv, V_leaving_riv, ero_sed_effective, depo_effective), the arrays for both ero_sed_effective and depo_effective are all zeros - is this correct? 

Also, I wanted to chime in to say that having the effective ero and depo rates as fields would also be useful for my work on the dynamic width version of the model

mcflugen:
> ModuleNotFoundError: No module named 'landlab.components.flexure.ext.flexure1d'

@gguryan This sounds like a *numpy* 2.0 issue. The extension modules (eg. `landlab.components.flexure.ext.flexure1d`) do not currently build with *numpy* 2 but #1956 will fix this. In the meantime, I've limited *numpy* to <2 in *pyproject.toml* but that change had not been merged into this branch. Now that this branch is up-to-date, when you build *landlab* all of the extension modules should build and be importable.

Try pulling in these newest updates and reinstall *landlab* (i.e. `pip install -e .`).

gguryan:
> > ModuleNotFoundError: No module named 'landlab.components.flexure.ext.flexure1d'
> 
> @gguryan This sounds like a _numpy_ 2.0 issue. The extension modules (eg. `landlab.components.flexure.ext.flexure1d`) do not currently build with _numpy_ 2 but #1956 will fix this. In the meantime, I've limited _numpy_ to <2 in _pyproject.toml_ but that change had not been merged into this branch. Now that this branch is up-to-date, when you build _landlab_ all of the extension modules should build and be importable.
> 
> Try pulling in these newest updates and reinstall _landlab_ (i.e. `pip install -e .`).

This solved everything, thank you!

loroberge:
@mcflugen & @BCampforts , the `ConcentrationTracker` also needs the bedrock erosion rate from `SpaceLargeScaleEroder`. I have updated this branch with all the necessary changes but can't push to @BCampforts fork. What's the best way to move forward?

loroberge:
@BCampforts & @mcflugen, looks like this is working and passes all the tests. It works with my in-dev ConcentrationTracker and I won't need to make any other changes. You can go ahead and merge.

mcflugen (approved):
👍

PR #1946, opened 2024-06-20
Add SoilGrading component

yuvalsmilo:
### Description

This PR introduces a new component, SoilGrading, designed to calculate soil particles fragmentation and weathering and changes in the soil layer grain size distribution.  SoilGrading component allows tracking the mass of grains in various size classes. Temporal changes in the grain size distribution can be carried out according to different fragmentation patterns. 

### Checklist - did you ...

- [x] All tests have passed?


yuvalsmilo:
Hey @gregtucker you might want to have a look at this PR

landlab/components/soil_grading/soil_grading.py
```diff
@@ -209,14 +208,15 @@ def __init__(
...
+        if std is None:
             std = self._initial_median_size * self._CV
         self._std = std
         self._is_bedrock_distribution_flag = is_bedrock_distribution_flag
         self._precent_of_volume_in_spread = precent_of_volume_in_spread
 
         # Create out fields.
-        # Note: Landlabs' init_out_field procedure will not work for the 'grains__weight' and 'grains_fractions__size' fields
+        # Note: Landlabs' init_out_field procedure will not work
+        # for the 'grains__weight' and 'grains_fractions__size' fields
```
gregtucker, line 219: Why are the fields configured with 2-3 dimensions instead of 1-2? That is, why isn't median_size__weight just number_of_nodes, and grains_fractions__size (for example) just number_of_nodes x number_of_size_classes?

gregtucker, line 219: Also, I'm wondering whether the try...except around soil__depth is necessary. Seems like this forces there not to be a pre-existing soil__depth field, yet I can imagine some applications wanting to start off with a given soil__depth field.

landlab/components/soil_grading/soil_grading.py
```diff
@@ -386,11 +386,11 @@ def create_dist(
         else:
 
             total_bedrock_weight = 10000  # There is no meaning for this number
-            if median_size == None:
+            if median_size is None:
```
gregtucker, line 389: Suggest adding a docstring to this function to explain briefly what it's for.

landlab/components/soil_grading/soil_grading.py
```diff
@@ -439,14 +439,14 @@ def create_dist(
             )
 
     def _generate_normal_distribution(
-        self, median_size=None, std=None, total_soil_weight=None
+        self, median_size = None, std = None, total_soil_weight = None
```
gregtucker: Suggest docstring to explain what the function does / how it works.

gregtucker (dismissed):
Looks great to me, just a few minor suggestions.

landlab/components/soil_grading/soil_grading.py
yuvalsmilo: Thanks. Yes, it forces that soil__depth field will not be created outside of SoilGrading. The point is-- if a soil__depth field is created outside of SoilGrading, the grid is not associated with the corresponding grain size information. The idea is not to allow this to happen in case SoilGrading is in use. Note that if a user wants to start with a given soil__depth field, SoilGrading can be initialized with a desired grain size distribution and total soil weight and it will create a corresponding soil__depth field.

yuvalsmilo: Oops, no reason. Thanks. I will change it accordingly.

landlab/components/soil_grading/soil_grading.py
yuvalsmilo: Added

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+            for _ in range(num_of_size_classes_plusone - 1):
+                upperlimits.append(maxsize)
+                maxsize = lower_limit_of(maxsize)
+                lowerlimits.append(maxsize)
+
+            self._upperlims = np.sort(upperlimits)
+            self._lowerlims = np.sort(lowerlimits)
+            self._meansizes = (
+                np.array(self._upperlims) + np.array(self._lowerlims)
+            ) / 2
```
mcflugen, line 366: How about just saving a `limits` array rather than both the upper and lower limits?

So rewrite as,
```python
        if self._input_sizes_flag:
            self._limits = np.insert(self._meansizes, 0, 0.0)
        else:
            limits = [self._grain_max_size]
            for _ in range(0, self._n_sizes):
                limits.append(limits[-1] * (1 / self._N) ** (self._power_of))
            self._limits = np.asarray(limits[::-1])
        self._meansizes = (self._limits[1:] + self._limits[:-1]) * 0.5
```

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+        grain size distribution can be assigned to the initial soil layer and
+        the bedrock layer
+        """
+
+        if not is_bedrock_distribution_flag:
+
+            is_bedrock_distribution_flag = self._is_bedrock_distribution_flag
+            median_size = self._initial_median_size
+            std = self._std
+            total_soil_weight = self._initial_total_soil_weight
```
mcflugen, line 385: No need to create temporary variables, just use the class attributes directly.

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+            grains_weight__distribution = self._generate_normal_distribution(
+                median_size=median_size, std=std, total_soil_weight=total_soil_weight
+            )
+
+        else:
+            total_bedrock_weight = 10000  # There is no meaning for this number
+            if median_size is None:
+                grains_weight__distribution = self.g_state0
+            else:
+                median_size = median_size
```
mcflugen, line 395: Remove.

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+                if std is None:
+                    std = self._CV * median_size
+                else:
+                    std = std
```
mcflugen, line 399: Remove.

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+            self.g_state = (
+                np.ones(
+                    (
+                        int(np.shape(self._grid)[0]),
+                        int(np.shape(self._grid)[1]),
+                        int(len(grains_weight__distribution)),
+                    )
+                )
+                * grains_weight__distribution
+            )
```
mcflugen, line 416: I think you can rewrite this more succinctly as,
```python
self.g_state = np.full(
    self.grid.shape + (len(grains_weight__distribution),), grains_weight__distribution
)
```

mcflugen (changes requested):

gregtucker:
It would be great if this could work with any grid type!

mcflugen:
> It would be great if this could work with any grid type!

I agree. I think all that would need to be done is to use something like `grid.area_of_cell` instead of `grid.dx * grid.dy`.

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+                if sample >= lower and sample <= upper:
+                    values.append(sample)
+
+            grains_weight__distribution = np.histogram(
+                values, np.insert(self._upperlims, 0, 0)
+            )[0]
+
+        return grains_weight__distribution
+
+    def update_median__size(self):
```
mcflugen, line 488: Rename without the double underscore and, while you're at it, maybe rename it to `update_median_grain_size`?

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+
+        # Note: Landlabs' init_out_field procedure will not work
+        # for the 'grains__weight' and 'grains_classes__size' fields
+        # because the shape of these fields is: n_nodes x n_grain_sizes.
+        grid.add_field(
+            "median_size__weight",
+            np.zeros(grid.shape[0] * grid.shape[1]),
+            at="node",
+            dtype=float,
+        )
```
mcflugen, line 227: This can be written more simply as,
```python
grid.add_zeros("median_size__weight", at="node")
```

and the subsequent ones as, for example,
```python
grid.at_node["grains_classes__size"] = np.ones((grid.number_of_nodes, self._n_sizes))
```

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+                dtype=float,
+            )
+
+        if "bedrock__elevation" not in grid.at_node:
+            grid.add_field(
+                "bedrock__elevation",
+                np.zeros_like(self._grid.nodes.flatten(), dtype=float),
+                at="node",
+                dtype=float,
+            )
```
mcflugen, line 271: As above, use `grid.add_zeros`. It's fewer lines of code and avoids creating an temporary array.

mcflugen:
@yuvalsmilo I would like to clean up some of the keyword parameters for `SoilGrading.__init__`. What do you think of replacing all of `n_of_grainsize_classes`, `meansizes`, `grain_max_size`, and `power_of` with just a single keyword, `bins` (*bins* is what *numpy.histogram* uses but we can go with something else). `bins` would be a monotonically increasing iterable that defines the upper and lower limit for each grain size class (so it would have a length of `n_classes + 1`).

For the case of a user wanting to use `N`, `power_of`, `grain_max_size`, etc. to calculate the grain class bins, we would provide a separate function to do so. For example,
```python
@staticmethod
def calc_bin_edges(n_bins, upper_edge, a=0.5, b=1):
    bin_edges = [upper_edge]
    for _ in range(n_bins):
        bin_edges.append(bin_edges[-1] * a ** b)
    return bin_edges[::-1]
```
What do you think about doing something like that? It would clean up the keyword parameters as well as making it easier to add new distributions in the future.

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+            self._grid.at_node["bed_grains__proportions"][
+                self._grid.core_nodes
+            ] *= np.divide(self.g_state_bedrock, np.sum(self.g_state_bedrock))
+
+        else:
+            self.g_state_bedrock = grains_weight__distribution
+            self._grid.at_node["bed_grains__proportions"][self._grid.core_nodes] = 1
+            self._grid.at_node["bed_grains__proportions"][
+                self._grid.core_nodes
+            ] *= np.divide(self.g_state_bedrock, np.sum(self.g_state_bedrock))
```
mcflugen, line 443: This looks like repeated code. How about removing one of these blocks and then dedenting the other?

landlab/components/soil_grading/soil_grading.py
mcflugen: Also, some more explanation to go with the comment, "There is no meaning to this number" would be good. 😄

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+                stacklevel=2,
+            )
+
+        elif median_size > upper:
+            grains_weight__distribution = np.zeros_like(self._upperlims)
+            grains_weight__distribution[-1] = total_soil_weight
+            warnings.warn(
+                "Median size provided is larger than the distribution upper bound",
+                stacklevel=2,
+            )
```
mcflugen, line 474: I wonder if these warnings should be errors. If someone is requesting a distribution that has a mean outside of the specified range, that doesn't seem like it should be allowed.

landlab/components/soil_grading/soil_grading.py
yuvalsmilo: Will do

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+       [ 1.66666667,  1.66666667,  1.66666667,  1.66666667,  1.66666667],
+       [ 1.66666667,  1.66666667,  1.66666667,  1.66666667,  1.66666667]])
+
+Run one step of soil fragmentation
+>>> soil_grading.run_one_step()
+>>> grid.at_node["median_size__weight"].reshape(grid.shape)
+array([[ 0.0028249,  0.0028249,  0.0028249,  0.0028249,  0.0028249],
+       [ 0.0028249,  0.0028249,  0.0028249,  0.0028249,  0.0028249],
+       [ 0.0028249,  0.0028249,  0.0028249,  0.0028249,  0.0028249],
+       [ 0.0028249,  0.0028249,  0.0028249,  0.0028249,  0.0028249]])
```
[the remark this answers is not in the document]
yuvalsmilo: Thanks, I will update the tests to include only the core nodes.

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+
+Create a grid on which to simulate soil grading.
+>>> grid = RasterModelGrid((4, 5))
+
+Create topographic_elevation and bedrock__elevation fields
+>>> _ = grid.add_zeros("topographic__elevation", at="node")
+>>> _ = grid.add_zeros("bedrock__elevation", at="node")
+
+Initialise the SoilGrading component
+>>> soil_grading = SoilGrading(grid)
```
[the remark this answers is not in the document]
yuvalsmilo: Thanks. I changed the default value to 0.5 and updated the tests accordingly.

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+            grains_weight__distribution = np.zeros_like(self._upperlims)
+            grains_weight__distribution[-1] = total_soil_weight
+            warnings.warn(
+                "Median size provided is larger than the distribution upper bound",
+                stacklevel=2,
+            )
+
+        else:
+            while np.size(values) < total_soil_weight:
+                sample = random.gauss(median_size, std)
```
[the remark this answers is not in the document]
yuvalsmilo: Added.

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+
+        try:
+            grid.add_field(
+                "soil__depth",
+                np.zeros(int(grid.shape[0] * grid.shape[1])),
+                at="node",
+                dtype=float,
+            )
+        except KeyError as exc:
+            raise ValueError("Soil field already exists") from exc
```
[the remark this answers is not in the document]
yuvalsmilo: Changed to: `grid.add_zeros("soil__depth", at="node")`.
Also, I changed the `KeyError` to `UserWarning` that will raise in case `soil__depth` field already exists (in this case, `soil__depth` will be overwritten by `SoilGrading`). 
Tests changed accordingly.

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+        CV: float, optional
+            Coefficient of variance of the grain size distribution.
+        is_bedrock_distribution_flag: bool, optional
+            A flag to indicate if the grain size distribution is generated for soil or
+            bedrock layer.
+        precent_of_volume_in_spread: float, optional
+            The precent of volume transferred from parent to daughter in case of 'spread'
+             grading pattern
+        """
+
```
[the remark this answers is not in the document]
yuvalsmilo: I will modify the component to work with any grid type

landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,540 @@
...
+        def lower_limit_of(maxsize):
+            lower_limit = maxsize * (1 / self._N) ** (power_of)
+            return lower_limit
+
+        upperlimits = []
+        lowerlimits = []
+        num_of_size_classes_plusone = self._n_sizes + 1
+
+        if input_sizes_flag:
```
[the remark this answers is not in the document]
yuvalsmilo: Changed.

landlab/components/soil_grading/soil_grading.py
yuvalsmilo: Sounds good.

landlab/components/soil_grading/soil_grading.py
yuvalsmilo: Got it.

landlab/components/soil_grading/soil_grading.py
yuvalsmilo: I removed `median_size = media_size`.
I think the `else` block is necessary here to differentiate between cases where `median_size` is provided or not. Without the `else` block, `grains_weight__distribution` will be rewritten. Am I missing something?

landlab/components/soil_grading/soil_grading.py
yuvalsmilo: Removed.

landlab/components/soil_grading/soil_grading.py
yuvalsmilo: Replaced.

yuvalsmilo:
> It would be great if this could work with any grid type!

Will do

landlab/components/soil_grading/soil_grading.py
mcflugen: @yuvalsmilo You're correct. I guess I wan't wearing my glasses and got the indentation mixed up!

landlab/components/soil_grading/soil_grading.py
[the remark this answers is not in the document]
mcflugen: Since the `soil__depth` field is an output field, it looks to me like you would want,
```python
if "soil__depth" not in grid.at_node:
    grid.add_empty("soil__depth", at="node")
grid.at_node["soil__depth"].fill(0.0)
```
There's no need to raise a warning—a user should expect an output field to be modified.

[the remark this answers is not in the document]
gregtucker: Maybe it should be an `inout`, so a user could provide an already-populated field as an initial condition?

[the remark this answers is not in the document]
mcflugen: I was thinking that too but then saw @yuvalsmilo's comment above https://github.com/landlab/landlab/pull/1946#discussion_r1660074058. It sounds as though the way to integrate a preexisting soil depth is to provide a grain size distribution and soil weight. `SoilGrading` then calculates the soil depth. I think.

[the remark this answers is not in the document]
yuvalsmilo: @gregtucker - yes, the user can init the component with a given grain size distribution, and then, the `soil__depth` field will be created accordingly. The case here is that `SoilGrading` component is initiated but a `soil__depth` field was already created for the grid (outside of `SoilGrading` ) and the match between grains mass and soil depth is not guarantee. In this case, `SoilGrading` will overwrite the `soil__depth` field to match grains mass.

landlab/components/soil_grading/soil_grading.py
yuvalsmilo: Changed

landlab/components/soil_grading/soil_grading.py
yuvalsmilo: Changed

yuvalsmilo:
> @yuvalsmilo I would like to clean up some of the keyword parameters for `SoilGrading.__init__`. What do you think of replacing all of `n_of_grainsize_classes`, `meansizes`, `grain_max_size`, and `power_of` with just a single keyword, `bins` (_bins_ is what _numpy.histogram_ uses but we can go with something else). `bins` would be a monotonically increasing iterable that defines the upper and lower limit for each grain size class (so it would have a length of `n_classes + 1`).
> 
> For the case of a user wanting to use `N`, `power_of`, `grain_max_size`, etc. to calculate the grain class bins, we would provide a separate function to do so. For example,
> 
> ```python
> @staticmethod
> def calc_bin_edges(n_bins, upper_edge, a=0.5, b=1):
>     bin_edges = [upper_edge]
>     for _ in range(n_bins):
>         bin_edges.append(bin_edges[-1] * a ** b)
>     return bin_edges[::-1]
> ```
> 
> What do you think about doing something like that? It would clean up the keyword parameters as well as making it easier to add new distributions in the future.

Hey @mcflugen, I think I do not entirely understand your point. Currently, if the user wants to init the component with a known grain size distribution he can do it by providing a list of `meansizes` instead of providing `bins`. Does that make sense?

mcflugen:
> Hey @mcflugen, I think I do not entirely understand your point. Currently, if the user wants to init the component with a known grain size distribution he can do it by providing a list of `meansizes` instead of providing `bins`. Does that make sense?

My main point is that you can reduce the number of keyword arguments related to grain size distribution from four to one. I believe the four keywords add unnecessary complexity to your component, making it more difficult to use, maintain, and test. I think it would be cleaner if a user simply passed a grain size distribution to your component (as they can do through `meansizes`). The other keywords`n_of_grainsize_classes`, `grain_max_size`, and `power_of` are not necessary.

If you like, you could provide some convenience functions (e.g. the `calc_bin_edges` function above) to create different distributions but I don't think this should be done within your component.

yuvalsmilo:
> > Hey @mcflugen, I think I do not entirely understand your point. Currently, if the user wants to init the component with a known grain size distribution he can do it by providing a list of `meansizes` instead of providing `bins`. Does that make sense?
> 
> My main point is that you can reduce the number of keyword arguments related to grain size distribution from four to one. I believe the four keywords add unnecessary complexity to your component, making it more difficult to use, maintain, and test. I think it would be cleaner if a user simply passed a grain size distribution to your component (as they can do through `meansizes`). The other keywords`n_of_grainsize_classes`, `grain_max_size`, and `power_of` are not necessary.
> 
> If you like, you could provide some convenience functions (e.g. the `calc_bin_edges` function above) to create different distributions but I don't think this should be done within your component.

@mcflugen, Ok got it now, but:
1. The thing is that a key element of the component is the ability to represent diffrent fragmentation/weathering patterns of soil grains (this is actually what the `run_one_step` procedure does). The idea is that a user can provide a desired fragmentation pattern by the `grading_name` keyword argument and then (together with  `grain_max_size`, and `power_of`) a grain size distribution that preserves this pattern is created. In other words,  `calc_bin_edges` is mandatory to represent a desired `grading_name`. I think that many users will find it useful NOT to provide `meansizes` but instead let the component build a grain size distribution that represents a specific fragmentation pattern. 
2. `calc_bin_edges` is actually calculate the `meansizes`. I changed the name and updated `limits` accordingly.

landlab/components/soil_grading/soil_grading.py
yuvalsmilo: I changed these warnings to be based on `meansizes`. Because `meansizes` are the center of grain size classes (bins), a user can ask for a specific median size that is within the distribution but is larger than the `grain_max_size`. I think it will be good to allow it, but, raise a warning.

yuvalsmilo:
Hey @mcflugen,
Please have a look at the new static function `create_grading`. It creates grading distribution based on `grading_name` and returns `meansizes`, `limits`, and `fragmentation_pattern`. 
Also, now `SoilGrading` can be initialized with user-defined `fragmentation_pattern` (generated by the static function or not).

src/landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,557 @@
...
+
+        return meansizes, limits, fragmentation_pattern
+
+    def check_fragmentation_pattern(self):
+        is_length = len(self._fragmentation_pattern) >= 2 * (
+            len(self._fragmentation_pattern) <= len(self._meansizes)
+        )
+        is_1 = np.sum(self._fragmentation_pattern) <= 1
+        if not is_length * is_1 is True:
+            raise ValueError("fragmentation pattern provided not valid")
```
mcflugen, line 363: @yuvalsmilo Could you clean this up? I find it hard to follow. Perhaps is should be the following,
```python
if (
    len(self._fragmentation_pattern) < 2
    or len(self._fragmentation_pattern) > len(self._meansizes)
    or np.sum(self._fragmentation_patter) > 1.0
):
    raise ValueError("fragmentation pattern provided not valid")
```

mcflugen:
@yuvalsmilo Once you fix up the failing tests, I think we can merge this.

src/landlab/components/soil_grading/soil_grading.py
yuvalsmilo: OK thanks, changed

src/landlab/components/soil_grading/soil_grading.py
```diff
@@ -0,0 +1,563 @@
...
+
+        self._A = np.zeros((self._n_sizes, self._n_sizes))
+        self._A_factor = np.ones_like(self._A) * self._A_factor
+
+        for i in range(self._n_sizes):
+
+            if i == 0:
+                self._A[i, i] = 0
+            elif i == self._n_sizes:
+                self._A[i, i] = -(1 - (self._fragmentation_pattern[0]))
```
mcflugen, line 386: @yuvalsmilo This line is never reached. Since the loop runs from 0 to `self._n_sizes - 1`, should this be,
```python
elif i == self._n_sizes - 1:
    self._A[i, i] = -(1 - self._fragmentation_pattern[0])
```

mcflugen: There's actually no need to have the *if* statement in the loop at all, is that right?
```python
self._A[0, 0] = 0.0
self._A[self._n_sizes - 1, self._n_sizes - 1] = -(1 - self._fragmentation_pattern[0])

for i in range(1, self._n_sizes - 1):
    self._A[i, i] = -(1 - self._fragmentation_pattern[0])
    ...

```

yuvalsmilo: @mcflugen, thanks. I have cleaned this code section.

mcflugen (approved):
👍

Issue #1947, opened 2024-06-24
Reshaping


ChristianMohr:
Dear all,
I am in the process of creating a Landlab model that combines different components, e.g. the ecohydrology and landslide components. While I had no problems using raster data provided to me, the model now has problems with the data I created myself.
I created raster data via QGIS, which I exported as ASCII. All rasters have the same resolution, are correctly aligned and have the same spatial extent. Everything has been checked several times. My raster data includes rasters of plant functional types, vegetation cover, soil density, soil thickness, biomass and a DEM. When I load the data as ASCII files into Landlab, I get the following error message:

**ValueError: unable to reshape array to field size (24011 != 24660)**

I suspect that this problem stems from where the data is stored, i.e. nodes or cells, which leads to this form error. Do you have any idea how to solve this problem? Does anyone know this problem and have any solutions?
Cheers and thanks in advance,
Christian

mcflugen:
@ChristianMohr I think your suspicion is correct. ``read_esri_ascii`` will try to add the data from your ASCII file to your grid's nodes. However, it sounds like your grid is sized such that the data should be placed at cells.

A couple ideas:
* You could use a smaller grid that whose shape matches that of your data file and then load the data onto its nodes.
* You could create a new grid from your data file using ``read_esri_ascii`` and then place that onto the cells of your current grid.

Your issue has reminded me that I had been working on enhancing the ESRI ASCII functionality a while ago but got distracted with something else and never finished the work. I'll get back on that and once I finish, I think it should make what you are trying to do easier.

ChristianMohr:
@mcflugen Thanks for the prompt reply. Would you have some exemplary code sniplets for that?

mcflugen:
Sure. Could you provide me with a minimal working example that reproduces the error? I can then modify it to, hopefully, fix your issue.

ChristianMohr:
Hi @mcflugen . I solved the issue - more or less. The error occurred when using one specific DEM dataset. I tried another dataset and it worked. For now, this is ok. However, I will see why this is occurring only when using one specific dataset.

PR #1948, opened 2024-06-24
Add load/dump functions for ESRI ASCII files

mcflugen:
### Description

It would be nice if *Landlab* functions to read/write to various data formats followed a standard interface. This pull request adds load/dump functions for the ESRI ASCII format. These are similar to those for VTK. The new functions also have some enhanced capabilities like being able to load data from a file and place it on different grid elements (i.e. nodes, corners, cells or faces), not just nodes (#1947).

### Checklist - did you ...

- [x] All tests have passed?


mdpiper (approved):
I think this is great! One comment: would it make sense to use the idiom `esri_ascii.load` instead of `load`, akin to `yaml.load` or `toml.load` (and the same for `dump`)? I think it makes it more transparent what sort of file is being read/written.

mcflugen:
> I think this is great! One comment: would it make sense to use the idiom `esri_ascii.load` instead of `load`, akin to `yaml.load` or `toml.load` (and the same for `dump`)? I think it makes it more transparent what sort of file is being read/written.

Thanks @mdpiper! Do you just mean in the examples and notebooks to use something like,
```python
from landlab.io import esri_ascii

with open(file) as fp:
    grid = esri_ascii.load(fp)
```

mdpiper:
@mcflugen Yup! Just in the examples & notebooks. (I'm still thinking about LL coding style.)

mcflugen:
> @mcflugen Yup! Just in the examples & notebooks. (I'm still thinking about LL coding style.)

Great. I've made those changes.

What do you think if I deprecate the old functions and move them to another module? I'm thinking moving them to a private module called `_deprecated_esri_ascii.py` and a user would import them as they can do now (i.e. `from landlab.io import read_esri_ascii`)?

mdpiper:
> What do you think if I deprecate the old functions and move them to another module? I'm thinking moving them to a private module called `_deprecated_esri_ascii.py` and a user would import them as they can do now (i.e. `from landlab.io import read_esri_ascii`)?

I think this a great idea because it'll help with backward compatibility.

PR #1950, opened 2024-06-25
Add Snow Components

### LINKED ISSUE #1949: Add SnowDegreeDay Component
This is a function request to add a model component to simulate the snow accumulation and melt process using the snow degree day method. This component is based on the method from TopoFlow modeling framework. See details at: 
https://github.com/peckhams/topoflow36/blob/master/topoflow/components/snow_degree_day.py
https://github.com/peckhams/topoflow36/blob/master/topoflow/components/snow_base.py

gantian127:
### Checklist - did you ...

- [x] All tests have passed?


gantian127:
@mcflugen This pull request is ready for code review. Thank you!

notebooks/tutorials/snow/era5_sdd.nc
mcflugen, line 1: I don't think this file is used, is that correct?

notebooks/tutorials/snow/snow_degree_day.ipynb
mcflugen, line 1: Replace
```python
print(SnowDegreeDay.__doc__)
```
with
```python
help(SnowDegreeDay)
```
Replace
```python
from tqdm import trange
```
with
```python
from tqdm.notebook import trange
```
I was able able to run the notebook without an API key (unless I have one and don't know it) so can we leave out that section that installs an API key?

mcflugen (commented):
@gantian127 Some initial comments.

landlab/components/snow/snow_degree_day.py
```diff
@@ -0,0 +1,300 @@
...
+
+import numpy as np
+
+from landlab import Component
+
+
+class SnowDegreeDay(Component):
+    """Simulate snowmelt process using degree-day method.
+
+      The degree-day method uese c0 (degree-day coefficient),
```
gregtucker, line 21: Typo (uses?)

landlab/components/snow/snow_degree_day.py
```diff
@@ -0,0 +1,300 @@
...
+      T0 (degree-day threshold temperature) and T_air (air temperature)
+      to determine the snow melt rate.
+      When T_air is larger than T0, the melt process will start.
+      The melt rate is calculated as SM = (T_air -T0) * c0
```
gregtucker, line 25: Suggest defining the units of SM and c0 (for example, if SM is in m/day or whatever)

landlab/components/snow/snow_degree_day.py
```diff
@@ -0,0 +1,300 @@
...
+
+    def __init__(self, grid, rho_H2O=1000, T_rain_snow=1, grid_area=0):
+        """Initialize SnowDegreeDay component"""
+
+        super().__init__(grid)
+
+        # parameters
+        self._rho_H2O = rho_H2O
+        self._T_rain_snow = T_rain_snow
+        self._grid_area = grid_area if grid_area > 0 else self._grid.dy * self._grid.dx
```
gregtucker, line 162: As @mcflugen mentioned, if you need this parameter, make it general enough to accommodate different grid types. There's a grid property grid.area_of_cell that contains the surface area of each grid cell (which might be uniform, as in raster and hex grids, or might be variable, as in VoronoiDelaunay grids)

gregtucker (commented):
Great to have this new component! I will try to look at the notebook next week. In the meantime, looks good to me pending @mcflugen's suggestions.

notebooks/tutorials/snow/snow_degree_day.ipynb
gantian127: The reason the notebook can run without an API key is that the era5_sdd.nc file exists. The ERA5 Data Component will check the output file name from the config yaml file. If there is a local file that matches the output file name, it will load the local file without downloading data from the remote server. Otherwise, an API key file is required for data download. 

If it is possible, I would like to keep the era5_sdd.nc file and the section to install the API key for the notebook. Because the data access request may be queued, which will take very long time or fail to download the data and break the notebook testing.

notebooks/tutorials/snow/era5_sdd.nc
gantian127: Please see my comment below. Thanks!

gantian127:
@mcflugen @gregtucker Ready for 2nd review. Thank you!

mcflugen:
@gantian127 I've made some changes that I haven't pushed to GitHub yet. Although there are quite a lot of them, they are mostly just refactoring. There are a couple, though, that I haven't made yet but that I would like your opinion on.

1. What do you think about removing removing `_vol_sm` and `_vol_swe`? These are two quantities that a user of the component could easily calculate as needed rather than calculating and storing them every time step. In addition, removing them would no longer require the need for `grid_area`, which currently requires the grid to be raster. Instead, the user could use `grid.area_of_cell` as the grid size and handle perimeter nodes (those without cells) as they see fit, thus allowing this component to work with any grid type.
2. I think when calculating melt rate, the melt rate shouldn't be limited by the amount of snow to melt. Obviously, we can't melt more snow than is available but the melt rate should just be what is calculated by the temperature difference and the degree-day coefficient. In addition, as currently written, if new snow is being added while snow is also being melted (i.e. the air temperature is between the degree-day threshold and the rain-snow temperature), this would incorrectly limit the melt rate.

What do you think?

mcflugen:
@gantian127 Based on our conversation, I've made the above additional changes. Could you have a look and let me know if you have any questions or concerns?

gantian127 (commented):
Thank you very much for refactoring this component! I made a few comments and tested to run the notebook with long term data. They look good to me. Thanks!

landlab/components/snow/snow_degree_day.py
```diff
@@ -0,0 +1,361 @@
...
+        self, grid, c0=2, threshold_temp=0.0, rho_water=1000.0, rain_snow_temp=1.0
+    ):
+        """Initialize SnowDegreeDay component."""
+
+        super().__init__(grid)
+
+        self.c0 = c0
+        self.threshold_temp = threshold_temp
+        self.rho_water = rho_water
+        self.rain_snow_temp = rain_snow_temp
```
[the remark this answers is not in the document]
mcflugen: These line are using the setters, which then set the private versions of the attributes. It's, admittedly, confusing because the private versions aren't explicitly set within the class's ``__init__`` method. I wrote it this way so I wouldn't have to rewrite some of the checks that ensure the parameters are positive.

mcflugen:
Thanks for your suggestions @gantian127! I think I've addressed each of them.

mcflugen:
@gantian127 I've rebased this branch with master so that there are no longer any conflicts. This means, though, that you ***will have to update your local branch if you want to make any additional changes***.

gantian127:
@mcflugen @gregtucker Could you please help review this pull request? Thank you!

PR #1965, opened 2024-08-02
Add FieldTracker class

mcflugen:
### Description

This pull request adds a `FieldTracker` class and an associated context manager, `open_tracker`.

@Sheehace I finally found a bit of time to work on something you requested some time (years?) ago. Although the code looks quite a bit different than what you had sent me, I think it does basically what you are wanting (i.e. tracking values of a grid's fields). The `FieldTracker` class doesn't track things like mean, standard deviation, etc. like yours did but, instead, leaves that up to the user. I think one could also use the `FieldTracker` as a base class to create a trackers that did track some statistics like what your example did. Anyway, please have a look and see if it makes sense and may be of some use to you.

@loroberge You may want to have a look at this as well. It's much simpler that your concentration tracker but is similar, I think, in spirit. Maybe there is something here you can use?

The docstrings and unit tests show various ways that it can be used but the following gives a simple example,
```python
>>> import numpy as np
>>> from landlab.field.tracker import open_tracker

>>> _ = grid.add_ones("z", at="node")

>>> with open_tracker(grid) as tracker:
...     grid.at_node["z"] += 2.0

>>> tuple(tracker.items())
(('at_node:z',
  [array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]),
   array([3., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3.])]),)

>>> tracker.reduce(np.subtract)
(('at_node:z', array([2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2., 2.])),)
```

The  `FieldTracker` object provides a read-only mapping as a means to access the tracked fields. For example,
```python
>>> tracker["at_node:z"]
(array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]),
 array([3., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3.]))
>>> [name for name in tracker]
['at_node:z']
>>> len(tracker)
1
>>> "at_node:z" in tracker
True
>>> tuple(tracker.keys())
('at_node:z',)
>>> tuple(tracker.values())
((array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]),
 array([3., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3.])),)
>>> tracker.get("at_node:z")
(array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]),
 array([3., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3., 3.]))
```

Because it's a read-only mapping, there is no `__setitem__`, `__delitem__`, `pop`, etc.

### Checklist - did you ...

- [x] All tests have passed?


mcflugen:
If you did want to calculate the standard deviation of some tracked field you could do something like,

```python
>>> import numpy as np
>>> from landlab.field.tracker import open_tracker

>>> _ = grid.add_ones("z", at="node")

>>> with open_tracker(grid) as tracker:
...     grid.at_node["z"] += 2.0
...     tracker.checkpoint()
...     grid.at_node["z"] += 1.0
...     tracker.checkpoint()

>>> fields = dict(tracker.fields)
>>> np.std(fields["at_node:z"], axis=0)
```

PR #1976, opened 2024-09-03
Add Stream Power Component

anweiii:
### Checklist - did you ...

- [x] All tests have passed?


mcflugen:
@anweiii When you have a chance, will you please have a look at the changes I made? I tidied up the docstrings for `ErosionDeposition` and `SharedStreamPower` so they render nicely as html, which you should be able to see on [readthedocs](https://landlab--1976.org.readthedocs.build/en/1976/reference/components/erosion_deposition.html). I also modified your code so that some of the input parameters can be either constants, arrays (one value for each node), or a string (that refers to an *at-node* field). Although the docs said this was already the case, this wasn't quite true. Now `runoff_rate`, `v_s`, and `sp_crit` can all be spatially variable.

mcflugen:
@anweiii I also changed the names of two parameters: `K_d` is now `k_bedrock` and `K_t` is now `k_transport`. There are a lot of parameters in *Landlab* called `K`, I think the extended names are more descriptive and make it easier for a user to keep straight.

anweiii:
Hi Eric,

Thanks for your help! Yes I can look over the changes once I get home. I
also just realized today that the title of my notebook only says stream
power, not shared stream power, so I need to fix that. I won’t be able to
sit down with my computer for a couple of hours, so I’m not sure if I can
get to it before the end of your workday.

Best,

Annie

anweiii:
@mcflugen I looked over all your changes and read over the docstring and it looks good! I also just fixed my notebook title, so everything should be ready.

mcflugen (approved):
👍

PR #1979, opened 2024-09-04
Add RiverFlowDynamics: A New Landlab 2D Flow Model

angelmons:
**_RiverFlowDynamics: A New Landlab 2D Flow Model_**

### Description

We are adding our 2D flow solver to Landlab. This component implements a semi-implicit, semi-Lagrangian finite-volume approximation of the depth-averaged shallow water equations, originally proposed by Casulli and Cheng in 1992, along with subsequent related work.

### Checklist - did you ...

- [x] All tests have passed?


mcflugen:
> It seems that nox is using an older version of SciPy

*nox* uses pinned versions of packages defined in the requirement files in the *requirements* folder. It looks like you just need to resync your fork with the latest version of *Landlab*.

news/1978.component.rst
mcflugen, line 1: Rename this file so that it points to the correct PR (i.e. #1979).

angelmons: Thanks!, I did it

angelmons:
Hi @mcflugen, I edited the files and made some small changes. Can you please approve the workflow?

Thanks!!

angelmons:
> > It seems that nox is using an older version of SciPy
> 
> _nox_ uses pinned versions of packages defined in the requirement files in the _requirements_ folder. It looks like you just need to resync your fork with the latest version of _Landlab_.

Thanks for the suggestion @mcflugen. We did it and all issues regarding that problem were solved.

angelmons:
Hello @mcflugen, would it be possible to take a look at this new component? We believe it will be a great contribution to Landlab

mcflugen:
@angelmons This branch has some conflicts with the *master* branch. If you're ok with it, I can rebase your branch with *master*, but that will mean you'll need to update your local branch if you make additional changes. Otherwise, you can rebase it and I'll update my local version. Which would you rather?

angelmons:
> @angelmons This branch has some conflicts with the _master_ branch. If you're ok with it, I can rebase your branch with _master_, but that will mean you'll need to update your local branch if you make additional changes. Otherwise, you can rebase it and I'll update my local version. Which would you rather?

@mcflugen go ahead and rebase my branch with _master_. Thanks

src/landlab/components/river_flow_dynamics/river_flow_dynamics.py
```diff
@@ -0,0 +1,1937 @@
...
+>>> linksAtCenter = grid.links_at_node[np.array(np.arange(600, 660))][:-1, 0]
+>>> flow_velocity = grid["link"]["surface_water__velocity"][linksAtCenter]
+>>> flow_velocity = np.round(flow_velocity, 3)
+
+The average (absolute) difference between predited and expected
+flow velocity in percentage is:
+
+>>> np.round(np.abs(np.mean(flow_velocity_expected - flow_velocity)) * 100, 0)
+0.0
+"""
```
mcflugen, line 280: @angelmons doctests are primarily intended as documentation, which is also tested, but not as a real test of your component. The example you provide here is probably better suited as an example in a Jupyter notebook. Would it be possible to simplify your example for this doctest. Perhaps using a small grid of, maybe, 5 rows and columns? It should demonstrate to a user how to use your model and what it is doing.

tests/components/river_flow_dynamics/test_river_flow_dynamics.py
```diff
@@ -0,0 +1,131 @@
...
+        3,
+    )
+    water_depth_obtained = grid.at_node["surface_water__depth"][
+        grid.nodes_at_right_edge
+    ]
+
+    water_depth_obtained = np.round(water_depth_obtained, 3)
+    np.testing.assert_array_almost_equal(
+        water_depth_solution, water_depth_obtained, decimal=3
+    )
```
mcflugen, line 131: @angelmons I would like to see more unit tests that demonstrate your model is working correctly. It looks to me like this is a snapshot test where you've run the model and then used that output as the expected value for the assertion. This is useful for identifying changes to the code that change the output, but not if the changes have fixed a bug or introduced a bug.

Are there analytic solutions you could test? Perhaps you could test different inputs to check if the model produces sensible results? Even though you may not know exactly the expected result, you probably still have an idea of what some bounds are (e.g. mass balance, non-negative water depths, etc.). Perhaps you could also test some edge cases to see if they also produce sensible results?

landlab/components/river_flow_dynamics/river_flow_dynamics.py
```diff
@@ -0,0 +1,2055 @@
...
+"""
+
+import numpy as np
+import scipy as sp
+
+from landlab import Component
+from landlab import FieldError
+
+
+class river_flow_dynamics(Component):
```
mcflugen, line 290: In Python, class names should be camel case.

mcflugen:
@angelmons It appears that when you merged your local copy of *riverFlowDynamics_dev* into the remote branch, you undid some of the changes from the rebase (notice, for example, there are now two copies of `river_flow_dynamics.py`). I think the best course of action would be to remove that merge commit by resetting the branch to the commit before that one. If I do this, it would mean, as with after the rebase, you would need to update your local branch before making any changes (i.e. something like `git reset --hard origin/riverFlowDynamics_dev`—note that doing this will discard any local changes that have not been committed).

What do you think?

angelmons:
> @angelmons It appears that when you merged your local copy of _riverFlowDynamics_dev_ into the remote branch, you undid some of the changes from the rebase (notice, for example, there are now two copies of `river_flow_dynamics.py`). I think the best course of action would be to remove that merge commit by resetting the branch to the commit before that one. If I do this, it would mean, as with after the rebase, you would need to update your local branch before making any changes (i.e. something like `git reset --hard origin/riverFlowDynamics_dev`—note that doing this will discard any local changes that have not been committed).
> 
> What do you think?

I apologize for doing that. I am confused about when/how do I have to do that. 

Please, go ahead and reset the branch. I do have a copy in my computer of the files I modified yesterday and know exactly what those are. So, feel free to do as you think is better.

Sorry for this confusion,

Thanks @mcflugen for your help.

mcflugen:
> I apologize for doing that. I am confused about when/how do I have to do that.

@angelmons No need to apologize! In my attempt to make things easier, I think I made them more complicated than they needed to be.

I believe I've cleaned up the branch now. To do so, I had to rewrite history and so did a force push, which means you will need to update your local copy of this branch. Before doing so, make sure you have a copy just in case something goes awry. A couple options:
1. `git reset --hard origin/riverFlowDynamics_dev`: this will reset your current branch to what it on the remote branch (i.e. here on GitHub). As noted above, this will discard any local changes you might have and set your branch to exactly what's here.
2. Reclone your fork into a new folder, which will grab this version of your branch, and then just work from that new ocal copy.

angelmons:
Hi @mcflugen, can you authorize the workflow, please? Thanks

angelmons:
Hello @mcflugen - I have been checking the checks report and for about one week it says "Some checks haven’t completed yet
46 successful and 1 expected checks"

There is one check "Run the tests Expected — Waiting for status to be reported" that does not start. Do I need to do something else?

mcflugen:
@angelmons This is all looking good to me, don't worry about that check that has run yet. Thank you so much for all of the your recent changes! They look great!

My only remaining question is about one of your tests. You have the following in one of your tests,
```python
    water_depth_solution = np.round(
        np.array(
            [
                0.4357753,
                0.4357753,
                0.43611027,
                0.43624251,
                0.43626605,
                0.43595278,
                0.43534349,
                0.43491662,
                0.43342158,
                0.43342158,
            ]
        ),
        3,
    )
```
Where do these numbers come from? It's to clear to me why these are the expected result?

angelmons:
Hi @mcflugen, you make a good point. This particular test comes from a verified example we used during development. While it serves as a regression test to catch unintended changes in the code, you're right that the numbers alone don't demonstrate physical correctness.

However, I've also implemented several physics-based tests that validate the model behavior:

test_uniform_flow: Compares results with Manning's equation
test_mass_conservation: Verifies conservation in closed systems
test_still_water: Checks stability of static conditions
test_numerical_stability: Validates behavior across different time steps

Would you prefer I remove this snapshot test since we now have these more rigorous physical validation tests?"


> @angelmons This is all looking good to me, don't worry about that check that has run yet. Thank you so much for all of the your recent changes! They look great!
> 
> My only remaining question is about one of your tests. You have the following in one of your tests,
> 
> ```python
>     water_depth_solution = np.round(
>         np.array(
>             [
>                 0.4357753,
>                 0.4357753,
>                 0.43611027,
>                 0.43624251,
>                 0.43626605,
>                 0.43595278,
>                 0.43534349,
>                 0.43491662,
>                 0.43342158,
>                 0.43342158,
>             ]
>         ),
>         3,
>     )
> ```
> 
> Where do these numbers come from? It's to clear to me why these are the expected result?

mcflugen:
> I've also implemented several physics-based tests that validate the model behavior

I saw that, and they look great! Thank you for adding those.

> Would you prefer I remove this snapshot test since we now have these more rigorous physical validation tests?"

No, I think it's ok if you leave them but maybe I'll add a note saying where it comes from.

I think this is ready to go so I'll rebase it with master and then merge it. We can make additional changes as new pull requests.

mcflugen (approved):
👍

mcflugen:
@angelmons Thank you for all your hard work on this! 🎉

Issue #1986, opened 2024-09-13
Plot_network_and_parcels doesn't play well with others


pfeiffea:
We'd like to be able to plot `NetworkSedimentTransporter` results using `plot_network_and_parcels` on top of a DEM. For example, using `imshowgrid`. However, it seems that `plot_network_and_parcels` doesn't play well with others - it creates a new subplot and covers over the grid. 

In addition, it returns an inline figure, rather than allowing the user to specify what to return, as `imshowgrid` does. 

I'm working on initial fixes. 

FYI @kbarnhart @mcflugen @Ashaniekam

`
import matplotlib.pyplot as plt
import numpy as np

from landlab.grid.create_network import network_grid_from_raster
from landlab.components import (
    BedParcelInitializerUserD50,
    FlowDirectorSteepest,
    NetworkSedimentTransporter
)
from landlab.io import read_esri_ascii
from landlab.plot.graph import plot_links, plot_nodes
from landlab.plot.imshow import imshow_grid
from landlab.plot import plot_network_and_parcels

grid, z = read_esri_ascii("hugo_site.asc", name="topographic__elevation")
grid.status_at_node[grid.nodes_at_right_edge] = grid.BC_NODE_IS_FIXED_VALUE
grid.status_at_node[np.isclose(z, -9999.0)] = grid.BC_NODE_IS_CLOSED

network_grid = network_grid_from_raster(
    grid,
    reducer=JustEndNodes(),
    minimum_channel_threshold=12000.0,
    include=["drainage_area", "topographic__elevation"],
)

network_grid.at_link["channel_width"]=np.ones(network_grid.number_of_links)
network_grid.at_link["reach_length"]=np.ones(network_grid.number_of_links)
network_grid.at_link["flow_depth"]=np.ones(network_grid.number_of_links)
network_grid.at_node["bedrock__elevation"]=network_grid.at_node["topographic__elevation"].copy()

initialize_parcels = BedParcelInitializerUserD50(
    network_grid,
    user_d50=0.05,
)

parcels = initialize_parcels()

fd = FlowDirectorSteepest(network_grid, "topographic__elevation")
fd.run_one_step()

nst = NetworkSedimentTransporter(network_grid,parcels,fd,)

nst.run_one_step(1)

plt.figure("DEM_and_Network")
imshow_grid(
    grid,
    "topographic__elevation",
    color_for_closed=None,
    allow_colorbar=False,
)

plot_network_and_parcels(
    network_grid,
    parcels,
[...]

PR #1988, opened 2024-09-14
Fix plot network and parcels

pfeiffea:
### Description

See issue #1986 
@Ashaniekam wants to be able to use `plot_network_and_parcels` in a straightforward way on top of `imshowgrid` (or similar), but it wasn't cooperating. I made two minor edits to bring `plot_network_and_parcels` into alignment with the other plotting tools: 
1. `plot_network_and_parcels` now has options for output, same as `imshowgrid` (used to automatically plot inline right away)
2. when you pass it an existing figure (e.g. DEM created via `imshowgrid`) via the `fig=` keyword argument, `plot_network_and_parcels` will plot the network on the same axes as the existing figure axes. 

Warning: I did not take into account making space for other preexisting fig items like colorbars or legends associated with the figure (e.g. DEM) that is being added to. So... if you find yourself needing to plot with that many different labels/legends I think you get to fight matplotlib on your own. 

### Checklist - did you ...

- [x] All tests have passed?

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -250,6 +250,7 @@ intersphinx_mapping = {
     "numpy": ("https://numpy.org/doc/stable/", None),
     "scipy": ("https://docs.scipy.org/doc/scipy/", None),
     "xarray": ("https://docs.xarray.dev/en/stable/", None),
+    "matplotlib": ("https://matplotlib.org/stable/", None),
 }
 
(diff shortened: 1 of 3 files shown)


pfeiffea:
@mcflugen Let me know if you think this warrants a new test, or if it's safe without given the minor edits.

Issue #2072, opened 2024-12-06
Enhancing Interactivity in Tutorial Notebooks


Snowdog85123:
Hi, I would like to propose some enhancements on the interactivity and educational value of the tutorial notebooks in the Landlab repository. By integrating more interactive components, users will be able to experiment with different parameters and directly observe the impact on model outputs, which will lead to a deeper understanding of the processes being simulated.

An example of an interactive notebook, such as the SimpleSubmarineDiffuser tutorial (https://github.com/landlab/landlab/blob/master/docs/source/tutorials/marine_sediment_transport/simple_submarine_diffuser_tutorial.ipynb), could be enhanced by allowing users to adjust the wave_base and shallow_water_diffusivity parameters through sliders. The changes would be reflected instantly in both the plot and in the model’s output, and users would see how these parameters influence the diffusion process.

PR #2073, opened 2024-12-06
Add interactive sliders to SimpleSubmarineDiffuser tutorial

Snowdog85123:
This PR addresses issue #2072 by adding interactive sliders for wave_base and diffusivity.

### Description
This pull request addresses issue #2072 by adding interactive sliders for wave_base and shallow_water_diffusivity to the SimpleSubmarineDiffuser tutorial notebook. These changes will allow users to dynamically adjust the model parameters and instantly observe how they influence the sediment diffusion process. The interactive elements are built using ipywidgets to enhance user engagement and educational value.

### Checklist - did you ...

- [x] All tests have passed?
- [ ] Formatted code with black?


mcflugen:
@Snowdog85123 Thanks for this contribution. I like the idea of adding sliders to the submarine diffusion to allow a user to see how the diffusion changes with wave base and diffusivity.

A couple things:
* You've added your notebooks to the top-level folder, which is the wrong location for them. Instead, they should sit alongside the other notebooks in the `docs/source/tutorials` folder. Which subfolder your notebook should be placed in is another question.
* You've created two new notebook but I think you only need, at most, one. Either add a new notebook that demonstrates how to add slider widgets to control a component, or add a few cells that add sliders to the existing submarine diffuser notebook. I'm leaning toward the former, in which case it should probably go in the visualization folder.
* Instead of using *ipywidgets*, I would rather you use *mpl-interactions*. I can help with that if you need but I find it to be simpler to use.

mcflugen:
I just poked around a bit and it looks like something along the lines of the following should work.

Create the grid and a new`SimpleSubmarineDiffuser`.
```python
grid = RasterModelGrid((3, 51))
grid.set_closed_boundaries_at_grid_edges(False, True, False, True)

z0 = (25.0 - np.abs(grid.x_of_node - 25.0)) - 15.0
z0[z0 < -5.0] = -5.0

grid.at_node["topographic__elevation"] = z0.copy()

ssd = SimpleSubmarineDiffuser(
    grid, sea_level=0.0, wave_base=1.0, shallow_water_diffusivity=1.0
)
```
Create a function that calculate the diffused topography.
```python
def diffused_topography(x, wave_base, diffusivity):
    ssd.grid.at_node["topographic__elevation"] = z0.copy()
    ssd.wave_base = wave_base
    ssd.shallow_water_diffusivity = diffusivity

    for _ in range(100):
        ssd.run_one_step(0.2)

    return ssd.grid.at_node["topographic__elevation"][51:102]
```
Create the plot.
```python
%matplotlib ipympl

x = grid.x_of_node[51:102]

fig, ax = plt.subplots()

controls = iplt.plot(x, diffused_topography, wave_base=(0.5, 5.0), diffusivity=(0.5, 5.0), label="Diffused Topography")
plt.plot(x, z0[51:102], "k", label="Initial Topography")
plt.plot([x[0], x[-1]], [0, 0], "b:", label="Sea Level")
plt.grid(True)
plt.xlabel("Distance (m)")
plt.ylabel("Elevation (m)")
plt.title("Effect of Diffusivity and Wave Base")
plt.legend()
plt.show()
```

Snowdog85123:
Hi @mcflugen, sorry for the late reply.

I understand what you mean and I will try to use mpl-interactions as you mentioned and put the file in the right file folder later.

Issue #2030, opened 2024-11-18
Creating notebook tutorial regarding issue #1982


vinceverdugo:
Hi, I am a new Landlab user and contributor. I saw issue #1982 about creating a notebook tutorial for different examples of combining Landlab and PyGMT, and this is something I would like to attempt. In the Developer Guide I could not really find any guidelines for documentation such as notebook tutorials, so if there is any information that is necessary if anyone could please let me know. Thank you!

gregtucker:
@vinceverdugo thanks for this! I actually wrote a basic tutorial notebook in PR #[1983](https://github.com/landlab/landlab/pull/1983). It's currently hung up on a windows installation issue with pygmt. If you would like to contribute a companion tutorial to this one, that would great. I also wonder if you or @gantian127 or @mcflugen or anyone else knows of a solution to this pygmt-windows installation snag?

PR #1983, opened 2024-09-06
Add notebook tutorial showing how to use PyGMT with Landlab

gregtucker:
### Description

This PR adds a notebook tutorial showing how PyGMT and Landlab can be used together to:

1) Use PyGMT's topography-import function to initialize a Landlab grid and topography field;
2) Use PyGMT's visualization tools to make nice maps;
3) Translate between PyGMT and Landlab grid formats.

Note: requires adding PyGMT as a dependency for the notebooks. PyGMT itself has gdal among its dependencies.

### Checklist 

- [x] Add a [news fragment](https://landlab.readthedocs.io/en/master/development/contribution/index.html#news-entries) file entry if necessary?
- [x] Add / update tests if necessary?
- [X] Add new / update outdated documentation?
- [ ] All tests have passed?
- [x] Formatted code with black?
- [x] Removed lint reported by flake8?
- [x] Successful documentation built? (if documentation added or modified)

PR #2074, opened 2024-12-07
Add companion notebook tutorial showing how to use Landlab and PyGMT together

vinceverdugo:
### Description

This PR addresses issue #2030. The notebook here is a companion tutorial notebook that expands on using Landlab and PyGMT together by creating an interactive erosion slider. This will allow learners to get more familiar with Landlab and PyGMT packages. 

### Checklist - did you ...

- [ x] Add new / update outdated documentation?
- [ x] All tests have passed?
- [ x] Formatted code with black?


mcflugen:
@vinceverdugo Thanks for this contribution. Although I like your idea, I don't see how this notebook demonstrates why a user would want to use *pyGMT* over the usual *landlab* plotting functions.

In addition, when I run the notebook, the slider at the end just shows an image of random data, regardless of the selected *K_sp* value.

mcflugen:
> In addition, when I run the notebook, the slider at the end just shows an image of random data, regardless of the selected _K_sp_ value.

@vinceverdugo Is this the case for you? Maybe there is something wrong with how my environment is set up?

vinceverdugo:
@mcflugen thank you for your feedback! My intention wasn't necessarily to demonstrate why a user would rather use PyGMT plotting over Landlab, but rather that it could be done. If you think it would be better that the tutorial use Landlab plotting instead that makes sense to me.

Regarding the slider, no I don't believe it is your environment. I had it working but I must have changed something because I am running into the same issue now. I will make sure to address that issue!

mcflugen:
> If you think it would be better that the tutorial use Landlab plotting instead that makes sense to me.

We have plenty of tutorials that show how to use *Landlab*'s usual plotting methods so I don't think an additional one is needed.

> My intention wasn't necessarily to demonstrate why a user would rather use PyGMT plotting over Landlab, but rather that it could be done.

Please have a look at the plots generated in #1983. It would be good if you could generate plots that look like those. By any chance, are you using Windows? If you look at #1983, you'll see we were having some issues on Windows when using PyGMT. If you can help with that problem, that would be wonderful.

PR #2083, opened 2024-12-16
Adjust pre-commit hook to ensure unix-style line endings

mcflugen:
### Description

This pull request adjusts the *mixed-line-ending* hook to ensure *lf* line endings for all files except Windows files (right now files with a `.bat` extension, which is just `make.bat`).

This will reduce the size of diffs where only line endings have changed.

This, I realize, is a gigantic changeset but, except for `.pre-commit-config.yaml`, all that's changed are line endings of files that had used *crlf*.

### Checklist - did you ...

- [x] All tests have passed?


gregtucker (dismissed):
Seems like a good idea to me to have uniform, unix-compatible line-endings. Is there a handy tool that someone working on windows can use to ensure the correct line endings?

mcflugen:
> Is there a handy tool that someone working on windows can use to ensure the correct line endings?

Good question. Most editors should, upon opening a file, notice the line-ending convention being used and stick to that. For new files, unless the user has set the default line-ending convention in their settings, I'm sure it would default to the convention used by their operating system. In such a case, running `nox -s lint` (or *pre-commit.ci*) would fix the line endings for those files.

PR #2098, opened 2024-12-30
Updating KinwaveImplicitOverlandFlow to accept distributed runoff and/or roughness

amanaster2:
### Description

Updated `KinwaveImplicitOverlandFlow` to accept fields or arrays of floats (in addition to scalars) for both `runoff_rate` and `roughness` keywords. Wanted to enhance the functionality for future use. 

### Checklist - did you ...

- [x] All tests have passed?


src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -228,8 +232,18 @@ def __init__(
...
+        # Allow roughness and runoff_rate to be distributed, if needed.
+        if isinstance(roughness, str):
+            self._roughness = self._grid.at_node[roughness]
+        else:
+            self._roughness = roughness
+
+        if isinstance(runoff_rate, str):
+            self._runoff_rate = self._grid.at_node[runoff_rate] / 3.6e6
+        else:
+            self._runoff_rate = runoff_rate / 3.6e6
```
mcflugen, line 244: We have a function that does this. The following is more compact and, I think, still does what you want,
```python
self._roughness = self._grid.return_array_or_field_values(roughness, at="node")
```

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -228,8 +232,18 @@ def __init__(
...
+
+        self._vel_coef = 1.0 / self._roughness  # do division now to save time
```
mcflugen, line 246: Because you calculate this in `__init__`, any changes to the roughness field will not be seen in future calculations. I think it would be better if this were moved into the `run_one_step` method. If you do that, there's probably no need to save this temporary calculation as a separate variable—just use the roughness directly.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -281,8 +295,11 @@ def runoff_rate(self):
 
     @runoff_rate.setter
     def runoff_rate(self, new_rate):
-        assert new_rate > 0
-        self._runoff_rate = new_rate / 3600000.0  # convert to m/s
+        if isinstance(new_rate, str):
```
mcflugen, line 298: As with a previous comment, you can use the `return_array_or_field_values` method.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -281,8 +295,11 @@ def runoff_rate(self):
...
+            self._runoff_rate = self._grid.at_node[new_rate] / 3.6e6
+        else:
+            assert new_rate > 0
```
mcflugen, line 301: Remove the `assert` statement, it should only be used for debugging. If you want an error to be raised, maybe something like the following,
```python
if new_rate <= 0.0:
    raise ValueError(f"runoff_rate must be positive ({new_rate})")
```

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -328,12 +345,23 @@ def run_one_step(self, dt):
...
+                    / (self._grid.area_of_cell[self._grid.cell_at_node[cores]])
+                )
+            elif hasattr(self._roughness, "__len__"):
+                # if manning's n in a field
+                self._alpha[cores] = (
+                    self._vel_coef[cores]
+                    * self._grad_width_sum[cores]
+                    * dt
+                    / (self._grid.area_of_cell[self._grid.cell_at_node[cores]])
+                )
```
mcflugen, line 364: To avoid repeated code and, I think, make this a little clearer, how about,
```python
if np.isscalar(self._roughness):
    roughness_at_core_nodes = self._roughness
else:
    roughness_at_core_nodes = self._roughness[cores]

self._alpha[cores] = (
    self._grad_width_sum[cores]
    * dt
    / (self._grid.area_of_cell[self._grid.cell_at_node[cores]])
    / roughness_at_core_nodes
)
```

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -345,11 +373,22 @@ def run_one_step(self, dt):
...
+                        dt
+                        * self._disch_in[n]
+                        / self._grid.area_of_cell[self._grid.cell_at_node[n]]
+                    )
+                elif hasattr(self._runoff_rate, "__len__"):
+                    ee = (dt * self._runoff_rate[n]) + (
+                        dt
+                        * self._disch_in[n]
+                        / self._grid.area_of_cell[self._grid.cell_at_node[n]]
+                    )
```
mcflugen, line 390: Same as above with the roughness, use `np.isscalar`.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -358,9 +397,20 @@ def run_one_step(self, dt):
...
+                        self._vel_coef
+                        * (Heff**self._depth_exp)
+                        * self._grad_width_sum[n]
+                    )  # this is manning/chezy/darcy
+                elif hasattr(self._roughness, "__len__"):
+                    outflow = (
+                        self._vel_coef[n]
+                        * (Heff**self._depth_exp)
+                        * self._grad_width_sum[n]
+                    )  # this is manning/chezy/darcy
```
mcflugen, line 413: Check if roughness is a scalar using `np.isscalar`.

tests/components/overland_flow/test_kinwave_implicit.py
```diff
@@ -133,8 +133,70 @@ def test_curved_surface():
...
+    kinwave1 = KinwaveImplicitOverlandFlow(
+        mg1, runoff_rate=r1, roughness=0.03, depth_exp=5 / 3
+    )
+
+    mg2 = RasterModelGrid((10, 10), xy_spacing=25)
+    mg2.add_zeros("surface_water__depth", at="node")
+    mg2.add_zeros("topographic__elevation", at="node")
+    mg2.set_closed_boundaries_at_grid_edges(True, True, True, True)
+    r2 = 1.0 * np.ones(100)
+    mg2.add_field("r2", r2, at="node")
```
mcflugen, line 156: Use,
```python
mg2.add_ones("r2", at="node")
```

tests/components/overland_flow/test_kinwave_implicit.py
```diff
@@ -133,8 +133,70 @@ def test_curved_surface():
...
+    kinwave1 = KinwaveImplicitOverlandFlow(
+        mg1, runoff_rate=1.0, roughness=r1, depth_exp=5 / 3
+    )
+
+    mg2 = RasterModelGrid((10, 10), xy_spacing=25)
+    mg2.add_zeros("surface_water__depth", at="node")
+    mg2.add_zeros("topographic__elevation", at="node")
+    mg2.set_closed_boundaries_at_grid_edges(True, True, True, True)
+    r2 = 0.03 * np.ones(100)
+    mg2.add_field("r2", r2, at="node")
```
mcflugen, line 186: Use,
```python
mg2.add_full("r2", 0.03, at="node")
```

tests/components/overland_flow/test_kinwave_implicit.py
```diff
@@ -133,8 +133,70 @@ def test_curved_surface():
...
+    kinwave2 = KinwaveImplicitOverlandFlow(
+        mg2, runoff_rate=1.0, roughness=r2, depth_exp=5 / 3
```
mcflugen, line 188: I think you want `roughness="r2"` to check that roughness can be passed as the name of an *at-node* field.

tests/components/overland_flow/test_kinwave_implicit.py
```diff
@@ -133,8 +133,70 @@ def test_curved_surface():
...
+    np.testing.assert_equal(kinwave1.depth, kinwave2.depth)
+
+
 if __name__ == "__main__":
     test_initialization()
     test_first_iteration()
     test_steady_basic_ramp()
     test_curved_surface()
+    test_kinwave_runoff_field()
+    test_kinwave_roughness_field()
```
mcflugen, line 202: None of this is necessary, I suggest removing these lines.

mcflugen (changes requested):
@amanaster2 Thanks for this contribution! I've added a few comments for you to address, but nothing major.

mcflugen:
> Add a [news fragment](https://landlab.readthedocs.io/en/master/development/contribution/index.html#news-entries) file entry if necessary? (Is this necessary?)

Yes, this is necessary, it will go into our release notes. Just one sentence that describes what you've added is fine.

> Successful documentation built? (if documentation added or modified) (This is repeatedly failing on my computer with an error about the module 'importlib_resources' not existing... I'm confused.)

The docs seem to be building fine, so you don't need to worry about this. There is a version of the docs [for this pull request](https://landlab--2098.org.readthedocs.build/en/2098/generated/api/landlab.components.overland_flow.generate_overland_flow_implicit_kinwave.html) you can have a look at to see that things look as you want them to.

tests/components/overland_flow/test_kinwave_implicit.py
amanaster2: Oops, definitely want `roughness="r2"`! Thanks for catching that.

tests/components/overland_flow/test_kinwave_implicit.py
amanaster2: Just to confirm: the whole block?

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
amanaster2: So it would also work for `self._runoff_rate  = self._grid.return_array_or_field_values(runoff_rate, at="node")`?

amanaster2:
Thanks, @mcflugen !! I'll get these changes incorporated and add a news fragment file entry.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -211,11 +211,15 @@ def __init__(
         ----------
         grid : ModelGrid
             Landlab ModelGrid object
-        runoff_rate : float, optional (defaults to 1 mm/hr)
+        runoff_rate : float or array of floats, optional (defaults to 1 mm/hr)
+            If array of floats, provide the field name str for where the values are
+            stored on the ModelGrid.
```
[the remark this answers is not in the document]
amanaster2: Units are on the next line; should I move them up?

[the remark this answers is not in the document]
mcflugen: Oops, I missed that. Sorry. You may as well move them up just to keep the line lengths consistent.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
amanaster2: Oops, this addresses my above question. Nevermind!

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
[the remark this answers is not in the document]
mcflugen: While you're at it, I think you can remove the sentence that begins with, "The value provided...".

tests/components/overland_flow/test_kinwave_implicit.py
mcflugen: Yeah, the whole block.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
mcflugen: Yes, that should also work.

As with the velocity coefficient, because you divide the runoff rate by 3.6e6 here, it creates a copy of the array and any changes to the field will not be reflected in future calculations. It probably makes sense to move some of this into `run_one_step`.

amanaster2: Since there's only one place in the `run_one_step` method where the runoff rate is used, I just moved the division by 3.6e6 to that location.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
amanaster2: Changed such that only roughness is used, and created a getter and setter for roughness so the value(s) can be updated easily.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
amanaster2: Note: `self._grid.return_array_or_field_values()` works only if the value provided _isn't_ a single float. Since I want to be able to use a single float or an array/field, I think I'm going to keep the code similar to what I had previously:
```python
if isinstance(roughness, float):
    self._roughness = roughness
else:
    self._roughness = self._grid.return_array_or_field_values(
        roughness, at="node"
    )

if isinstance(runoff_rate, float):
    self._runoff_rate = runoff_rate
else:
    self._runoff_rate = self._grid.return_array_or_field_values(
        runoff_rate, at="node"
    )
```

mcflugen: Ok. I think `return_array_or_field_values` should probably work with a scalar value but I'll make that change in a different pull request and then update this code later.

amanaster2: Oh cool. Does that mean we wouldn't have to do any `np.isscalar` checks further down in the code? Since `return_array_or_field_values` would return an array?

mcflugen: Yes, I think that would be the case but I haven't thought it through. There might be some reason we would want to keep a scalar, a scalar.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -281,13 +299,45 @@ def runoff_rate(self):
 
     @runoff_rate.setter
     def runoff_rate(self, new_rate):
-        assert new_rate > 0
-        self._runoff_rate = new_rate / 3600000.0  # convert to m/s
+        if new_rate <= 0.0:
+            raise ValueError(f"runoff_rate must be positive ({new_rate})")
+
+        if np.isinstance(new_rate, float):
```
mcflugen, line 305: Just `isinstance`, not `np.isinstance`.

amanaster2: Lol, I also fixed it before I saw that you fixed it :-D

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -211,11 +211,15 @@ def __init__(
         ----------
         grid : ModelGrid
             Landlab ModelGrid object
-        runoff_rate : float, optional (defaults to 1 mm/hr)
-            Precipitation rate, mm/hr. The value provided is divided by
-            3600000.0.
-        roughness : float, defaults to 0.01
-            Manning roughness coefficient; units depend on depth_exp.
+        runoff_rate : array_like of float or str (defaults to 1 mm/hr)
```
mcflugen, line 214: For both `runoff_rate` and `roughness`, how about:
```
runoff_rate : str or array_like of float
```
This makes it clear that it's not an array of either floats or strings but, rather, a string or an array of floats.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -211,11 +211,15 @@ def __init__(
...
+            a numpy array or the field name str of where the values
+            are stored on the nodes of the ModelGrid.
+        roughness : array_like of float or str (defaults to 0.01)
+            Manning's roughness coefficient(s); units depend on depth_exp.
+            If array, you may either provide a numpy array or the field
+            name str of where the values are stored on the nodes of the
+            ModelGrid.
         changing_topo : boolean, optional (defaults to False)
             Flag indicating whether topography changes between time steps
         depth_exp : float (defaults to 1.5)
```
mcflugen, line 225: For all of these keyword parameters, it's not needed to include the default values since they are given in the function signature.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -281,13 +299,45 @@ def runoff_rate(self):
...
+        Returns
+        -------
+        The current value of the roughness.
+        """
+        return self._roughness
+
+    @roughness.setter
+    def roughness(self, new_rough):
+        if new_rough <= 0.0:
+            raise ValueError(f"roughness must be positive ({new_rough})")
```
mcflugen, line 333: If `new_rough` is a `str`, this check will fail.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -281,13 +299,45 @@ def runoff_rate(self):
...
+            ModelGrid.
+
+        Returns
+        -------
+        The current value of the roughness.
+        """
+        return self._roughness
+
+    @roughness.setter
+    def roughness(self, new_rough):
```
mcflugen, line 331: I'm not sure if `new_rough` should be able to be a field name here. What do you think?

Are there cases where someone would want to change the field that contains the roughness?

What should happen if this component was created with a roughness specified as a field name and then a user subsequently uses this setter with a roughness that's an array? Should the field that was specified when the component was instantiated be updated?

amanaster2: Hmmm. This is a good question.

When I wrote this, I was thinking about a very specific case: a roadside ditch line that goes from a "bare" state (i.e., loose soil) to an "eroded" state (i.e., armored bed)---the roughness approximately doubles. This would probably happen over the course of a couple of storms, so if I used a single kinwave instantiation over the course of those storms, I'd probably want to update the roughness... But, like I said, this is a super specific case... So we could probably get rid of this setter, if it makes more sense to do so.

If we keep the setter: 
If the component is instantiated with a roughness field then is subsequently updated with an array, we would probably want to update the initial field, no? If the person wanted to maintain the initial values, they could create a copy. 

Let me think about this a bit more. I'll come back to this next week, but let me know if you have any thoughts in the meantime. 😄

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
amanaster2: Same with the check for `new_rate` above... Hrm. I'll come back to this when it's not after 11pm on a Friday 😴

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
mcflugen: I think the setter is a fine idea. My comment was about the specific case of the setter accepting a string, rather than an array-like object. Is there ever a case where you would want to change the field that the component is tracking for roughness values? That is, something like,
```python
overland_flow = KinwaveImplicitOverlandFlow(grid, roughness="roughness")
overland_flow.roughness = "roughness_new"
```
I'm not sure this is something we would want to support.

How about the following: if the component is instantiated using a string for the roughness, that tells the component to track that field for roughness values and if you want to update those values, you ***must to do that through the field***?

So maybe the setter would look something like,
```python
@roughness.setter
def roughness(self, new_rough):
    if isinstance(self._roughness, str):
        raise ValueError(
            "unable to set roughness. KinwaveImplicitOverlandFlow is currently tracking a node field"
            f" named {self._roughness} for roughness values"
        )
    if np.any(new_rough < 0.0):
        raise ValueError("roughness must be positive")
    self._roughness = np.asarray(new_rough)
```

amanaster2: OH, I see what you mean! Sorry, I misunderstood originally. Thank you for the clarification. 

Yes, I agree with this! Accepting a string in the setter would complicate things unnecessarily. 
I can make these updates today!

amanaster2: So, I realized that I made everything much more complicated than necessary by allowing `roughness` and `runoff_rate` to be fields. This was made apparent as I was trying to fix the setter (`self._roughness` never yielded a `str` with how I had coded the rest). After fumbling around for a while, I decided to streamline the code such that those keywords accept only array_like objects (and updated the docstrings and news fragment as necessary). I'm running `nox -s test` on the updates now to ensure everything is working, then I'll push my commits.

mcflugen: Are you sure? It seemed like a good idea to me! I do like simpler and streamlined, though.

amanaster2: I think so, yes! At least for now. We could merge this PR now (assuming it's all good), and include the fields in a future update of the component? (Should I open an issue and reference this PR as a reminder?) I'm going to be deep in Landlab land for at least the next 6 months, so I can definitely come back to it 😸

mcflugen: Sure, that sounds good. It's probably better as two separate pull requests, anyway.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -211,25 +211,24 @@ def __init__(
...
             implicit; 0 = explicit)
         """
         super().__init__(grid)
-        # Store parameters and do unit conversion
 
-        self._runoff_rate = runoff_rate / 3600000.0  # convert to m/s
-        self._vel_coef = 1.0 / roughness  # do division now to save time
+        # Store parameters
+        self._runoff_rate = runoff_rate
+        self._roughness = roughness
```
mcflugen, line 231: If either of these are arrays, a reference to the array is being saved. Is that what you want or should a copy be made? i.e.
```python
self._runoff_rate = np.array(runoff_rate)
```
If a reference is saved, then if the passed array is changed outside of the component, the component's value will also be updated. Using a reference would save memory and the time taken to make a copy but could, potentially lead to unexpected behavior.

mcflugen: The same applies lower down for the getters and setters.

mcflugen (changes requested):
A few more comments.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
amanaster2: Good point! I've made this change. 

Since all values for `runoff_rate` and `roughness` will now be arrays, I've also updated the code in `run_one_step` to check for a single float using `np.ndim() == 0` rather than `np.isscalar()`.

src/landlab/components/overland_flow/generate_overland_flow_implicit_kinwave.py
```diff
@@ -265,29 +264,33 @@ def __init__(
...
-        ----------
-        runoff_rate : float, optional (defaults to 1 mm/hr)
-            Precipitation rate, mm/hr. The value provide is divided by
-            3600000.0.
-
-        Returns
-        -------
-        The current value of the runoff rate.
-        """
+        """Runoff rate at nodes."""
```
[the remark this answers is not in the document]
amanaster2: I liked the idea of returning a read-only view of the arrays, so I did that! Committed and pushed, tests running now!

amanaster2:
Side note: I merged `master` into my branch, so the checks ran again, and it looks like there was 1 that errored and 1 that skipped. I didn't make any changes to my code.

amanaster2:
Hey @mcflugen ! Are there other changes I need to make to the component, or is it good to merge? (Once the checks finish; I just updated the branch to be in sync with `master`, so the checks are running again.) One of my colleagues was asking about using this updated version of the component, so I figured I'd give this a bump!

mcflugen:
@amanaster2 I think this is good to go but let me have another quick look.

mcflugen:
@amanaster2 Sorry, I added some tests to your branch that cover the code you modified and was waiting for you to have a look to see if they're ok. I see now, though, that I never actually asked to to have a look. So sorry! If they look good to you, this is ready to go.

amanaster2:
@mcflugen No worries!! I just looked through all of it, and it looks good to me!

PR #1911, opened 2024-04-20
New Component FlowRouter 2022 rebranch to current master 20240419

sebastien-lenard:
### Description

The new FlowRouter Component is a component of the family flow direction, accumulation and priority flood flow router. Its main interest is that it produces a filled topography ready to be used either directly by production of steepest descent flow accumulation or in combination of other flow routers, in particular the ones which split the flow in several directions.
The component can be used with any type of grid.
A full new notebook indicates the usages.

The 2022 PR is: FlowRouter component: Flow directions and depression filling for all type of grids #1461 https://github.com/landlab/landlab/pull/1487/files

### Checklist - did you ...

- [x] Add a [news fragment](https://landlab.readthedocs.io/en/master/development/contribution/index.html#news-entries) file entry if necessary?
- [x] Add / update tests if necessary?
- [x] Add new / update outdated documentation?
- [x] All tests have passed?
- [x] Formatted code with black?
- [x] Removed lint reported by flake8?
- [x] Sucessful documentation built? (if documentation added or modified)

PR #2127, opened 2025-02-07
Update the flow router for all grid types

mcflugen:
### Description

This pull request supersedes #1911.

### Checklist - did you ...

- [ ] Add a [news fragment](https://landlab.csdms.io/development/contribution/#news-entries) file entry if necessary?
- [ ] All tests have passed?
- [ ] Formatted code with black?


mcflugen:
@sebastien-lenard When you have a moment, would you please have a look at the failing unit tests? I think this pr should be ready once we get those taken care of.

mcflugen:
@sebastien-lenard I dug into this a bit and made a couple changes that fix the failing tests. I believe the source of the problem was that the sorting algorithm used by `np.argsort` wan't reproducibly sorting nodes when there were repeated nodes. Calling `np.argsort` with the `stable=True` keyword ([d7a6ecd](https://github.com/landlab/landlab/pull/2127/commits/d7a6ecdbe8d4a53155fef3d5f24fa68390123020) and [2c6190e](https://github.com/landlab/landlab/pull/2127/commits/2c6190e3115040a102c9dcadfa7adfec27627f3b)) ensures a consistent sorting by maintaining the relative order of repeated values. All of the tests (regardless of operating system or Python version) now generate the same sequence of sorted values.

After making this change, I had to modify the expected values of the tests to match the new output, essentially treating these as snapshot tests ([e422875](https://github.com/landlab/landlab/pull/2127/commits/e422875561f863f1be7c8551d4d6e50eba8ec54d)). It appeared to me that the actual ordering of these arrays wasn't important but you really should have a look and make sure I didn't break anything.

I don't know if it matters, but the stable sort is slower than other algorithms that don't ensure a consistent ordering. If it does matter, we can switch to *quicksort*, for example, but we would have to modify the tests to not rely on a particular ordering.

Another issue is that the `stable` keyword to `argsort` was introduced in *numpy* 2.0 so, if we stick with it, we'll need to update the requirements to require *numpy>=2*.

sebastien-lenard:
@mcflugen thank you for all this.

I installed this branch locally and will soon run comparison tests between the flow router component and the priority flood flow router, to check if this sort impacts performance for bigger grids.

mcflugen:
Thanks @sebastien-lenard!

FYI, I undid your merge commit and did a rebase with the *master* branch instead. I hope I didn't mess you up, but I didn't see that you had made any other changes to the branch.

PR #2129, opened 2025-02-11
Run the Landlab richdem tests as a separate job

mcflugen:
### Description

The *richdem* tests require *richdem* to be installed via *conda*. I would, however, like to switch our CI from using *setup-miniconda* to use the *setup-python* action because it's faster, and mirrors the recommended way to install landlab (if not using *richdem*, anyway). To that end, I've broken out the *richdem* tests as a separate job that uses *setup-miniconda* to allow for the rest of the jobs to be able to use *setup-python*.

* Tests and notebooks that use *richdem* are now marked with the *pytest* tag, `richdem` (they can be run with `pytest -m richdem` or not run with `pytest -m 'not richdem'`).
* The *test-landlab* and *test-notebooks* are now run with *setup-python* and I've removed the *conda* backend from their corresponding sessions in the *noxfile*.

### Checklist - did you ...

- [x] Add a [news fragment](https://landlab.csdms.io/development/contribution/#news-entries) file entry if necessary?
- [x] All tests have passed?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -284,7 +284,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12"]
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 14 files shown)


mdpiper (approved):
I think you've converted me to minimizing my use of conda.

PR #2132, opened 2025-02-12
Use arm64 runners to the CI

mcflugen:
### Description

GitHub Actions now provides *arm64* runners, so I've added them to our CI. We had been building *arm64* wheels using *QEMU* but that was sooooo slow (it took nearly an hour to build a wheel). It was so slow, if fact, that I had disabled those builds except when publishing to PyPI.

### Checklist - did you ...

- [x] Add a [news fragment](https://landlab.csdms.io/development/contribution/#news-entries) file entry if necessary?
- [x] All tests have passed?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -366,7 +366,6 @@ jobs:
       - test-notebooks
       - test-richdem
       - build-sdist
-      - build-aarch64
     name: "Publish to PyPI/TestPyPI"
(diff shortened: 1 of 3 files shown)


mdpiper:
@mcflugen This issue bit me yesterday when I was using Docker, I think. I'll do the review this morning.

mdpiper (approved):
There are many interesting (to me) changes here.

PR #2140, opened 2025-02-17
Add a quick test workflow

mcflugen:
### Description

I've added a new workflow, *quick.yml*, that is triggered by a push or pull request. This workflow only runs the tests for a single operating system (*macos-14*) and python version (3.13). The complete test suite is triggered only on a pull request and then run if it is "ready for review" (i.e. not a draft pr).

The idea here is that we don't need to run our entire battery of tests while the pr is in development because most issues can be found through the quicker tests. The full set of tests can take 45 minutes to run—that's 45 minutes of wall time. Because many of the tests are run in parallel, the actual cpu time is many hours.

I chose to run the quick tests using *macos-14* (arm) because it seemed to be the fastest.

### Checklist - did you ...

- [ ] Add a [news fragment](https://landlab.csdms.io/development/contribution/#news-entries) file entry if necessary?
- [x] All tests have passed?

### DIFF
--- a/.github/actions/run-notebooks/action.yml
+++ b/.github/actions/run-notebooks/action.yml
@@ -10,6 +10,9 @@ inputs:
   upload-name:
     required: false
     default: ''
+  opentopography-api-key:
+    required: false
+    default: ''
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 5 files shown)


mdpiper (dismissed):
I think this is a neat idea.

PR #2167, opened 2025-04-20
Typos fixed in index.md

JuanGuerrero09:
I was checking the tutorial and found those two typos, just two words, so I'm not adding all the description, checklist, etc

### DIFF
--- a/docs/source/tutorials/index.md
+++ b/docs/source/tutorials/index.md
@@ -72,7 +72,7 @@ How Landlab stores spatial data on the grid; a little on naming conventions.
 Extra credit: Go back to the [Hobley et al. 2017 publication][esurf-2017]
 and identify the ordering conventions of nodes, links, and other grid elements.
 
-### 1.4 Working with Digital Elevtion Models (DEMs)
+### 1.4 Working with Digital Elevation Models (DEMs)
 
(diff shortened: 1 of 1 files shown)


JuanGuerrero09:
I noticed that an external link refers to this page as a "Syllabus", but the current content doesn't follow a typical syllabus structure—it mainly consists of a single section titled Format and Outline. Are there any plans to restructure it into a more academic-style syllabus with clearer sections?

gregtucker:
@JuanGuerrero09 thank you for spotting and fixing this typo. I take your point about the word "syllabus". We don't have any immediate plans to develop a more traditional Landlab syllabus, but would be opening to doing so if there is sufficient interest. Perhaps you or others interested could raise this idea on the [CSDMS Forum](https://forum.csdms.io/)?

gregtucker (approved):
looks good, thanks for the typo fix

JuanGuerrero09:
> @JuanGuerrero09 thank you for spotting and fixing this typo. I take your point about the word "syllabus". We don't have any immediate plans to develop a more traditional academic syllabus, but would be opening to doing so if there is sufficient interest. Perhaps you or others interested could raise this idea on the [CSDMS Forum](https://forum.csdms.io/)?

Thank you! I'll raise the idea there, I found it interesting since a syllabus can work as a road map to learn from the start, and maybe improve the order of the topics that can be found there.

PR #2241, opened 2025-08-31
Rewrite PET Component, Add Spatial Distribution + Method Improvements

BerkM125:
## Description

The Potential Evapotranspiration field component currently relies on the use of an external instance of the Radiation component in order to obtain fields necessary to compute spatially distributed daily potential evapotranspiration/ET, making its usage more difficult for users uninvolved with the extra specifics of solar radiation, and who are focused on calculating PET. The PET component also currently uses three methods for calculating daily PET: the Priestley Taylor method, the Measured Rad PT method, and the Cosine method. Among these methods, Measured Rad PT and Cosine are not as practical as Priestley Taylor, and Priestley Taylor contains redundant steps for calculating radiation factors that are and should be handled in the Radiation component, alone. The PET component currently does not feature the use of the Penman-Monteith method, which would be useful for users whose purposes require the incorporation of crop factors in their model calculations.

This PR and update introduces the use of the Radiation component inside the PET component to eliminate the requirement of having to couple an external instance of Radiation with a user's PET code. This update maintains a user-provided solar radiation field for developers who want to use their own. This update also ensures that closed nodes are omitted from PET calculations and mapped correctly when imshow is called.

This update also introduces a new Penman-Monteith method function, and all its associated variables / inputs. The Measured Rad PT method is removed, and all unnecessary radiation-related code is removed and handled instead by the internally running Radiation component (etpRad). This also means that all radiation-related output fields are now removed from PET.

Finally, this update ensures that temperature and leaf area index can be passed as spatially distributed fields to the PET component, and be calculated as fields and not constants. Temperature and LAI can now be passed as constants, cell value fields, or even node value fields. This increased support also extends to the https://github.com/landlab/landlab/pull/1755, giving spatial distribution support to its associated temperature variables as well, without affecting anything else about Radiation.

---

## Pull Request Checklist

Please confirm the following before marking the pull request as "Ready for Review"
(if any of the following items aren't relevant for your contribution please still
tick them so we know you've gone through the checklist):

- [x] The PR is opened in **draft mode**
- [x] The PR addresses a single feature, fix, or issue
- [x] Code has been linted and is free of style issues (`nox -s lint`)
- [x] All tests pass locally (`pytest`, or `nox -s test`)
- [ ] Documentation builds without errors (`nox -s docs-build`)
- [x] A [**news fragment**](https://landlab.csdms.io/development/contribution/#news-entries)
      describing the change has been added
- [x] Relevant docstrings, examples, or changelog entries have been updated
[...]

PR #2250, opened 2025-09-05
Rewrite PET Component, Add Spatial Distribution + Method Improvements

mcflugen:
# Pull Request Guidelines

  not just *what* was changed.

---

## Pull Request Checklist

---

## Description

This is a cleaned-up version of #2241 by @BerkM125.

---

### DIFF
--- a/news/2241.feature
+++ b/news/2241.feature
@@ -1,0 +1,6 @@
+The Potential Evapotranspiration component was updated to
+internally handle solar radiation calculations.
+A new Penman-Monteith method was added, while the Measured
+Rad PT method and redundant radiation code were removed.
+Enhanced support for spatially distributed fields for
+temperature and leaf area index.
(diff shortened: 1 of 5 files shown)


BerkM125:
@mcflugen Should we update this PR to be up to date?

mcflugen:
@BerkM125 so sorry for the delay!

I have two design concerns that can be addressed together:

1. The grid copy. You introduce `self._gridCopy`, a full deep copy of the user's grid passed to the internal `Radiation` instance. It looks to me like the reason is to keep radiation fields off the user's grid, is that right? This is an expensive workaround that breaks the landlab component contract (that all component share a single grid). If the user modifies their grid between timesteps (node status, fields, etc.), the copy won't reflect those changes, possibly creating silent and difficult to find bugs.

2. Embedding `Radiation` as a component. Instantiating a full `Radiation` component internally means users who already have `Radiation` running on their grid will have two independent radiation calculations happening with potentially different parameters. It also makes it impossible to substitute a different radiation model.

Both concerns may have the same fix: don't instantiate a `Radiation` component internally at all. The radiation quantities PET needs (`Rn`, `delta`, vapor pressure, etc.) can simply be stored as instance attributes (`self._Rn`, `self._delta`, ...), which is a standard *landlab* pattern. Note that `map_node_to_cell` accepts raw arrays of node values, not just field names, so I don't think there's a need to write these quantities to the grid at any point.

What do you think?

PR #2272, opened 2025-10-08
Update the soil-creep parameter in DepthDependentLinearDiffuser

### LINKED ISSUE #2258: Correct parameterization of DepthDependentLinearDiffuser
The `DepthDependentLinearDiffuser` uses the sediment-flux law of Johnstone and Hilley (2015), which looks like:

$$
\mathbf{q}_s = -K H_* (1 - e^{-H/H_*}) \nabla \eta
$$

Here $\eta$ is ground-surface height, $H$ is soil thickness, and $H_*$ is a characteristic soil thickness parameter. The unit soil flux $\mathbf{q}_s$ should have dimensions of length squared per time. Hence $K$ is a transport coefficient with dimensions of length per time.

However, the component's documentation describes $K$ as a diffusivity, implying units of length *squared* per time. In fact, the effective diffusivity should be $KH_*$.

Either the documentation or the calculation need to be corrected. Correcting the documentation would seem to be the preferred choice to preserve backward compatibility.

However, the name of the rate coefficient parameter, `linear_diffusivity`, implies L$^2$/T even if the documentation describes it otherwise. Therefore, it would be best to change the parameter name. However, that too would break compatibility. Here's one potential solution:

1. Change the function signature from

`(self, grid, linear_diffusivity=1.0, soil_transport_decay_depth=1.0)`

to

`(self, grid, soil_creep_coefficient=None, soil_transport_decay_depth=1.0, linear_diffusivity=None)`

2. If no value is given for `soil_creep_coefficient`, but a value for `linear_diffusivity` has been passed, use the latter.

3. If a value is passed for `linear_diffusivity` (whether or not there's also one for `soil_creep_coefficient`), issue a deprecation warning.

4. If no values for either `soil_creep_coefficient` or `linear_diffusivity` are given, use the current default value of 1.0 for `soil_creep_coefficient`.

5. Correct the documentation, and add a note to the effect that `linear_diffusivity` is now deprecated.

6. In the next major release after this change has been introduced, remove `linear_diffusivity`.
 
I think this should take care of cases with no arguments given (keeps the current defaults), positional arguments given (uses the correct thing in the position previously held by `linear_diffusivity`), and keyword arguments given.

What do @nicgaspar @loroberge @mcflugen think of this as a solution?

gregtucker:
## Pull Request Checklist

---

## Description

In the `DepthDependentLinearDiffuser` component, the name of the parameter that represents the coefficient for soil creep was misleading. It was previously called `linear_diffusivity`, implying that it is the commonly used "hillslope diffusion" parameter, with dimensions of length squared per time. In fact, however, the creep equation is formulated with a coefficient that has units of length per time:

qs = K H* (1 - exp(-H / H*)) S

where qs is creep flux in L^2/T, H* is a characteristic soil depth, and S is downward-positive gradient. Therefore K has dimensions of length per time. In this formulation, the classic "geomorphic diffusivity" is K time H*. 

This PR fixes the problem as suggested in Issue #2258. The `linear_diffusivity` parameter is now deprecated in favor of a `soil_transport_velocity` parameter, and the header documentation in the component's docstring is updated to reflect the correct math. This change should not have any effect on legacy code, and should be fully backward compatible.

## Related Issues

Closes #2258

### DIFF
--- a/src/landlab/components/depth_dependent_diffusion/hillslope_depth_dependent_linear_flux.py
+++ b/src/landlab/components/depth_dependent_diffusion/hillslope_depth_dependent_linear_flux.py
@@ -218,18 +218,29 @@ class DepthDependentDiffuser(Component):
 
         # Store grid and parameters
         if linear_diffusivity is not None:
-            print(
-                deprecation_message(
-                    "Use of linear_diffusivity is deprecated, because the name is"
-                    " misleading: it is actually a velocity; diffusivity is obtained"
-                    " by multiplying by soil transport decay depth. Use"
-                    " soil_transport_velocity instead."
+            if soil_transport_velocity is not None:
+                print(
+                    deprecation_message(
+                        "Both 'linear_diffusivity' (deprecated) and"
+                        " 'soil_transport_velocity' were provided."
+                        " Ignoring 'linear_diffusivity' and using"
+                        " 'soil_transport_velocity' instead.",
+                    )
                 )
-            )
-            if soil_transport_velocity is None:
+            else:
+                print(
+                    deprecation_message(
+                        "Use of linear_diffusivity is deprecated, because the name is"
+                        " misleading: it is actually a velocity; diffusivity is obtained"
+                        " by multiplying by soil transport decay depth. Use"
+                        " soil_transport_velocity instead."
+                    )
+                )
                 soil_transport_velocity = linear_diffusivity
-        elif soil_transport_velocity is None:
+
+        if soil_transport_velocity is None:
             soil_transport_velocity = 1.0
+
         self._K = soil_transport_velocity
         self._soil_transport_decay_depth = soil_transport_decay_depth
 


src/landlab/components/depth_dependent_diffusion/hillslope_depth_dependent_linear_flux.py
```diff
@@ -190,21 +195,40 @@ class DepthDependentDiffuser(Component):
...
         ----------
         grid: ModelGrid
             Landlab ModelGrid object
-        linear_diffusivity: float
+        soil_transport_velocity : float, optional
+            Coefficient for soil creep speed, m/yr (default 1.0)
+        soil_transport_decay_depth: float, optional
+            Characteristic transport soil depth, m (default 1.0)
+        linear_diffusivity: float, DEPRECATED
             Hillslope diffusivity, m**2/yr
```
mcflugen: Let's move the DEPRECTED tag into the description.



```
        linear_diffusivity: float, optional
            **DEPRECATED**. Hillslope diffusivity, m**2/yr
```

src/landlab/components/depth_dependent_diffusion/hillslope_depth_dependent_linear_flux.py
```diff
@@ -190,21 +195,40 @@ class DepthDependentDiffuser(Component):
...
 
-        self._K = linear_diffusivity
+        # Store grid and parameters
+        if linear_diffusivity is not None:
+            message = """Use of linear_diffusivity is deprecated, because the
+                         name is misleading: it is actually a velocity;
+                         diffusivity is obtained by multiplying by soil
+                         transport decay depth. Use soil_transport_velocity
+                         instead."""
+            print(deprecation_message(message))
```
mcflugen:
```python
print(
    deprecation_message(
      "Use of linear_diffusivity is deprecated, because the"
      " name is misleading: it is actually a velocity;"
      " diffusivity is obtained by multiplying by soil"
      " transport decay depth. Use soil_transport_velocity"
      " instead."
    )
)
```

PR #2304, opened 2025-11-05
Channel network grid tools 2

keckje:
## Pull Request Checklist

---

## Description

this pull request adds the map_rmg_nodes_to_nmg_links function to the channel_network_grid_tools.py utility

map_rmg_nodes_to_nmg_links maps the nodes representing the channel location in a DEM to the closest network model grid location. Network model grid location is described in terms of link id and distance down link, measured from the inlet (tail) node of the link.

---


src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -329,9 +330,172 @@ def map_nmg_links_to_rmg_coincident_nodes(
...
+        return _dist_func(xc, row["x"], yc, row["y"])
+
+    link_ = []
+    for n in rmg_nodes:  # for each rmg node
+        xc = grid.node_x[n]
+        yc = grid.node_y[n]
+        # compute the distance to all link coincident rmg nodes
+        dist = nmg_link_to_rmg_coincident_nodes_mapper.apply(
+            lambda row: dist_between_nmg_and_rmg_nodes(row, xc, yc), axis=1
+        )
```
mcflugen, line 456: Instead of looping over all of the rows of the dataframe, I think the following, or something along these lines should be much faster,

```python
x = nmg_link_to_rmg_coincident_nodes_mapper["x"].to_numpy()
y = nmg_link_to_rmg_coincident_nodes_mapper["y"].to_numpy()

dist = np.hypot(x - xc, y - yc)
```

src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -329,9 +330,172 @@ def map_nmg_links_to_rmg_coincident_nodes(
...
+        # pick closest coincident node and corresponding link
+        # if more than one (which can happen because the confluence between two
+        # links overlay the same node), pick link with largest contributing area
+        mask = dist == dist.min()
```
mcflugen, line 460: I think you can use `np.argmin(dist)` to directly get the index of the closest node. For ties, you could use `dist == dist[idx]` and then select the nodes with the largest drainage areas.

src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -329,9 +330,172 @@ def map_nmg_links_to_rmg_coincident_nodes(
...
+        dist_min_links = nmg_link_to_rmg_coincident_nodes_mapper[
```
mcflugen, line 461: I'm not totally sure what's happening here but, I think, a new dataframe is being created every iteration, which seems like it could be quite slow. It seems like this and the rest of this loop could be converted to use just numpy arrays.

mcflugen (commented):
I have a few initial comments on the `map_rmg_nodes_to_nmg_links` function. Basically, I think we should be able to rewrite it to use numpy arrays instead of pandas dataframes. That should speed things up considerable and make it a easier to follow.

PR #2318, opened 2025-11-14
overland flow: add adjust_unstable_discharge

mcflugen:
# Pull Request Guidelines

  not just *what* was changed.

---

## Pull Request Checklist

---

## Description

I've refactored the adjustment of discharge for stability, fixing a small error (I think) and significantly speeding up the original block of code. I think it's also clearer to read.

This replaces the following code,
```python
                q_courant = q_at_link * dt_local / self._grid.dx
                water_div_4 = h_at_link / 4.0
                (positive_q,) = np.where(q_at_link > 0)
                (negative_q,) = np.where(q_at_link < 0)

                (water_logical,) = np.where(q_courant > water_div_4)
                (water_abs_logical,) = np.where(abs(q_courant) > water_div_4)

                self._if_statement_3 = np.intersect1d(positive_q, water_logical)
                self._if_statement_4 = np.intersect1d(negative_q, water_abs_logical)

                q_at_link[self._if_statement_3] = (
                    (h_at_link[self._if_statement_3] * self._grid.dx) / 5.0
                ) / dt_local

                q_at_link[self._if_statement_4] = (
                    0.0
                    - (h_at_link[self._if_statement_4] * self._grid.dx / 5.0) / dt_local
                )
```
I believe the only difference between this and my new version is when clamping is triggered. In the old version, `q` values are clamped when `abs(q) > 0.25 * h * dx / dt` and were clamped to `abs(q) = 0.2 * h * dx / dt`. In my new version, the trigger and the clamped values are the same (i.e. the trigger and the cap are both `abs(q) > 0.2 * h * dx / dt`). I think this is what we probably want. In the original, values between the cap and the trigger were unchanged.

---

## Related Issues

#2327

### DIFF
--- a/news/2318.misc
+++ b/news/2318.misc
@@ -1,0 +1,2 @@
+Refactored discharge stability adjustments in the overland flow component
+into a new *cython* function, `adjust_unstable_discharge`.
--- a/src/landlab/components/overland_flow/_calc.pyx
+++ b/src/landlab/components/overland_flow/_calc.pyx
@@ -19,6 +19,38 @@ ctypedef fused id_t:
 @cython.boundscheck(False)
 @cython.wraparound(False)
 @cython.nonecheck(False)
+def adjust_unstable_discharge(
+    cython.floating [::1] q_at_link,
+    const cython.floating [::1] h_at_link,
+    *,
+    const double dx,
+    const double dt,
+    const id_t [::1] where,
+    cython.floating [::1] out,
+):
+    cdef Py_ssize_t n_links = where.shape[0]
+    cdef Py_ssize_t i
+    cdef id_t link
+    cdef double dx_over_dt = dx / dt
+    cdef double factor_of_safety = 0.8
+    cdef double q_threshold
+    cdef double q
+    cdef double h
+
+    for i in prange(n_links, nogil=True, schedule="static"):
+        link = where[i]
+        q = q_at_link[link]
+        h = h_at_link[link]
+
+        q_threshold = 0.25 * h * dx_over_dt * factor_of_safety
+        out[link] = copysign(fmin(fabs(q), q_threshold), q)
+
+    return (<object>out).base
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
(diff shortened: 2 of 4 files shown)


gregtucker (commented):
My main comment at this point is really a question: do we know where this q limiter comes from? I re-skimmed de Almeida et al. (2012) and didn't see it. I'm wondering what the numerical rationale is. The time-step size is in theory supposed to keep keep time steps small enough that a shallow-water wave would not pass across a full cell width: dt = alpha dx / root(g hmax), where hmax is the deepest water depth and therefore the fastest wave propagation.

My guess is that this becomes insufficient when you get supercritical flow, U > root(g h). I thought that's what the old steep_slopes flag was supposed to do: prevent flow speeds above supercritical. I think there's a physical justification for that: if there are any substantial roughness elements, flows above supercritical will be prone to generating hydraulic jumps, and therefore dropping back down below critical - effectively "hugging" the critical point (Grant 1997 wrote above this). So you could argue that any "natural" situation will have U = root(g h) as an upper limit.

I think the limiter does more than this though. If you plug in dt = alpha dx / root(g hmax) into qmax,

qmax = (1/4) f alpha h root(g hmax)

At the node where, h = hmax, this means that qmax = (0.8 x 0.7 x 0.25) times h root(g h). Since h root(g h) is critical flow, this implies that velocity will be capped at a relatively small fraction of critical speed. At nodes where h < hmax, this won't be the case, and I guess the effective fraction of critical will be bigger: (0.8 x 0.7 x 0.25 x root(hmax / h)).

Before proceeding, maybe it would be good to ask @jadams15 or @nicgaspar if they remember where this limiter comes from. If it's from a paper, it would be good to cite that paper. If it was just an ad hoc attempt to improve stability, it would be good to note that.

nicgaspar:
@mcflugen Regarding the discharge limiter ... I looked in Jordan's paper and in the user manual. I did not find anything explicit about the discharge limiter. But I think it could be that the steep slopes limiter is optional and must be turned on, but the courant check always happens, if I'm interpreting things correctly.

PR #2323, opened 2025-11-17
overland flow: refactor calc_time_step

mcflugen:
# Pull Request Guidelines

  not just *what* was changed.

---

## Pull Request Checklist

---

## Description

I've modified the `calc_time_step` function so that it only considers nodes from which extend an active link, since these are the only nodes we consider in our calculations. Before, water depths for the entire grid were considered when calculating the time step, even if those nodes were outside of the drainage basin (and, possibly, uninitialized).

---

## Related Issues

#2327

### DIFF
--- a/tests/components/overland_flow/test_dealmeida_overland_flow.py
+++ b/tests/components/overland_flow/test_dealmeida_overland_flow.py
@@ -105,6 +105,13 @@ def test_calc_time_step_no_water():
 
     assert overland_flow.run_one_step(1.0) == 1.0
 
+    grid.status_at_node[:] = NodeStatus.CLOSED
+    grid.at_node["surface_water__depth"][:] = 1.0
+
+    overland_flow._update_active_nodes(clear_cache=True)
+    with pytest.raises(NoWaterError, match="no active links"):
+        overland_flow.calc_time_step()
+
 
 def test_deAlm_name(deAlm):
     assert deAlm.name == "OverlandFlow"


src/landlab/components/overland_flow/generate_overland_flow_deAlmeida.py
```diff
@@ -356,14 +363,60 @@ def calc_time_step(self):
...
+        Raises
+        ------
+        NoWaterError
+            If there is no water on active nodes (i.e., the maximum water depth is
+            zero), this exception is raised to indicate that a time step cannot be
+            determined.
         """
+        active_nodes = self._update_active_nodes()
+
+        h: NDArray = self.grid.at_node["surface_water__depth"][active_nodes]
```
gregtucker, line 383: Yay type hints! I must get into this habit too ...

src/landlab/components/overland_flow/generate_overland_flow_deAlmeida.py
```diff
@@ -356,14 +363,60 @@ def calc_time_step(self):
...
         return (
             self._alpha
-            * self._grid.dx
-            / np.sqrt(self._g * np.amax(self._grid.at_node["surface_water__depth"]))
+            * min(self._grid.dx, self._grid.dy)
+            / np.sqrt(self._g * max_water_depth)
         )
 
+    def _update_active_nodes(self, *, clear_cache: bool = False) -> NDArray[np.bool_]:
+        """Compute and optionally re-compute the active-node mask.
```
gregtucker, line 398: This is a bit confusing. What's the difference between compute and re-compute? Does it basically return a previously computed mask and only compute it if clear_cache is True?

mcflugen: Yes, that's correct. Since it is somewhat time consuming to compute the active nodes, we cache the result so that we don't re-compute every time we call `calc_time_step`, which is every iteration of the time loop. If node/link status changes though, this needs to be recomputed.

Now that I think about it, we should probably call,
```python
self._update_active_nodes(clear_cache=True)
```
at the start of `run_one_step` just in case node/link status has changed since the last time `run_one_step` was called.

PR #2329, opened 2025-11-20
Add draft tutorial notebook on linear diffusion stability

cloudydlou:
### Summary

This pull request adds a new draft tutorial notebook to the teaching section:
`docs/source/teaching/geomorphology_exercises/hillslope_notebooks/`.

The notebook demonstrates how the choice of explicit time step controls whether a hillslope diffusion simulation using the
`LinearDiffuser` remains numerically stable. By comparing stable and unstable runs side by side, the notebook provides users with intuition for diagnosing instability and selecting appropriate model parameters.

### Motivation

Numerical stability is essential for users working with the `LinearDiffuser`, as explicit schemes require the time step to remain below a well-defined limit. When this condition is violated, the model can produce non-physical results or rapidly diverge. This notebook complements the existing *“Linear diffusion exercise with Landlab”* by focusing specifically on how stability affects simulation behavior.

### Status

- [x] Created initial version of the tutorial notebook  
- [x] Implemented stable vs. unstable diffusion examples  
- [x] Added explanatory text, figures, and step-by-step comparisons  
- [x] Added cross-links to the existing hillslope diffusion notebook  
- [x] Cleaned notebook text and removed draft/WIP markers  
- [x] Ensured all cells run without errors 

### Notes

Feedback is welcome, and I am happy to revise the notebook further based on reviewer suggestions.


gregtucker:
Thanks for the contribution! It's great to have a demonstration of numerical stability and instability. I've asked @jrymart to take a look too.

The main challenge I see is that `LinearDiffuser` actually has a built-in mechanism to ensure stability by detecting when the Courant condition is exceeded, and if it is, taking multiple smaller steps. You can see this in the `run_one_step()` method in the [code](https://landlab.csdms.io/_modules/landlab/components/diffusion/diffusion.html#LinearDiffuser): dt is calculated as `_CFL_prefactor` divided by the creep coefficient, and on a raster grid, `_CFL_prefactor` is 0.15 times dx^2. The method then takes however many steps with the given "local" time step to satisfy the duration of the full "global" time step that it's been passed as an argument. So, actually this component makes it impossible (or at least difficult) to recreate instability!

Possible solutions:

1. You could copy over some of the non-component Landlab code from the Fault Scarp tutorial, and use that to illustrate instability. I'm thinking of the example that uses `calc_grad_at_link` and `calc_flux_div_at_node` to implement 2d diffusion (instead of using `LinearDiffuser`). It doesn't check the CFL condition, if you run a version of that code with too big a dt it should produce wild oscillations that get bigger over time.
 
And/or:

2. You could make this a tutorial about dynamic time-stepping, the Courant condition, and how dynamic time-stepping prevents instability (at the cost of longer run times, as your tutorial nicely points out).

It would be neat to do both, but I don't know how much time/energy you have...

Here also are a couple of very minor suggestions:
- How about listing yourself as author at the top? And maybe a date (or just a year).
- Nice link to the *Linear diffusion exercise with Landlab* tutorial - might be even better to link to online doc version: [https://landlab.csdms.io/tutorials/fault_scarp/landlab-fault-scarp.html](https://landlab.csdms.io/tutorials/fault_scarp/landlab-fault-scarp.html) so readers can see the output
- There's a \Delta x in the Background section that needs $$

Issue #2348, opened 2026-01-26
load_params should provide informative error message when file not found


gregtucker:
The Landlab function `load_params` is a wonderful way to read model parameters from a yaml-format file. However, in the common case when the user specifies a file that doesn't exist, is mis-typed, or is in the wrong folder, the function produces a cryptic error message:

`ValueError: parsing of parameter file did not produce a dict-like object`

This error could also arise if the file is present but has format problems.

It would be helpful to detect whether the problem is a file-not-found, and report when that's the case. This should be an easy fix.

If anyone wants to try their hand at contributing to Landlab, this could be a nice one to start with! If you want to address it, please add a comment below that you're working on it and expect to have a PR by such-and-such a date. Thanks!

PR #2365, opened 2026-02-13
Updated load params to have more descriptive error messages and tests

keckje:
## Pull Request Checklist

---

## Description
When a user provides load_params.py with a problematic input, a cryptic error message is returned. With this update, load_params determines why the input was problematic and returns a message that will hopefully help the user figure out how to fix it.

## Related Issues

Response to Issue #2348


gregtucker (dismissed):
Looks great to me :)

news/2347.misc
mcflugen, line 1: This should be in a separate pull request.

news/2365.bugfix
mcflugen, line 1: Please add some content

src/landlab/core/model_parameter_loader.py
```diff
@@ -37,10 +37,15 @@ def load_file_contents(file_like):
     try:
         contents = file_like.read()
     except AttributeError:  # was a str
```
mcflugen: Instead of the `try/except` here, I think it would be better simply as,
```python
if hasattr(file_like, "read"):
    return file_like.read()
```

src/landlab/core/model_parameter_loader.py
```diff
@@ -37,10 +37,15 @@ def load_file_contents(file_like):
...
+        root, ext = os.path.splitext(file_like)
         if os.path.isfile(file_like):
             with open(file_like) as fp:
                 contents = fp.read()
-        else:
+        # if file was not found but string looks like a file path string (includes an extension)
+        # throw a FileNotFoundError.
+        elif ext:
+            raise FileNotFoundError(f"The parameter file '{file_like}' was not found")
+        else:  # otherwise treat string as raw YAML
```
mcflugen: For the case of `file_like` being a `str`, let `open` raise the `FileNotFoundError`,
```python
if isinstance(file_like, str):
    try:
        with open(file_like) as fp:
            return fp.read()
    except FileNotFoundError:
        _, ext = os.path.splitext(file_like)
        if ext:
            raise
        return file_like
```
This will be the key block that addresses #2348. If the input is a `str` and looks like a file name (i.e. it has an extension), raise a `FileNotFoundError`.

src/landlab/core/model_parameter_loader.py
```diff
@@ -37,10 +37,15 @@ def load_file_contents(file_like):
...
             contents = file_like
 
     return contents
```
mcflugen: If we get this far, raise a `TypeError` that indicates `file_like` must either be a file-like object, a path to a file, or a `str`.
```python
raise TypeError(...)
```

src/landlab/core/model_parameter_loader.py
```diff
@@ -75,10 +80,25 @@ def load_params(file_like):
     (0.0, 10.0, 2.0)
     """
     contents = load_file_contents(file_like)
+    try:
+        params = yaml.load(contents, Loader=_loader)
+    except yaml.YAMLError as exc:
+        raise ValueError(f"Error parsing YAML content: {exc}") from exc
```
mcflugen: I think we can just let `yaml.load` raise an error if it can't parse the string.

src/landlab/core/model_parameter_loader.py
```diff
@@ -75,10 +80,25 @@ def load_params(file_like):
...
 
-    params = yaml.load(contents, Loader=_loader)
+    if params is None:
+        if os.path.isfile(file_like):
+            raise ValueError("File found but is empty")
+        else:
+            raise ValueError("The parameter file is empty")
```
mcflugen: I don't think we need this block. We can just let `yaml.load` raise if it can't parse, otherwise check that it parses to a `dict` (which you do below).

news/2347.misc
keckje: Sorry for that, 2347.misc can be ignored. It's something I added to my fork master a while ago. When I created this branch, that change transferred.

src/landlab/core/model_parameter_loader.py
```diff
@@ -75,10 +80,25 @@ def load_params(file_like):
...
     if not isinstance(params, dict):
-        raise ValueError("parsing of parameter file did not produce a dict-like object")
+        if os.path.isfile(file_like):
+            raise ValueError(
+                "File found but parsing did not produce a dict-like object. Ensure that your file uses 'key: value' syntax"
+            )
+        else:
+            raise ValueError(
+                f"The YAML content was parsed successfully but produced a {type(params).__name__} instead of a dictionary. Ensure that your file uses 'key: value' syntax"
+            )
```
mcflugen: Combine these errors, they're almost identical. How about,
```python
if isinstance(file_like, str) and os.path.isfile(file_like):
    src = f"file {file_like!r}"
else:
    src = "string input"

raise ValueError(
    f"Expected YAML from {src} to parse to a dict, got {type(params).__name__}."
)
```

tests/core/test_load_params.py
```diff
@@ -37,3 +51,67 @@ def test_from_yaml_path(tmpdir):
...
+            match="File found but parsing did not produce a dict-like object. Ensure that your file uses 'key: value' syntax",
+        ):
+            load_params("./params_list.txt")
+
+
+def test_yaml_string_but_is_not_a_dict():
+    yaml_string = YAML_PARAMS_LIST_STR
+    with pytest.raises(
+        ValueError,
+        match="The YAML content was parsed successfully but produced a list instead of a dictionary. Ensure that your file uses 'key: value' syntax",
```
mcflugen: You don't need to match the entire error string, just enough to check that it's raising the error it should be raising (i.e. your `ValueError` not some other `ValueError`).

USEDBY.md
```diff
@@ -8,6 +8,8 @@ Alvis, A. D., Luce, C. H., Istanbulluoglu, E., Knuth, F., Wittkopf, L., Shean, D
 
 Baran, Z. J., & Cardenas, B. T. (2025). **Modeling Lake Bonneville paleoshoreline erosion at Mars‐like rates and durations: Implications for the preservation of erosional Martian shorelines and viability as evidence for a Martian ocean.** *Journal of Geophysical Research: Planets,* 130(4), e2024JE008851.
 
+Keck, J., Istanbulluoglu, E., Strauch, R., Cristea, N. (2025). **Could climate change decrease landslide hazard in snow-dominated mountainous regions? Insights from a distributed hydrology-shallow landslide model of the North Cascades, USA.** *Water Resources Research,* 61, e2025WR040071, [https://doi.org/10.1029/2025WR040071](https://doi.org/10.1029/2025WR040071).
```
mcflugen, line 33: This should be in another pull request.

mcflugen (changes requested):
Thanks @keckje! Looks good! Just a few changes.

src/landlab/core/model_parameter_loader.py
keckje: changed as suggested

src/landlab/core/model_parameter_loader.py
keckje: Changed as suggested but added "OSError" to exception

src/landlab/core/model_parameter_loader.py
keckje: changed as suggested

src/landlab/core/model_parameter_loader.py
keckje: I left this unchanged because yaml.load throws a cryptic error message and the test that checks for a YAMLError (test_file_found_but_faulty_yaml_syntax) does not work unless the YAML error is passed as currently written

src/landlab/core/model_parameter_loader.py
keckje: I left this unchanged because it can be helpful to know if a file is empty. If this is removed, the error message from the check that params is a dict does tell the user that their input is "NoneType" but people may not know that "NoneType" means empty. But removing this is not problem too.

src/landlab/core/model_parameter_loader.py
keckje: I also left this unchanged because in some cases, a file may not have an extension (see test_read_file_no_extension). In this case, clarifying that a file was found versus just stating the file name would clarify that an external file was read and that the problem is the external file. But changing as suggested is fine too.

tests/core/test_load_params.py
keckje: These are to check that the returned type is correct (in this case list verus dict). But the "match" part of the tests can be removed if that is better, no problem

news/2365.bugfix
keckje: added content

mcflugen (approved):
👍

PR #2367, opened 2026-02-20
Fix some typos

danieldouglas92:
I noticed a typo when going through some of the documentation and then when I went to fix it I noticed it existed in other places. Then I got copilot to look for other typos and this is the culmination of what it found. I hope that all of these are correct, I went through and looked at most of the changes and they didn't seem like they would affect any actual code, but I'll guess we'll see if the testers pass!


gregtucker (dismissed):
This is great - wonderful and long-needed cleanup. I reviewed all the diffs except one large file, a mass wasting demo notebook, that was too big to diff. The vast majority are doc changes, but there are a few cases of code being changed (e.g., variable name including "reciever" changed to "receiver"). As long as all the tests pass, this should be ok.

danieldouglas92:
@gregtucker @mcflugen Thanks for taking a look at this so quickly and  can definitely separate this into two PR's! I'll keep this one as doc changes and open a second one that just changes names of variables in actual code sometime later today

PR #2383, opened 2026-03-13
Vectorized version of map_nmg_links_to_rmg_coincident_nodes that returns a dict rather than a pandas dataframe

keckje:
Speeds up map_nmg_links_to_rmg_coincident_nodes and removes pandas operations. Updated syntax of tests to compare expected values to values in a dictionary rather than a pandas dataframe.

## Pull Request Checklist

- [x] A [**news fragment**](https://landlab.csdms.io/development/contribution/#news-entries describing the change has been added

---

## Description

The previous version of map_nmg_links_to_rmg_coincident_nodes iterated over each node and returned a pandas dataframe. This version uses vector operations to process all nodes at once and returns a dict. The pandas dependency is removed from channel_network_grid_tools.py.

---

## Related Issues

 corrects PR 2264


src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -266,9 +265,11 @@ def map_nmg_links_to_rmg_coincident_nodes(
     grid, nmgrid, link_nodes, remove_duplicates=False
 ):
     """Map links of a network model grid to all coincident raster model grid
```
mcflugen, line 267: The first line of the *docstring* should be a one-line description of the function.

src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -278,60 +279,76 @@ def map_nmg_links_to_rmg_coincident_nodes(
...
     -------
 
-    nmg_link_to_rmg_coincident_nodes_mapper: pandas dataframe
-        each row of the dataframe lists the link ID, the coincident node ID, the
-        x and y coordinates and the downstream distance of the coincident node
-        and the drainage area of the link
+    nmg_link_to_rmg_coincident_nodes_mapper: dict
+        each key of the dictionary contains an np.array whose length is equal to the
+        number of coincident nodes. Keys include link ID, coincident node ID,
+        downstream distance of the coincident node, x coordinate of the coicident
```
mcflugen, line 293: Typo: "coincident" instead of "coicident".

src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -278,60 +279,76 @@ def map_nmg_links_to_rmg_coincident_nodes(
...
-    # network model grid links
-    # loop through all links in network model grid to determine raster grid cells
-    # coincident with each link and equivalent distance from upstream (tail) node
-    for linkID, lknd in enumerate(link_nodes):  # for each link in network grid
-
-        x0 = nmgrid.x_of_node[lknd[0]]  # x and y of downstream link node
+    linkIDs_list = []
+    nodes_list = []
+    Xs_list = []
+    Ys_list = []
```
mcflugen, line 300: Use snake case for variable names.
* `linkIDs_list` -> `link_ids_list`
*  `Xs_list` -> `xs_list`
*  `Ys_list` -> `ys_list`

etc.

src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -278,60 +279,76 @@ def map_nmg_links_to_rmg_coincident_nodes(
...
+            )
+        )
+
+    nmg_link_to_rmg_coincident_nodes_mapper = {
+        "linkID": np.concatenate(linkIDs_list),
+        "coincident_node": np.concatenate(nodes_list),
+        "x": np.concatenate(Xs_list),
+        "y": np.concatenate(Ys_list),
+        "coincident_node_downstream_dist": np.concatenate(dists_list),
+        "link_drainage_area": np.concatenate(link_drainage_areas_list),
```
mcflugen, line 340: You've changed these names from `"dist"` and `"drainage_area"`.  Is that ok? It seems like it could break someone's code.

src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -266,9 +265,11 @@ def map_nmg_links_to_rmg_coincident_nodes(
     grid, nmgrid, link_nodes, remove_duplicates=False
 ):
```
mcflugen, line 266: Add type annotations. Maybe something like,
```python
def map_nmg_links_to_rmg_coincident_nodes(
    grid: RasterModelGrid,
    nmgrid: NetworkModelGrid,
    link_nodes: ArrayLike,
    remove_duplicates: bool = False,
) -> dict[str, NDArray]:
```

src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -278,60 +279,76 @@ def map_nmg_links_to_rmg_coincident_nodes(
...
-        if True, when two or more links are coincident with the same node,
-        the node is assigned to the link with the largest drainage area. If False,
-        the node is assigned to each coincident link. The default is False.
+        if True, when two or more links are coincident with the same node, which
+        can occur at stream junctions, the node is assigned to the link with the
+        largest drainage area. If False, the node is assigned to each coincident
+        link. The default is False.
 
     Returns
     -------
```
mcflugen, line 289: Remove the extra blank line.

src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -278,60 +279,76 @@ def map_nmg_links_to_rmg_coincident_nodes(
     link_nodes : np array
         head and tail node of each link
     remove_duplicates : bool
```
mcflugen, line 281: Fix the return types, document all parameters. Descriptions should start with a capital letter.

* `RasterModelGrid` instead of "raster model grid", same with `NetworkModelGrid`
* `array_like` instead of "np array"
* `bool, optional` to indicate an optional parameter

src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -278,60 +279,76 @@ def map_nmg_links_to_rmg_coincident_nodes(
...
 
     """
-    Lxy = []  # list of all nodes and node attributes that coincide with the
-    # network model grid links
-    # loop through all links in network model grid to determine raster grid cells
-    # coincident with each link and equivalent distance from upstream (tail) node
-    for linkID, lknd in enumerate(link_nodes):  # for each link in network grid
-
-        x0 = nmgrid.x_of_node[lknd[0]]  # x and y of downstream link node
+    linkIDs_list = []
```
mcflugen, line 297: Validate `link_nodes` before using it. The following should work,
```python
require_id_array(
    link_nodes,
    shape=("n_links", 2),
    max_id=nmgrid.number_of_nodes - 1,
    bad_id=None,
    name="link_nodes",
)
```

mcflugen (changes requested):
Thanks @keckje! Overall, this looks great. I've requested some changes but nothing major.

src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -278,60 +279,76 @@ def map_nmg_links_to_rmg_coincident_nodes(
...
-                Lxy.append(xy)
-    df = pd.DataFrame(Lxy)
-
-    # if remove_duplicates, remove duplicate node id from link with smaller
-    # contributing area.
+        Xs, Ys, dists = _link_to_points_and_dist(
+            (x0, y0), (x1, y1), number_of_points=1000
+        )
+        dists = dists.max() - dists  # convert to distance from tail node
+        nodes = grid.find_nearest_node((Xs, Ys))
```
mcflugen, line 317: Now that there is a grid-agnostic `find_nearest_node`, you should probably use that for grids that are not `RasterModelGrids`.

src/landlab/utils/channel_network_grid_tools.py
mcflugen, line 264: After you make this grid-agnostic, this function should be renamed. I suggest something simpler like `map_network_links_to_nodes`.

mcflugen (changes requested):

src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -278,60 +279,76 @@ def map_nmg_links_to_rmg_coincident_nodes(
...
+        linkIDs_list.append((np.ones(len(nodes)) * linkID).astype(int))
+        nodes_list.append(nodes)
+        Xs_list.append(grid.node_x[nodes])
+        Ys_list.append(grid.node_y[nodes])
+        dists_list.append(dists[mask])
+        link_drainage_areas_list.append(
+            (np.ones(len(nodes)) * nmgrid.at_link["drainage_area"][linkID]).astype(
+                float
+            )
+        )
```
mcflugen, line 332: This can be expressed more simply as,
```python
link_drainage_areas.append(
    np.full(
        nodes.size,
        nmgrid.at_link["drainage_area"][link_id],
        dtype=float,
    )
)
```

src/landlab/utils/channel_network_grid_tools.py
```diff
@@ -278,60 +279,76 @@ def map_nmg_links_to_rmg_coincident_nodes(
...
+
+    nmg_link_to_rmg_coincident_nodes_mapper = {
```
mcflugen, line 334: Rename. Maybe `network_link_to_nodes`?

mcflugen (changes requested):

src/landlab/utils/channel_network_grid_tools.py
keckje: thanks, I corrected the docstrings of all functions in channel_network_grid_tools

src/landlab/utils/channel_network_grid_tools.py
keckje: corrected, thanks

src/landlab/utils/channel_network_grid_tools.py
keckje: corrected, thanks

src/landlab/utils/channel_network_grid_tools.py
keckje: I think this is fine because these are local variables, but can change if needed

src/landlab/utils/channel_network_grid_tools.py
keckje: I added type annotations to all functions in channel_network_grid_tools that lacked them

src/landlab/utils/channel_network_grid_tools.py
keckje: corrected, thanks

src/landlab/utils/channel_network_grid_tools.py
keckje: corrected, thanks

src/landlab/utils/channel_network_grid_tools.py
keckje: thanks, not sure how do do this correction....would require_id_array be a separate function that is called within map_nmg_links_to_rmg_coincident_nodes?

src/landlab/utils/channel_network_grid_tools.py
keckje: OK, I'll modify this function to use planar.find_nearest_node, thanks

src/landlab/utils/channel_network_grid_tools.py
keckje: Alright, hoping to send the updated version tomorrow morning, thanks for the review

src/landlab/utils/channel_network_grid_tools.py
mcflugen:
```python
from landlab.core.utils import require_id_array
```

keckje: Got it, thanks, I added this validation

src/landlab/utils/channel_network_grid_tools.py
keckje: The function now uses planar.find_nearest_node and passes all tests

src/landlab/utils/channel_network_grid_tools.py
keckje: I changed the name but kept coincident in the name because that is an important distinction between a mapper introduced in the next pull request

src/landlab/utils/channel_network_grid_tools.py
keckje: Thanks, changed it, I always forget about np.full

src/landlab/utils/channel_network_grid_tools.py
keckje: I changed this to "nmg_link_to_coincident_nodes_mapper"....its a little long but keeps "coincident" in the name

src/landlab/utils/channel_network_grid_tools.py
keckje: what do you think about "map_network_links_to_mg_coincident_nodes"? There are a number of mapping functions in channel_network_grid_tools like a network 1 link to network 2 link mapping function. Using a long but descriptive function title may help the user select the right function. "mg" rather than "rmg" indicates a raster or any other cellular-type grid.

mcflugen: I'm not crazy about "mg" because it's an internal shorthand that might not be clear to a user. If you think you need something, I would use "grid".

I think "coindicent nodes" could be confusing since the nodes, in general, do not lie on the links. Maybe `coincident_cells`, but you actually return the nodes associated with those cells, so that would also be confusing. Are you expecting there to be other functions that use a different rule to map links to nodes? Perhaps a base function, `map_network_links_to_grid_nodes(..., method="coincident")` that delegates to specific functions based on the method?

keckje: Alright, thanks, your original suggestion, "map_network_links_to_nodes" is probably best then. I'll modify the other functions to match this naming convention.

PR #2420, opened 2026-05-27
Fix and modernise gFlex Landlab component; add test suite

awickert:
## Summary

This PR fixes pre-existing bugs in the `gFlex` Landlab component, modernises it to the gFlex v2.x API, and completely redesigns the component around clean Landlab and CSDMS BMI principles. It also adds a comprehensive test suite. **Requires `gflex>=2.0.0`** (not yet on PyPI as a stable release — CI will fail on install until then; this is intentional).

---

### Field names

| Old | New | Reason |
|-----|-----|--------|
| `surface_load__stress` | `load__normal_component_of_stress` | Material-agnostic CSDMS-style name. The load can come from ice, water, sediment, topography, etc. — the caller converts their material-specific quantity (thickness × ρg) into a normal stress [Pa] before calling `run_one_step()`. |
| `lithosphere_surface__elevation_increment` | `lithosphere__vertical_displacement` | Matches the gFlex BMI output name. Stores the total deflection `w`, not a per-step increment (the old name was a misnomer). |

### New optional field

`lithosphere__elastic_thickness` [m] — optional node input. When present on the grid, read at every `run_one_step()` call, following the standard Landlab BMI input pattern. Allows T_e to be updated between coupling steps via `bmi.set_value()`.

### Parameter renames

| Old | New |
|-----|-----|
| `Method` | `method` |
| `BC_W`, `BC_E`, `BC_N`, `BC_S` | `bc_west`, `bc_east`, `bc_north`, `bc_south` |

### Parameters removed

- `Solver` — iterative solver dropped in gFlex v2.0; `'direct'` is the only valid value and is now the default
- `PlateSolutionType` — gFlex v2.0 removed the G2009 stencil; vWC1994 is the only implementation

### Default boundary condition changed

`zero_displacement_zero_slope` (clamped plate) → `no_outside_loads` (infinite plate approximation). The new default is more physically appropriate for most geological use cases and avoids artefacts from boundaries close to the load.

### `topographic__elevation` interaction removed

The component no longer reads or modifies `topographic__elevation`. Applying the deflection to topography is now the caller's responsibility:

```python
gf.run_one_step()
z += mg.at_node["lithosphere__vertical_displacement"] - w_prev
```

### `flex_lithosphere()` method removed

Folded into `run_one_step()`.

---

## Bug fixes

- `run_one_step()` called the non-existent method `_flex_lithosphere()`; fixed to `flex_lithosphere()` then eliminated by folding into `run_one_step()`
- `Youngs_modulus` default was `6.5e11` Pa (650 GPa — **ten times too stiff**); corrected to `65e9` Pa (65 GPa). This bug was present since the component was written.
- `topographic__elevation` declared as a required output field in `_info`; it is optional
- `initialize()` was called on every `run_one_step()`, resetting the solver state; it is now called once in `__init__`

---

## gFlex v2.x API updates

All internal attribute names updated from gFlex v1 PascalCase to v2 lowercase properties:

- `flex.Method` → `flex.method` (with `.lower()` normalisation, so `method="FD"` still works)
- `flex.Quiet` → `flex.quiet`
- `flex.BC_W/E/N/S` → `flex.bc_west/east/north/south`
- `flex.Te` → `flex.T_e`
- BC strings: `"0Displacement0Slope"` → `"zero_displacement_zero_slope"`, etc.
- `finalize()` no longer called between timesteps (gFlex v2 `finalize()` deletes `w`, `qs`, and the coefficient matrix)

---

## Improvements

- **LU factorisation caching** — `cache_factorization = True` set for scalar T_e after `initialize()`, reusing the sparse LU factorisation across `run()` calls when the load changes but T_e and the grid are fixed
- **BC validation** — delegates to `gflex.VALID_BC_STRINGS_2D`, the authoritative frozenset; new gFlex BC aliases (`clamped`, `free`, `mirror`, `infinite`, `pinned`, `no_outside_loads`) are accepted automatically
- **Inhomogeneous (dict) BCs** — dict BCs (e.g. `{"displacement": arr, "slope": arr}`) for nested-domain coupling pass through Landlab's string validation and reach gFlex directly
- **`standard_names.py` updated** — `load__normal_component_of_stress` and `lithosphere__vertical_displacement` added to the BMI standard names mapping

---

## Coupling pattern

gFlex is load-source agnostic. The caller computes the normal stress from whatever drives isostasy and writes it to the grid before each step:

```python
# Glacial isostasy example
mg.at_node["load__normal_component_of_stress"][:] = rho_ice * g * ice_thickness

gf.run_one_step()

# Apply deflection to topography (caller's responsibility)
w_new = mg.at_node["lithosphere__vertical_displacement"]
z += w_new - w_prev
w_prev = w_new.copy()
```

To vary T_e between coupling steps (e.g. from a lithospheric thermal model), create the optional field and update it before each call:

```python
mg.at_node["lithosphere__elastic_thickness"][:] = new_te
gf.run_one_step()
```

---

## Test suite (`tests/components/gflex/`)

30 tests, all skipped automatically when `gflex` is not installed (`pytest.importorskip`).

**Metadata:** component name, field names, `optional_var_names`, BC string set from `gflex.VALID_BC_STRINGS_2D`

**Input validation:** bad grid type, bad BC string, unpaired periodic BCs, invalid density contrast (`rho_fill >= rho_m`), nonexistent T_e field name, dict BC pass-through, uppercase `method` normalisation

**Physical correctness:** zero load → zero deflection; uniform load → exact isostasy (with and without non-zero `rho_fill`); sign convention (positive load → negative `w`; negative load → positive `w`); linearity; FFT solver smoke test

**Repeated calls and load changes:** idempotent under unchanged load; deflection updates when load changes

**Variable elastic thickness:** scalar, array, and field-name forms; variable T_e array differs from uniform; `lithosphere__elastic_thickness` field update (BMI pattern)

**Mirror BC:** accepted and produces correct-sign deflection

**Analytical benchmark:** point-load deflection vs. Kelvin-function solution (5 % tolerance at dx = 5 km, Te = 10 km; grid-convergence confirmed second-order)

---

## Dependency

- `gflex>=2.0.0` added as optional dependency in `pyproject.toml` (`pip install landlab[gflex]`)
- `gflex==2.0.0` pinned in `requirements/testing.txt` so CI fails visibly until the stable release is on PyPI

---

## Test plan

- [x] 30/30 tests passing locally (Python 3.12, gFlex 2.0.0b1)
- [ ] CI passes once `gflex==2.0.0` is available on PyPI

🤖 Generated with [Claude Code](https://claude.com/claude-code)


awickert:
I'm closing this PR: currently building breaking changes into `gFlex`. We'll see you for `gFlex v2.0`.

awickert:
Reopening this PR because gFlex v2.0.0 is nearly finalized and likely stable.

awickert:
Pushed two more commits tracking new additions in gFlex since the original submission:

**`f5bf3f3` — `lithosphere__elastic_thickness` optional input field**

Mirrors gFlex's new BMI `lithosphere__elastic_thickness` input variable (added upstream to allow `set_value()` updates). If the field exists on the Landlab grid it is read at each `flex_lithosphere()` call, passing it to `flex.T_e`; gFlex's property setter handles LU cache invalidation automatically when the value changes. The field is registered in `_info` as `intent='in', optional=True`.

**`db6e9d3` — dict BCs pass through BC validation**

gFlex v2.0 added inhomogeneous (prescribed-value) BCs via dicts, e.g. `{"displacement": w_arr, "slope": s_arr}`, primarily for nested-domain (coarse-to-fine) coupling. The string-membership check was rejecting these with a misleading `ValueError`. Dicts now bypass the string check and pass straight to gFlex. The BC parameter docstring has been updated to document the dict form and to list the `'pinned'`, `'infinite'`, and `'no_outside_loads'` aliases that were missing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

awickert:
Follow-up to `f5bf3f3`: fixed a silent-override bug in the T_e field update (`ab5a93a`).

The previous implementation checked for `lithosphere__elastic_thickness` unconditionally at every step, meaning any other component that happened to create that field on the grid would silently change T_e — even when the user initialized with a plain scalar or a different field name. No error, no warning, just wrong physics.

Fix: store the field name (or `None`) at construction and re-read only that field at runtime. T_e is now fixed for scalar and array initialization; it updates each step only when the user explicitly passed a field name at construction. No special-cased field names anywhere in the code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

awickert:
Corrected the T_e field approach to follow the standard Landlab BMI pattern (`741be1e`).

The prior two commits used a non-standard `self._te_field` mechanism. Looking at how Landlab's BMI bridge actually works: `set_value()` writes directly to `grid.at_node[fieldname]`, and components simply read from the grid on every `run_one_step()` call — that's the whole contract. There's no separate "dynamic" vs "one-time" mechanism.

So `lithosphere__elastic_thickness` is back in `_info` as `optional=True`, and `flex_lithosphere()` reads it via `try/except FieldError` on every call, consistent with how e.g. `OverlandFlow` reads `topographic__elevation`. A BMI coupling framework can update T_e between steps via `set_value("lithosphere__elastic_thickness", new_te)` and the component picks it up naturally.

`self._te_field` is removed — it was non-standard and unnecessary.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

awickert:
Major redesign of the component (`26dbdb1`), prompted by recognising that the original design carried too much legacy state-tracking logic.

**Breaking changes in this commit:**

- Output field `lithosphere_surface__elevation_increment` → `lithosphere__vertical_displacement` (CSDMS standard name, matching the gFlex BMI output). The field now stores the total deflection `w`, not a per-step increment — the name was previously a misnomer.
- `topographic__elevation` interaction removed entirely. Applying the deflection to topography is now the caller's responsibility (e.g. `z += mg.at_node["lithosphere__vertical_displacement"] - prev_w`). This removes an implicit, surprising side-effect from the component.
- BC parameters renamed `BC_W/E/N/S` → `bc_west/east/north/south` to match gFlex v2 API.
- Default BC changed from `zero_displacement_zero_slope` to `no_outside_loads` (alias `infinite`): the infinite-plate assumption is more physically appropriate for most geological use cases and avoids artefacts from boundary proximity.
- `flex_lithosphere()` method removed; logic folded into `run_one_step()`.
- `self._pre_flex` tracking removed (no longer needed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

mcflugen:
@awickert Thanks for updating *gFlex*! This will be great! 🎉

Please let me know when this is ready for review. Until then, will you please change the pull request to draft mode? That will limit the test matrix to a small subset of the full test suite. Also, could you rebase your branch with master?

awickert:
Draft mode: done.
Rebased to master: done.
Thanks @mcflugen!

awickert:
@mcflugen : We're ready for review! (Actually, have been, but I did not hit the button.)

PR #2427, opened 2026-06-08
Add sandsuet dump for RasterModelGrid

mcflugen:
## Pull Request Checklist

---

## Description

I've added a new function `landlab.io.sandsuet.dump` that dumps a *Landlab* `RasterModelGrid` to *sandsuet* format.

Because a `RasterModelGrid` has four different grid elements that can all be represented as rectilinear grids with uniform spacing (i.e. *nodes*, *corners*, *patches* and *cells*), the user can choose which of those locations to include in the *sandsuet* output. They cannot, however, choose a combination of elements. That is, the output cannot include, for example, values at nodes and values at cells. I don't believe *sandsuet* supports that but I could be wrong.

---

### DIFF
--- a/news/2427.feature
+++ b/news/2427.feature
@@ -1,0 +1,2 @@
+Added `sandsuet.dump` that serializes a `RasterModelGrid` and its node fields
+to *sandsuet*-compliant NetCDF format.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ dependencies = [
   "py-richdem",
   "pyyaml",
   "pyshp != 2.3.0",
-  "requireit>=0.8.0",
+  "requireit>=0.9.0",
   "rich-click",
   "scipy",
   "statsmodels",
--- a/requirements/required.txt
+++ b/requirements/required.txt
@@ -7,7 +7,7 @@ pandas==3.0.0
 py-richdem==2.2.0rc2
 pyshp==3.0.3
 pyyaml==6.0.3
-requireit==0.8.0
+requireit==0.9.0
 rich-click==1.9.7
 scipy==1.17.0
 statsmodels==0.14.6
--- a/src/landlab/io/sandsuet.py
+++ b/src/landlab/io/sandsuet.py
@@ -1,0 +1,116 @@
+from collections.abc import Iterable
+from typing import BinaryIO
+
+import xarray as xr
+from requireit import require_instance
+
+from landlab.grid.raster import RasterModelGrid
+
+
+def dump(
+    grid: RasterModelGrid,
+    *,
+    stream: BinaryIO | None = None,
+    include: str = "*",
+    exclude: str | None = None,
+) -> memoryview | None:
+    """Dump grid fields to sandsuet format.
+
+    Parameters
+    ----------
+    grid : RasterModelGrid
+        The grid whose fields will be serialized.
+    stream : BinaryIO, optional
+        A binary stream to write to. If not provided, the serialized
+        dataset is returned as a ``memoryview``.
+    include : str, optional
+        Glob-style pattern of fields to include (default is ``"*"``).
+    exclude : str, optional
+        Glob-style pattern of fields to exclude.
+
+    Returns
+    -------
+    memoryview or None
+        The serialized dataset as a ``memoryview`` if no stream is
+        provided, otherwise ``None``.
+
+    Raises
+    ------
+    ValueError
+        If the selected fields are not all at the same, supported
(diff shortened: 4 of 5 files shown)


mcflugen:
@amoodie, @ericbarefoot If you have time, could you please have a look at this and let me know what you think?

amoodie (commented):
I think the only thing missing is sandsuet 6c:

> "variable attributes shall specify variable units and a description of the variable information."

You have used SVO names for the variable names, which I think can satisfy the "description" part (though we have been putting SVO in the "long_name" variable-level attribute typically). The "units" attribute for each variable should be filled though. 

Can you get this information from the RasterModelGrid? Filing the units metadata with "none" or similar would be acceptable if there are no units.

If there is typically other information in the RasterModelGrid, you may choose to add that as a another dataset-level attribute in "description", or add a generic date and time in "history", or add a "source" as landlab. None of this is required to be sandsuet compliant though.

amoodie:
Also, will this output the history of the RasterModelGrid if the grid is evolved as part of a model simulation? That would be a natural thing to want to do from landlab, I think. Could come in a future PR too though...

mcflugen:
> Also, will this output the history of the RasterModelGrid if the grid is evolved as part of a model simulation? That would be a natural thing to want to do from landlab, I think. Could come in a future PR too though...

Definitely not. In *Landlab* fields are always overwritten as components advance in time. If a user wanted to save a history of some fields, they would have to do so themselves. Perhaps a future version of this `dump` function could append fields to an existing file so that it contained a time series.

> "variable attributes shall specify variable units and a description of the variable information."

The units stored on a Landlab field, if present, are oftentimes what a component happened to set as a default and are not guaranteed to be correct. Rather than silently writing potentially wrong units into the output file, I think `dump` should probably accept an explicit `attrs` dict (keyed by field name, following the same format as a component's `_info` dict) that the caller can use to supply accurate units and a description. If `attrs` isn't provided for a field, units could be set to `"?"`. This puts the responsibility for metadata correctness on the caller rather than trusting values that may be unreliable. The description would follow the same pattern.

For example,
```python
dump(
    grid,
    attrs={
        "topographic__elevation": {
            "units": "cm",
            "description": "height above sea level",
        }
    },
)
```

Does that seem reasonable?

amoodie:
I see. Having support to export over time seems like it would be useful to users, but time is not required for sandsuet. So, what you have definitely complies there. Maybe some examples would go a long way there.

Wrt units, all we require is that units are specified. Putting in a "?" matches the letter of the rules, but seems to skirt the spirit of the rules imo. I understand what you are saying about fields being unreliable though. An error or warning if units are not given could help. Anyway, the "?" solution would comply with sandsuet, so that would be good to there as well.

gregtucker:
I don't know if it's relevant here, but the way I often handle time-series of grids is to write a sequence of files (either NetCDF, VTK, or native Landlab .grid files) numbered like `frog000.nc`, `frog001.nc`, etc. Each individual .nc file holds all the grid fields (at least by default). ParaView is smart enough to detect that these are time-based frames, and will import them as a cluster of frames that can be run as an animation.

ericbarefoot:
> Putting in a "?" matches the letter of the rules, but seems to skirt the spirit of the rules imo. 

This could be fine actually. If we consistently have a predictable "units not found" key, then it will be easy to raise informative errors that guide the user to the right fix. 

@gregtucker, this is an interesting point. If the landlab workflow already allows for exporting individual numbered NetCDF. Is that the way most users do it? If accumulating a directory of outputs is easy, then outputting time series by appending should be just as easy. Would it be better for that change to occur not here, but rather a change somewhere else in landlab?

gregtucker:
@ericbarefoot it's not really a standard or built-in Landlab thing, just a practice I use that's also demonstrated in some of our tutorials.

ericbarefoot:
FWIW, I like the solution for requesting an attrs dir, and filling it with a clear placeholder if not supplied and then raising a warning to the user. It would be noisy, but if they are taking the time to try and dump something explicit to sandsuet, then they should probably go the next step and give the units and description. 

@mcflugen are you accepting edits to this PR directly? I could take an attempt at adding that feature.

I am still thinking about the best way to append to a netcdf. `xarray` can do this, but you do have to specify one dimension to be the unlimited "infinite" dimension, and I think you can only specify one dimension this way. Perhaps this could be a helper function?

mcflugen:
> @mcflugen are you accepting edits to this PR directly? I could take an attempt at adding that feature.

Of course, that would be great! Thanks, @ericbarefoot.

Issue #2431, opened 2026-07-01
Fix stale grid field array aliases in components


mcflugen:
Some components store a reference to a grid field array at initialization time (e.g. `self._z = self.grid.at_node["topographic__elevation"]`). If another component replaces that field in the grid, the alias silently points to stale data. The fix is to convert these to properties that re-fetch from the grid on each access.

Here's a list of some components I've found that need updating.

- [ ] `src/landlab/components/priority_flood_flow_router/priority_flood_flow_router.py`
- [ ] `src/landlab/components/tidal_flow/tidal_flow_calculator.py`
- [ ] `src/landlab/components/overland_flow/kinematic_wave_rengers.py`
- [ ] `src/landlab/components/advection/advection_solver_tvd.py`
- [ ] `src/landlab/components/groundwater/dupuit_percolator.py`
- [ ] `src/landlab/utils/fault_facet_finder.py`

attn: @ethan-pierce

PR #2437, opened 2026-07-21
Add validate_field and resolve_field utilities

mcflugen:
## Pull Request Checklist

---

## Description

As described in #2431, many *Landlab* components accept a parameter that can be a field name, or an array-like object. The component then stores it at construction time, and then need to re-resolve it against the grid on every use to ensure that a field name always reflects the field's *current* values rather than a stale snapshot.

I would like to make it easier for component authors to do this correctly.

I've added two small utilities that, I hope, make this easier and standardizes this pattern.

* `validate_field`: checks if a value is either the name of an existing field or is an array-like object that could be a field at a given grid location.
* `resolve_field`: resolves a field-like object (i.e. a field name or a *numpy* array) into a *numpy* array that reflects the current value of the field.

I'm thinking that in `__init__` components would do something like,
```python
self._value = validate_field(value, grid=self.grid, at="node")
```
and then in `run_one_step`,
```python
value = resolve_field(self._value, grid=self.grid, at="node")
```

I've also added a `FieldLike` type that components can use for typing parameters that can be either a field name or an array-like object. Note that because *numpy*'s `ArrayLike` already includes `str`, this doesn't do anything for type checkers, it's just for human readibility and documentation.

### Why `resolve_field` instead of the `return_array_at_*` functions?

* The `return_array_at_*` functions bake the grid location into the function itself, so a new location would require a nearly identical wrapper. `resolve_field`, on the other hand, takes `at` as a keyword and so covers every location with one function.
* The `return_array_at_*` functions validate input with every call whereas `resolve_field` assumes the input has already been validated with `validate_field`.
* The `return_array_at_*` functions incorrectly handle legitimate multi-dimensional per-element arrays (e.g. `(n_nodes, 2)`); `resolve_field`/`validate_field` check only the first dimension, so these pass through correctly. Patching `return_array_at_node` in place wouldn't fix the generality problem — it would still only work at `"node"`.

We could fix the last two of these issues for all of the `return_array_at_` functions but, given the first point, I think a new function is the way to go and we should deprecate the older functions in the future.

---

## Related Issues

#2431

### DIFF
--- a/src/landlab/utils/return_array.py
+++ b/src/landlab/utils/return_array.py
@@ -1,7 +1,159 @@
 #! /usr/bin/env python
 """Return array with same shape as grid elements."""
 
+import numpy as np
+from numpy.typing import ArrayLike
+from numpy.typing import NDArray
+
+from landlab.field.errors import FieldError
+from landlab.field.graph_field import GraphFields
 from landlab.utils.decorators import use_field_name_array_or_value
+
+type FieldLike = str | ArrayLike
+
+
+def validate_field(
+    value: FieldLike,
+    *,
+    grid: GraphFields | None = None,
+    at: str | None = None,
+) -> str | NDArray:
+    """Validate a field name or array-like value.
+
+    Check that *value* is either the name of an existing field, or an
+    array-like object that could represent values located on a grid.
+    Unlike :func:`resolve_field`, a field name is *not* resolved to its
+    values, so the returned value can be stored and later passed to
+    :func:`resolve_field` to pick up the field's current
+    values at the time it's needed. *value* is converted to a *numpy*
+    array, but, unlike :func:`resolve_field`, it is not broadcast.
+
+    Parameters
+    ----------
+    value : str or array_like
+        A field name, or an array of values.
+    grid : GraphFields, optional
+        Grid used to validate *value* against. If not given, *value* is
+        not checked against a grid (and, if a field name, is returned
+        unchecked).
+    at : str, optional
+        Name of the group (e.g. "node", "link") that *value* is defined
+        on. Must be given if, and only if, *grid* is given.
+
+    Returns
+    -------
+    str or ndarray
+        *value* unchanged, if it is a field name, otherwise *value* as an
+        array (not copied, if *value* is already an ndarray).
+
+    Raises
+    ------
+    ValueError
+        If only one of *grid* and *at* is given, or if *value* is
+        array-like (and not a scalar) but its first dimension does not
+        match the number of *at* elements of *grid*.
+    landlab.field.errors.FieldError
+        If *value* is a field name that does not exist in *grid* at *at*.
+
+    Examples
+    --------
+    >>> from landlab.utils.return_array import validate_field
+    >>> from landlab import RasterModelGrid
+
+    >>> grid = RasterModelGrid((3, 4))
+    >>> _ = grid.add_ones("foo", at="node")
+
+    >>> validate_field("foo", grid=grid, at="node")
+    'foo'
+    >>> validate_field(range(12), grid=grid, at="node")
+    array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11])
+    >>> validate_field(42, grid=grid, at="node")
+    array(42)
+    """
+    if (grid is None) != (at is None):
+        raise ValueError("grid and at must both be given, or neither")
+
+    if isinstance(value, str):
+        if grid is not None and value not in grid[at]:
+            raise FieldError(f"{value!r} is not a field at {at!r}")
+        return value
+
+    value = np.asarray(value)
+
+    if at is not None:
+        size = grid.size(at)
+        if value.ndim > 0 and value.shape[0] != size:
+            raise ValueError(
+                f"first dimension of value, {value.shape}, does not match the"
+                f" number of {at!r} elements of grid, {size}"
+            )
+
+    return value
+
+
+def resolve_field(
+    field: FieldLike,
+    *,
+    grid: GraphFields,
+    at: str,
+) -> NDArray:
+    """Resolve a field name or array to an array of values.
+
+    If *field* is a field name, look up and return its current values on
+    *grid*. If *field* is 0-dimensional, broadcast it to the number of
+    *at* elements of *grid*. Otherwise, return *field* unchanged. Call
+    this each time you need the values of a *field* previously validated with
+    :func:`validate_field`, so that a field name always resolves to that
+    field's current values, even if they have changed since *field* was
+    validated.
+
+    Parameters
+    ----------
+    field : str or array_like
+        A field name, or an array of values, as returned by
+        :func:`validate_field`.
+    grid : GraphFields
+        Grid to look up *field* on, or broadcast it against, as needed.
+    at : str
+        Name of the group (e.g. "node", "link") that *field* is defined
+        on.
+
+    Returns
+    -------
+    ndarray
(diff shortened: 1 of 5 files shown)


src/landlab/utils/return_array.py
```diff
@@ -1,8 +1,158 @@
...
+    Unlike :func:`resolve_field`, a field name is *not* resolved to its
+    values, so the returned value can be stored and later passed to
+    :func:`resolve_field` to pick up the field's current
+    values at the time it's needed. Although a scalar is converted to
+    a *numpy* array, it is not broadcast.
+
+    Parameters
+    ----------
+    value : str or array_like
+        A field name, scalar, or array of values.
```
gregtucker: Potentially confusing because "str or array_like" implies two options, but the text below says "scalar" is also an option. Suggest adding "or scalar". Also, I'm not familiar with "scalar" as a formal type. Maybe that's ok but could be good to a reference the first time used in this file, e.g., "scalar (see numpy.ScalarType)".

src/landlab/utils/return_array.py
```diff
@@ -1,8 +1,158 @@
...
+    >>> grid = RasterModelGrid((3, 4))
+    >>> _ = grid.add_ones("foo", at="node")
+
+    >>> validate_field("foo", grid=grid, at="node")
+    'foo'
+    >>> validate_field(range(12), grid=grid, at="node")
+    array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11])
+    >>> validate_field(42, grid=grid, at="node")
+    array(42)
+    """
```
gregtucker: If I'm understanding right, this means any scalars remain 1-element arrays rather than being broadcast into N-element arrays. Is that different from the previous behavior of return_array_at_*? I thought that function automatically broadcast scalars into N-element arrays. If that's a newly introduced change, it means a welcome reduction of memory uses for cases where the user just wants a scalar. It still leaves the problem of the user wanting either a scalar OR a subset (like core nodes) of the array, but that's a problem for another day.

src/landlab/utils/return_array.py
```diff
@@ -1,8 +1,158 @@
...
+                f"first dimension of value, {value.shape}, does not match the"
+                f" number of {at!r} elements of grid, {size}"
+            )
+
+    return value
+
+
+def resolve_field(
+    field: str | NDArray,
+    *,
```
gregtucker: Why not "FieldLike" instead of "str | NDArray"?

src/landlab/utils/return_array.py
```diff
@@ -1,8 +1,158 @@
...
+    grid: GraphFields,
+    at: str,
+) -> NDArray:
+    """Resolve a field name, array, or scalar to an array of values.
+
+    If *field* is a field name, look up and return its current values on
+    *grid*. If *field* is a scalar, broadcast it to the number of *at*
+    elements of *grid*. Otherwise, return *field* unchanged. Call this
+    each time you need the values of a *field* previously validated with
```
gregtucker: Ok, this answers my question above: broadcasting happens at the time of validation. But does this introduce an extra performance overhead as broadcasting needs to happen repeatedly?

src/landlab/utils/return_array.py
```diff
@@ -1,8 +1,158 @@
...
+
+    >>> resolve_field("foo", grid=grid, at="node") is foo
+    True
+    >>> resolve_field(range(12), grid=grid, at="node")
+    array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11])
+    >>> resolve_field(42, grid=grid, at="node")
+    array([42, 42, 42, 42, 42, 42, 42, 42, 42, 42, 42, 42])
+    """
+    if isinstance(field, str):
+        return grid.field_values(field, at=at)
```
gregtucker: Didn't realize before that we have field_values so you don't need to hard-code the "at", nice!

gregtucker (changes requested):

src/landlab/utils/return_array.py
mcflugen: Elsewhere we use "scalar" to refer to `float`, `int`, etc. However, those types are included in `ArrayLike` so maybe it would make sense to not even mention scalar here. `ArrayLike` is just anything that can be turned into a *numpy* array with `np.asarray`.

src/landlab/utils/return_array.py
mcflugen: The `validate_field` just checks that the value is valid field, it doesn't do any coercing to an array of any particular size. That's done by `resolve_field`.

src/landlab/utils/return_array.py
mcflugen: This is because `FieldLike` means either a `str` or an array-like object, not a *numpy* array. For example, `[1, 2, 3]` is array-like, but isn't a *numpy* array. `resolve_field` only accepts `str` or *numpy*-array objects, which are the types returned by `validate_field`. If `resolve_field` accepted more general array-like object (e.g.  lists of floats) every time `resolve_array` was called, it would create a new *numpy* array.

src/landlab/utils/return_array.py
mcflugen: Yes, it could introduce a slight performance hit, however it's probably quite small. When *numpy* broadcasts, for example, a scalar to a larger array, it doesn't actually create a new array filled with that scalar value. Instead, it creates a new view of the value that makes it look like a large array.

This is worth verifing, though.

mcflugen: I verified this on my laptop. The broadcast of a scalar to an array, regardless of the grid size, is about 4µs (so *O(1)*). A simple element-wise *numpy* operation on that array is *O(n)* and quickly swamps the `broadcast_to` call. Since our components do many more than a single array operation per time step, I don't think we need to worry about the overhead of the broadcast.

src/landlab/utils/return_array.py
mcflugen: I've modified the docstrings for both `validate_field` and `resolve_field` to remove references to "scalar". This is now more in line with what *numpy*'s docstrings look like (e.g. `np.asarray` and `np.broadcast_to`). There are also still *doctests* that give examples of what happens with input that is a just single `float`. Please have a look and let me know if you think it is clearer.

src/landlab/utils/return_array.py
mcflugen: I changed my mind. `resolve_field` was already passing its input through `np.asarray`, so it was handling general array-like objects (my own *doctests* even demonstrate this). So I've widened the annotation to `FieldLike` to match. If you pass something that isn't already a *numpy* array (e.g. a list of floats), which means it wasn't first passed through `validate_field`, you'll get a copy rather than a reference to your original data. I've added a note to that effect in the *docstring*.

mcflugen:
@gregtucker, I'd like to move `validate_field`, `resolve_field`, and `FieldLike` out of `landlab/utils/return_array.py` into a new `landlab/core/component_utils.py`. These are utilities to support best-pratices patterns for component authors (validate a field-name-or-array constructor parameter once, re-resolve it fresh on every step), so they fit better alongside `model_component.py` than next to `return_array_at_*`. I'd also like `component_utils.py` to become the home for future utilities aimed at component authors specifically, as they come up (I think I have another one on deck, `iter_time_step`).

The poblem I'm trying to solve is how to best make sure authors know about these, and I think having them in on well-named module, rather than scattered around, might help

What do you think?

gregtucker:
Makes sense to me, though it sounds as if the plan is evolving ... ;)

PR #2439, opened 2026-07-27
Fix issue with spatially variable soil transport velocity in DepthDependentTaylorDiffuser

### LINKED ISSUE #2426: DepthDependentTaylorDiffuser gives an error when soil transport velocity is variable
Hello Landlab team! 

I'm working with DepthDependentTaylorDiffuser in a landlab version I installed in 2024. In this version I was using the soil transport velocity K variable across my model grid (based on a vegetation field I'm currently using). Recently I installed the latest landlab version in another environment and my model was crashing whenever I tried to use an array with varying soil transport velocity values. 

I compared both versions of the component and saw that, at line 426, the latest version only allows to check for the courant factor for one value of dt_max (see screenshot), and crashes if I give an array with a non-homogeneous _K. It works well again if I change (self._dt_max < dt) to (self._dt_max < dt).any()

Image

I just wanted to point this out, but I don't know if this change was something intentional. 

Thanks a lot!!

matheuswalmeida:
# Pull Request Guidelines

  not just *what* was changed.

---

## Pull Request Checklist

---

## Description

This PR updates DepthDependentTaylorDiffuser to prevent errors when running the component with a spatially variable soil transport velocity array `K` . 

---

## Related Issues

Closes #2426

### DIFF
--- a/src/landlab/components/depth_dependent_taylor_soil_creep/hillslope_depth_dependent_taylor_flux.py
+++ b/src/landlab/components/depth_dependent_taylor_soil_creep/hillslope_depth_dependent_taylor_flux.py
@@ -317,7 +317,7 @@ class DepthDependentTaylorDiffuser(Component):
         ----------
         grid: ModelGrid
             Landlab ModelGrid object
-        linear_diffusivity: float, optional, DEPRECATED
+        linear_diffusivity: float, ndarray of float, or str, optional, DEPRECATED
             Hillslope diffusivity / decay depth, m/yr
             Default = 1.0
         slope_crit: float, optional
@@ -337,7 +337,7 @@ class DepthDependentTaylorDiffuser(Component):
             "raise", "warn")
         courant_factor : float, optional
             Courant factor for timestep calculation.
-        soil_transport_velocity : float, optional, default = 1.0
+        soil_transport_velocity : float, ndarray of float, or str, optional, default = 1.0
             Velocity parameter for soil transport, m/yr. Diffusivity is the
             product of this parameter and soil_transport_decay_depth.
         """
@@ -405,7 +405,7 @@ class DepthDependentTaylorDiffuser(Component):
             # Test for time stepping courant condition
             # Test for time stepping courant condition
             courant_slope_term = 0.0
-            courant_s_over_scrit = self._slope.max() / self._slope_crit
+            courant_s_over_scrit = self._slope / self._slope_crit
             for i in range(0, 2 * self._nterms, 2):
                 courant_slope_term += courant_s_over_scrit**i
                 if np.any(np.isinf(courant_slope_term)):
@@ -417,8 +417,10 @@ class DepthDependentTaylorDiffuser(Component):
                     raise RuntimeError(message)
             # Calculate De Max
             de_max = self._K * self._soil_transport_decay_depth * courant_slope_term
-            # Calculate longest stable timestep
-            self._dt_max = self._courant_factor * self._shortest_link**2 / de_max
+            # Calculate longest stable timestep per link
+            dt_max_per_link = self._courant_factor * self._grid.length_of_link**2 / de_max
+            # Uses the lowest dt (in case of variable K) 
+            self._dt_max = np.min(dt_max_per_link)
 
             # Test for the Courant condition and print warning if user intended
             # for it to be printed.
--- a/src/landlab/components/depth_dependent_taylor_soil_creep/hillslope_depth_dependent_taylor_flux.py
+++ b/src/landlab/components/depth_dependent_taylor_soil_creep/hillslope_depth_dependent_taylor_flux.py
@@ -418,8 +418,10 @@ class DepthDependentTaylorDiffuser(Component):
             # Calculate De Max
             de_max = self._K * self._soil_transport_decay_depth * courant_slope_term
             # Calculate longest stable timestep per link
-            dt_max_per_link = self._courant_factor * self._grid.length_of_link**2 / de_max
-            # Uses the lowest dt (in case of variable K) 
+            dt_max_per_link = (
+                self._courant_factor * self._grid.length_of_link**2 / de_max
+            )
+            # Uses the lowest dt (in case of variable K)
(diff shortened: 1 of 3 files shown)


mcflugen:
@matheuswalmeida Thank you for this contribution!

Some comments:
1. The documentation for this component says that `soil_transport_velocity` should be a float, not spatially variable. I'm not familiar enough with this component and its algorithm to know if spatially variable `k` should be allowed.

Assuming a spatially variable soild transport velocity is ok:
1. The description of `soil_transport_velocity` should be updated.
2. The calculation of the time step should probably be updated,
    ```python
    de_max = self._K * self._soil_transport_decay_depth * courant_slope_term
    # Calculate longest stable timestep
    self._dt_max = self._courant_factor * self._shortest_link**2 / de_max
    ```
    Instead of `self._shortest_link`, I imagine it should now be `self.grid.length_of_link`.
3. Please add tests that verify both that a variable `k` works and also that the solution matches what you would get with a scalar value (for example, that a soil transport velocity of `2.0` matches `np.full(grid.number_of_links, 2.0)`).

@gregtucker Do you have thoughts on this?

matheuswalmeida:
@mcflugen Thank you for the comments! 

I have done the modifications you mentioned (updated the description and time step calculation) and added a couple of unit tests to check for non homogeneous K. I'm happy to do any additional alterations if needed.

src/landlab/components/depth_dependent_taylor_soil_creep/hillslope_depth_dependent_taylor_flux.py
```diff
@@ -337,7 +337,7 @@ def __init__(
             "raise", "warn")
         courant_factor : float, optional
             Courant factor for timestep calculation.
-        soil_transport_velocity : float, optional, default = 1.0
+        soil_transport_velocity : float, ndarray of float, or str, optional, default = 1.0
```
mcflugen: Is this true? Can `soil_trasport_velocity` be a `str`? Same with `linear_diffusivity` above. It doesn't look like it to me.

src/landlab/components/depth_dependent_taylor_soil_creep/hillslope_depth_dependent_taylor_flux.py
```diff
@@ -405,7 +405,7 @@ def soilflux(self, dt):
             # Test for time stepping courant condition
             # Test for time stepping courant condition
             courant_slope_term = 0.0
-            courant_s_over_scrit = self._slope.max() / self._slope_crit
+            courant_s_over_scrit = self._slope / self._slope_crit
```
mcflugen: Why do you take the max of all the slopes here?

gregtucker:
@matheuswalmeida thanks for this PR! It's a nice capability to add. The classic linear-diffusion, regular-grid CFL condition is $\Delta t < \Delta x^2 / 2 D$. I think what's needed here is to modify this to account for both the nonlinearity in the effective $D$ (which now can also vary in space), and the possibility of spatially variable grid size, which the current code handles by looking for the shortest link (though in truth I'm not sure whether this should be link length or some other measure of grid-cell size).

src/landlab/components/depth_dependent_taylor_soil_creep/hillslope_depth_dependent_taylor_flux.py
```diff
@@ -417,13 +417,17 @@ def soilflux(self, dt):
...
+                self._courant_factor * self._grid.length_of_link**2 / de_max
+            )
+            # Uses the lowest dt (in case of variable K)
+            self._dt_max = np.min(dt_max_per_link)
 
             # Test for the Courant condition and print warning if user intended
             # for it to be printed.
             if (
-                (self._dt_max < dt)
+                (self._dt_max < dt).any()
```
mcflugen: Since you take the min of `dt_max_per_link` above, `_dt_max` is a scalar so you should have to use `any` here.

tests/components/depth_dependent_taylor_soil_creep/test_depth_dependent_taylor_diffuser.py
```diff
@@ -36,7 +36,66 @@ def test_4x7_grid_vs_analytical_solution():
...
+    assert_array_equal(
+        np.round(z[my_nodes], 1), np.array([0.0, 6.2, 10.7, 12.6, 10.7, 6.2, 0.0])
+    )
+    assert_array_equal(
+        np.round(mg.at_node["soil__depth"][8:13], 2),
+        np.array([0.35, 0.35, 0.35, 0.35, 0.35]),
+    )
+
+
+def test_4x7_grid_vs_analytical_solution_array_K():
```
mcflugen: This test looks like repeated code from the scalar k test above. If that's the case, you should have a single test that's parametrized. So something like,
```python
@pytest.mark.slow
@pytest.mark.parametrize("k", (0.01, np.full(45, 0.01)))
def test_4x7_grid_vs_analytical_solution(k):
    ...
```

tests/components/depth_dependent_taylor_soil_creep/test_depth_dependent_taylor_diffuser.py
```diff
@@ -111,6 +170,45 @@ def test_infinite_taylor_error():
         DDdiff.soilflux(10)
 
 
+def test_variable_K_matches_uniform_scalar():
```
mcflugen: Because the previous test tests a scalar and an array to the same solution, I'm not sure this test is necessary.

mcflugen (changes requested):

src/landlab/components/depth_dependent_taylor_soil_creep/hillslope_depth_dependent_taylor_flux.py
mcflugen: You can remove this line now since you're no longer using the shortest link.

matheuswalmeida:
@mcflugen thanks for the review! I believe I have addressed everything correctly now.

PR #2440, opened 2026-08-03
𝚫🏞️ GeoEnthalpyDelta Component and Tutorials for 2D version

vietbuiminh:
## Pull Request Checklist

Please confirm the following before marking the pull request as "Ready for Review"
(if any of the following items aren't relevant for your contribution please still
tick them so we know you've gone through the checklist):

- [x] The PR is opened in **draft mode**
- [x] The PR addresses a single feature, fix, or issue
- [x] Code has been linted and is free of style issues (`nox -s lint`)
- [x] All tests pass locally (`pytest`, or `nox -s test`)
- [x] Documentation builds without errors (`nox -s docs-build`)
- [x] A [**news fragment**](https://landlab.csdms.io/development/contribution/#news-entries) describing the change has been added
- [x] Relevant docstrings, examples, or changelog entries have been updated
- [x] Related issues or discussions are linked in the description (e.g., `Closes #1973`)

---

This pull request adds the new `GeoEnthalpyDelta` component to the Landlab package, making it available for import and use like other core components. The main changes involve registering the new component in the package's initialization files.

**Component registration:**

* Added import and export of the `GeoEnthalpyDelta` class in `src/landlab/components/geoenthalpy_delta/__init__.py`, and updated the `__all__` list to include it.
* Registered `GeoEnthalpyDelta` in the main `src/landlab/components/__init__.py` file by adding it to the import statements and the `__all__` list, ensuring it's accessible as part of the public API.

---

## Description

Adds `GeoEnthalpyDelta`, a new Landlab component implementing the 2D enthalpy-based delta model from Lorenzo-Trueba, Anderson, Bui, and Voller's GeoEnthalpy-Delta v1.0 manuscript (in preparation for Geoscientific Model Development).

No existing Landlab component handles the coupled subaerial/subaqueous (topset/foreset) delta problem with thresholded, slope-driven transport and mass-conserving evolution of the alluvial–bedrock transition, shoreline, and delta toe. Rather than tracking these geomorphic boundaries explicitly, the component evolves a single sediment-thickness field on a fixed grid using an explicit, positivity-preserving finite-volume update (lateral fluxes limited first, then a west-to-east downstream sweep), so the boundaries can be diagnosed afterward from the thickness and reconstructed-surface fields instead of being tracked as moving computational edges.

This version (v1.0) supports only a fixed, 2D non-erodible basement; an erodible/evolving basement is planned as a v1.1 follow-up.

Included:

[...]

PR #2441, opened 2026-08-03
✨Add a new component with pytest! GeoEnthalpyDelta

vietbuiminh:
Adds a new component, GeoEnthalpyDelta, that models 2D coupled subaerial and subaqueous delta evolution using a fixed-grid, enthalpy-based formulation with thresholded, slope-driven sediment transport. Sediment thickness is evolved on a fixed grid and the topset/foreset transport regime, mass conservation, and positivity of the deposit are covered by the accompanying tests.

Based on Lorenzo-Trueba, Anderson, Bui, and Voller, GeoEnthalpy-Delta v1.0 (manuscript in preparation for Geoscientific Model Development).

# Pull Request Guidelines

  not just *what* was changed.

---

## Pull Request Checklist

---

## Description

Adds `GeoEnthalpyDelta`, a new Landlab component implementing the 2D enthalpy-based delta model from Lorenzo-Trueba, Anderson, Bui, and Voller's GeoEnthalpy-Delta v1.0 manuscript (in preparation for Geoscientific Model Development).

No existing Landlab component handles the coupled subaerial/subaqueous (topset/foreset) delta problem with critical slope-driven transport and mass-conserving evolution of the alluvial–bedrock transition, shoreline, and delta toe. Rather than tracking these geomorphic boundaries explicitly, the component evolves a single sediment-thickness field on a fixed grid using an explicit, positivity-preserving finite-volume update (lateral fluxes limited first, then a west-to-east downstream sweep), so the boundaries can be diagnosed afterward from the thickness and reconstructed-surface fields instead of being tracked as moving computational edges.

This version (v1.0) supports only a fixed, 2D non-erodible basement; an erodible/evolving basement is planned as a v1.1 follow-up.

Included:

`landlab.components.GeoEnthalpyDelta`, wrapping the enthalpy finite-volume core around a `RasterModelGrid`.

---

## Related Issues

Following up of #2440
This PR is about the component, test, and news only.

### DIFF
--- a/src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
+++ b/src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
@@ -451,9 +451,7 @@ class GeoEnthalpyDelta(Component):
         # between calls is picked up as a shift in basement, while our
         # tracked thickness is left untouched.
         old_thickness = self._thickness.reshape(self.grid.shape)
-        elevation = self.grid.at_node["topographic__elevation"].reshape(
-            self.grid.shape
-        )
+        elevation = self.grid.at_node["topographic__elevation"].reshape(self.grid.shape)
         basement = elevation - old_thickness
         influx = self.grid.at_node["sediment__influx"].reshape(self.grid.shape)
         require_nonnegative(influx, name="sediment__influx")
--- a/src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
+++ b/src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
@@ -215,8 +215,7 @@ class GeoEnthalpyDelta(Component):
         self.initialize_output_fields()
 
         self._basement = (
-            grid.at_node["topographic__elevation"]
-            - grid.at_node["sediment__thickness"]
+            grid.at_node["topographic__elevation"] - grid.at_node["sediment__thickness"]
         )
 
         require_nonnegative(topset_threshold, name="topset_threshold")
@@ -378,9 +377,7 @@ class GeoEnthalpyDelta(Component):
             self.grid.at_node["sediment__thickness"].reshape(self.grid.shape).T.copy()
         )
         basement_xy = self._basement.reshape(self.grid.shape).T
-        influx_xy = (
-            self.grid.at_node["sediment__influx"].reshape(self.grid.shape).T
-        )
+        influx_xy = self.grid.at_node["sediment__influx"].reshape(self.grid.shape).T
 
         # Internal loop for running with user input dt
         for _ in range(n_steps):
--- a/src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
+++ b/src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
@@ -7,6 +7,9 @@ will account for an input (erodible) topography.
 import math
 
 import numpy as np
+from requireit import require_between
+from requireit import require_nonnegative
+from requireit import require_positive
 
 from landlab import Component
 from landlab import RasterModelGrid
@@ -138,10 +141,10 @@ class GeoEnthalpyDelta(Component):
         grid : RasterModelGrid
         sediment_flux : float, optional
             Total volumetric sediment input rate delivered through the
-            feeder (volume per time).
+            feeder (volume per time). Must be non-negative.
         feeder_width : float, optional
             Width, in grid-y units, of the sediment feeder positioned along
-            the grid's west (minimum-x) edge.
+            the grid's west (minimum-x) edge. Must be positive.
         feeder_y_center : float, optional
             Center, in grid-y coordinates, of the feeder. Defaults to the
             midpoint of the grid's y-extent.
@@ -152,18 +155,21 @@ class GeoEnthalpyDelta(Component):
         topset_threshold : (float, float), optional
             Critical slope thresholds, in the grid-x and grid-y directions
             respectively, above which topset (subaerial) transport occurs.
+            Must be non-negative.
         foreset_threshold : (float, float), optional
             Critical slope thresholds, in the grid-x and grid-y directions
             respectively, above which foreset (subaqueous) transport occurs.
+            Must be non-negative.
         topset_diffusivity : (float, float), optional
             Diffusivities for topset transport, in the grid-x and grid-y
-            directions respectively.
+            directions respectively. Must be positive.
         foreset_diffusivity : (float, float), optional
             Diffusivities for foreset transport, in the grid-x and grid-y
-            directions respectively.
+            directions respectively. Must be positive.
         cfl : float, optional
-            Courant-Friedrichs-Lewy stability factor (0, 1] used to pick a
-            stable time step automatically in :meth:`run_one_step`.
+            Courant-Friedrichs-Lewy stability factor used to pick a stable
+            time step automatically in :meth:`run_one_step`. Must be in
+            the interval (0, 1].
         """
         if not isinstance(grid, RasterModelGrid):
             raise TypeError(
@@ -178,6 +184,8 @@ class GeoEnthalpyDelta(Component):
 
         grid.at_node["topographic__elevation"][:] = grid.at_node["basement__elevation"]
 
+        feeder_width = require_positive(feeder_width, name="feeder_width")
+
         y_of_row = grid.y_of_node.reshape(grid.shape)[:, 0]
         if feeder_y_center is None:
             feeder_y_center = 0.5 * (y_of_row.min() + y_of_row.max())
@@ -190,14 +198,30 @@ class GeoEnthalpyDelta(Component):
                 "fall within the feeder."
             )
 
-        self._sediment_flux = sediment_flux
+        self._sediment_flux = require_nonnegative(sediment_flux, name="sediment_flux")
         self._sea_level_start = sea_level_start
         self._sea_level_rise_rate = sea_level_rise_rate
-        self._topset_threshold = topset_threshold
-        self._foreset_threshold = foreset_threshold
-        self._topset_diffusivity = topset_diffusivity
-        self._foreset_diffusivity = foreset_diffusivity
-        self._cfl = cfl
+
+        topset_threshold = require_nonnegative(
+            topset_threshold, name="topset_threshold"
+        )
+        foreset_threshold = require_nonnegative(
+            foreset_threshold, name="foreset_threshold"
+        )
+        topset_diffusivity = require_positive(
+            topset_diffusivity, name="topset_diffusivity"
+        )
+        foreset_diffusivity = require_positive(
+            foreset_diffusivity, name="foreset_diffusivity"
+        )
+        self._topset_threshold = tuple(float(v) for v in topset_threshold)
+        self._foreset_threshold = tuple(float(v) for v in foreset_threshold)
+        self._topset_diffusivity = tuple(float(v) for v in topset_diffusivity)
+        self._foreset_diffusivity = tuple(float(v) for v in foreset_diffusivity)
+
+        self._cfl = require_between(
+            cfl, 0.0, 1.0, inclusive_min=False, inclusive_max=True, name="cfl"
+        )
 
         self._time_elapsed = 0.0
 
@@ -212,7 +236,7 @@ class GeoEnthalpyDelta(Component):
 
     @sediment_flux.setter
     def sediment_flux(self, sediment_flux):
-        self._sediment_flux = sediment_flux
+        self._sediment_flux = require_nonnegative(sediment_flux, name="sediment_flux")
 
     @property
     def time_elapsed(self):
--- a/src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
+++ b/src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
@@ -364,15 +364,15 @@ class GeoEnthalpyDelta(Component):
             .reshape(self.grid.shape)
             .T.copy()
         )
-        basement_xy = self.grid.at_node["basement__elevation"].reshape(
-            self.grid.shape
-        ).T
+        basement_xy = (
+            self.grid.at_node["basement__elevation"].reshape(self.grid.shape).T
+        )
 
         # Internal loop for running with user input dt
         for _ in range(n_steps):
             H_xy = self._advance_substep(basement_xy, H_xy, substep_dt)
             self._time_elapsed += substep_dt
 
-        self.grid.at_node["topographic__elevation"][:] = (
-            self.grid.at_node["basement__elevation"] + H_xy.T.reshape(-1)
-        )
+        self.grid.at_node["topographic__elevation"][:] = self.grid.at_node[
+            "basement__elevation"
+        ] + H_xy.T.reshape(-1)
--- a/src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
+++ b/src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
@@ -32,7 +32,7 @@ class GeoEnthalpyDelta(Component):
     The land surface elevation is the sum of a fixed, non-erodible
     ``basement__elevation`` and a mobile sediment thickness ``H``:
     ``eta = eta_b + H``. Rather than tracking ``H`` as its own field, it is
-    exposed as the derived :attr:`sediment_thickness` property,
+    exposed through :meth:`calc_sediment_thickness`, which calculates
     ``topographic__elevation - basement__elevation``.
 
     Lateral (grid-y) fluxes between neighboring nodes are calculated first,
@@ -77,7 +77,7 @@ class GeoEnthalpyDelta(Component):
     >>> for _ in range(nsteps):
     ...     esd.run_one_step()  # dt chosen automatically for CFL stability
     ...
-    >>> model_volume = np.sum(esd.sediment_thickness) * grid.dx * grid.dy
+    >>> model_volume = np.sum(esd.calc_sediment_thickness()) * grid.dx * grid.dy
     >>> expected_volume = esd.sediment_flux * esd.time_elapsed
     >>> np.isclose(model_volume, expected_volume, rtol=1e-6)
     True
@@ -116,7 +116,7 @@ class GeoEnthalpyDelta(Component):
             "optional": False,
             "units": "-",
             "mapping": "node",
-            "doc": "Land surface elevation (basement__elevation + sediment thickness)",
+            "doc": "Land surface topographic elevation",
         },
     }
 
@@ -248,18 +248,26 @@ class GeoEnthalpyDelta(Component):
         """Sea level elevation at the current model time."""
         return self._sea_level_start + self._sea_level_rise_rate * self._time_elapsed
 
-    @property
-    def sediment_thickness(self):
-        """Thickness of the mobile sediment deposit at each node.
+    def calc_sediment_thickness(self):
+        """Calculate the thickness of the mobile sediment deposit at each node.
 
         Calculated as ``topographic__elevation - basement__elevation``
         rather than tracked as its own field, so it is always consistent
         with the two elevation fields even if another component modifies
-        them between calls to :meth:`run_one_step`.
+        them between calls to :meth:`run_one_step`. Clipped at zero in case
+        ``basement__elevation`` is above ``topographic__elevation``, which
+        can only happen if another component has moved one of the two
+        fields since this component last ran.
+
+        Returns
+        -------
+        ndarray
+            Sediment thickness at each node.
         """
-        return (
+        return np.maximum(
             self.grid.at_node["topographic__elevation"]
-            - self.grid.at_node["basement__elevation"]
+            - self.grid.at_node["basement__elevation"],
+            0.0,
         )
 
     def _calc_stable_time_step(self):
--- a/src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
+++ b/src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
@@ -7,6 +7,7 @@ will account for an input (erodible) topography.
 import math
 
 import numpy as np
+from requireit import ValidationError
 from requireit import require_between
 from requireit import require_nonnegative
 from requireit import require_positive
@@ -27,23 +28,27 @@ class GeoEnthalpyDelta(Component):
     This component is a structured Landlab wrapper around the core physics
     of a manuscript on 2D GeoEnthalpy-Delta modeling.
 
-    Sediment is supplied at the grid's west (minimum-x) edge according to the
-    ``sediment__influx`` field and is transported as a nonlinear,
-    slope-threshold diffusive flux. At every node the transport diffusivity and slope threshold
-    depend on whether that node is a "topset" (subaerial or shallow, ``eta >= Z``) or
-    "foreset" (below sea level, ``eta < Z``) node, where ``eta`` is the land
+    Sediment is supplied at every node according to the ``sediment__influx``
+    field (a per-node volumetric rate mixed into the local sediment
+    thickness before transport, each substep) and is transported as a
+    nonlinear, slope-threshold diffusive flux. At every node the
+    transport diffusivity and slope threshold depend on whether that
+    node is a "topset" (subaerial or shallow, ``eta >= Z``) or "foreset"
+    (below sea level, ``eta < Z``) node, where ``eta`` is the land
     surface elevation and ``Z`` is the (possibly time varying) sea level.
 
     The land surface elevation ``eta`` (``topographic__elevation``) is the
-    sum of a fixed, non-erodible basement elevation ``eta_b`` and a mobile
+    sum of a non-erodible basement elevation ``eta_b`` and a mobile
     sediment thickness ``H``: ``eta = eta_b + H``. Neither the basement nor
-    the sediment thickness is a field; both are tracked internally and
-    held fixed or updated, respectively, only by this component, so
-    another component sharing the grid can't corrupt them. The basement is
-    computed once at construction as
-    ``topographic__elevation - sediment_thickness`` (see
-    :attr:`sediment_thickness` to read the current thickness) and
-    held fixed for the lifetime of the component.
+    the sediment thickness is a field; the thickness is tracked
+    internally (see :attr:`sediment_thickness` to read it) and the
+    basement is re-derived, at the start of every :meth:`run_one_step`
+    call, as ``topographic__elevation - sediment_thickness``. Because
+    ``topographic__elevation`` is the only mutable field this component
+    exposes, another component can't corrupt the tracked thickness by
+    writing to a shared grid field; it can only shift the elevation,
+    which this component then reinterprets as a change in basement on
+    its next step.
 
     Lateral (grid-y) fluxes between neighboring nodes are calculated first,
     limited so that no node can lose more sediment than it holds. Downstream
@@ -55,6 +60,14 @@ class GeoEnthalpyDelta(Component):
     Because the transport scheme distinguishes an upstream (grid-x) and a
     cross-stream (grid-y) direction and assumes a uniform, structured grid,
     this component requires a :class:`~.RasterModelGrid`.
+
+    Every node, including those on the grid's perimeter, is part of the
+    active transport domain; no distinction is made based on
+    ``status_at_node``, except that construction raises a ``ValueError``
+    if any node is closed (``BC_NODE_IS_CLOSED``), since this component
+    would otherwise silently compute transport there anyway. No flux
+    crosses the grid's outer edges: sediment can only enter through
+    ``sediment__influx`` and never leaves the domain.
 
     Examples
     --------
@@ -84,7 +97,7 @@ class GeoEnthalpyDelta(Component):
     >>> nsteps = 50
     >>> for _ in range(nsteps):
     ...     component.run_one_step()  # dt chosen automatically for CFL stability
-    ...     component.sea_level += 0.1  # sea level rises by 0.1/step, set externally
+    ...     sea_level += 0.1  # sea level rises by 0.1/step, set externally
     ...
     >>> model_volume = np.sum(component.sediment_thickness) * grid.dx * grid.dy
     >>> expected_volume = np.sum(influx) * component.time_elapsed
@@ -155,15 +168,18 @@ class GeoEnthalpyDelta(Component):
         ``sediment__influx`` must already exist on the grid before this
         component is constructed. The component has no
         ``basement__elevation`` or ``sediment__thickness`` field: instead,
-        it tracks a fixed, non-erodible basement and the mobile sediment
-        thickness internally (see :attr:`sediment_thickness`), so
-        neither can be corrupted by another component writing to a shared
-        grid field. The basement is computed once at construction as
-        ``topographic__elevation - sediment_thickness`` and held
-        internally, never updated afterward. This means the initial
-        ``topographic__elevation`` you supply (e.g. from a DEM) should
-        already include any sediment thickness you pass in via
+        it tracks the mobile sediment thickness internally (see
+        :attr:`sediment_thickness`) and re-derives the basement, at the
+        start of every :meth:`run_one_step` call, as
+        ``topographic__elevation - sediment_thickness``. This means the
+        initial ``topographic__elevation`` you supply (e.g. from a DEM)
+        should already include any sediment thickness you pass in via
         *sediment_thickness*.
+
+        This component transports sediment at every node, including the
+        grid's perimeter, and raises ``ValueError`` if the grid has any
+        closed nodes (``status_at_node == grid.BC_NODE_IS_CLOSED``); no
+        other ``status_at_node`` distinction is honored.
 
         Sea level and sediment supply are both external forcing:
 
@@ -171,11 +187,13 @@ class GeoEnthalpyDelta(Component):
           via the :attr:`sea_level` property (or ``grid.at_grid``
           directly) between calls to :meth:`run_one_step`.
         - Sediment supply: set ``sediment__influx`` values (volume per
-          time) at nodes along the grid's west (minimum-x) edge before
-          construction. Those values define both the feeder's location
-          and its spatial distribution; their sum is the total sediment
-          supply rate. Update the field yourself between calls to
-          :meth:`run_one_step` for a time-varying or nonuniform supply.
+          time) at any node(s) before construction; a value at a given
+          node is mixed into that node's sediment thickness before
+          transport each substep. A west-edge feeder is just the special
+          case of setting influx only on the grid's west (minimum-x)
+          column. Must be non-negative and finite. Update the field
+          yourself between calls to :meth:`run_one_step` for a
+          time-varying or nonuniform supply.
 
         Parameters
         ----------
@@ -211,6 +229,15 @@ class GeoEnthalpyDelta(Component):
                 "structured (x, y) grid."
             )
 
+        if np.any(grid.status_at_node == grid.BC_NODE_IS_CLOSED):
+            raise ValueError(
+                "GeoEnthalpyDelta computes sediment transport at every "
+                "node, including the grid's perimeter, and does not honor "
+                "closed boundary nodes. The grid has one or more closed "
+                "nodes in status_at_node; open them before constructing "
+                "the component."
+            )
+
         super().__init__(grid)
 
         sediment_thickness = require_nonnegative(
@@ -219,7 +246,6 @@ class GeoEnthalpyDelta(Component):
         self._thickness = np.broadcast_to(
             sediment_thickness, (grid.number_of_nodes,)
         ).astype(float)
-        self._basement = grid.at_node["topographic__elevation"] - self._thickness
 
         require_nonnegative(topset_threshold, name="topset_threshold")
         require_nonnegative(foreset_threshold, name="foreset_threshold")
@@ -246,15 +272,11 @@ class GeoEnthalpyDelta(Component):
     def sea_level(self):
         """Sea level elevation, read from the ``sea_level__elevation`` grid field.
 
-        This is external forcing: update ``grid.at_grid["sea_level__elevation"]``
-        (or set this property) between calls to :meth:`run_one_step` if you
-        want sea level to vary with time.
+        This is external forcing owned by the caller: update
+        ``grid.at_grid["sea_level__elevation"]`` directly between calls to
+        :meth:`run_one_step` if you want sea level to vary with time.
         """
         return self.grid.at_grid["sea_level__elevation"]
-
-    @sea_level.setter
-    def sea_level(self, sea_level):
-        self.grid.at_grid["sea_level__elevation"] = float(sea_level)
 
     @property
     def sediment_thickness(self):
@@ -263,7 +285,9 @@ class GeoEnthalpyDelta(Component):
         Tracked internally rather than as a field, so it can't be
         corrupted by another component writing to a shared grid field.
         """
-        return self._thickness
+        thickness = self._thickness.view()
+        thickness.flags.writeable = False
+        return thickness
 
     def _calc_stable_time_step(self):
         """Calculate a Courant-Friedrichs-Lewy limited stable time step.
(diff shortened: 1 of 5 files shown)


src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,402 @@
...
+    Cfore_x,
+    Cfore_y,
+    Dtop_x,
+    Dtop_y,
+    Dfore_x,
+    Dfore_y,
+    dx,
+    dy,
+    dt,
+):
```
mcflugen: This function has too many parameters. I think it would probably make sense to turn this into a method of `GeoEnthalpyDelta` so that it can read data from the component. Maybe something like,
```python
def _advance_substep(self, basement, thickness, dt):
    ...
```

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,402 @@
...
+    nx, ny = basement.shape
+    eta = basement + H
+    Hnew = np.zeros((nx, ny))
+    qx = np.zeros((nx + 1, ny))
+    qy = np.zeros((nx, ny + 1))
+    fac_y = np.ones((nx, ny))
+
+    for j in range(ny):
+        if feeder_mask[j]:
+            qx[0, j] = qdens
```
mcflugen: All of the loops in this function should be vectorized using numpy array operations. For example, this loop could be written more simply as, `qx[0, feeder_mask] = qdens`.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,402 @@
...
+                f = (H[i, j] / dt) / (out_rate + tiny)
+                if f < 1.0:
+                    fac_y[i, j] = max(f, 0.0)
+    for i in range(nx):
+        for j in range(ny - 1):
+            face = qy[i, j + 1]
+            qy[i, j + 1] = fac_y[i, j] * face if face >= 0.0 else fac_y[i, j + 1] * face
+
+    # Sequential downstream limiter, including incoming and lateral flux.
+    for i in range(nx - 1):
```
mcflugen: I'm not sure this loop can be fully vectorized but I think the inner loop could be.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,402 @@
...
+    depend on whether that node is a "topset" (subaerial or shallow, ``eta >= Z``) or
+    "foreset" (below sea level, ``eta < Z``) node, where ``eta`` is the land
+    surface elevation and ``Z`` is the (possibly time varying) sea level.
+
+    The land surface elevation is the sum of a fixed, non-erodible
+    ``basement__elevation`` and a mobile ``sediment__thickness``:
+
+    .. math::
+
+       \eta = \eta_b + H
```
mcflugen: Don't bother with using TeX here. `eta = eta_b + H` is easier to read.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,402 @@
...
+        topset_threshold_x=0.0,
+        topset_threshold_y=0.0,
+        foreset_threshold_x=2.0,
+        foreset_threshold_y=2.0,
+        topset_diffusivity_x=1.0,
+        topset_diffusivity_y=1.0,
+        foreset_diffusivity_x=1.0,
+        foreset_diffusivity_y=1.0,
+        cfl=0.4,
+    ):
```
mcflugen, line 178: There are too many parameters here. We can begin with condensing the directional pairs into tuples. For examples,
```python
        topset_threshold=(0.0, 0.0),
        foreset_threshold=(2.0, 2.0),
        topset_diffusivity=(1.0, 1.0),
        foreset_diffusivity=(1.0, 1.0),
```
If any of these are given a scalar, assume that value for both directions. Internally, these should all be normalized to tuples.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,402 @@
...
+
+        super().__init__(grid)
+
+        self.initialize_output_fields()
+        self.initialize_optional_output_fields()
+
+        self._basement = grid.at_node["basement__elevation"]
+        self._thickness = grid.at_node["sediment__thickness"]
+        self._topo = grid.at_node["topographic__elevation"]
+        self._topo[:] = self._basement + self._thickness
```
mcflugen: The component shouldn't retain array aliases for grid fields. Another component can replace a grid field array after this component is instantiated, leaving those references stale. You should retrieve fields from `self.grid.at_node` whenever they are needed.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,402 @@
...
+
+        self._nrows, self._ncols = grid.shape
+        self._dx = float(grid.dx)
+        self._dy = float(grid.dy)
```
mcflugen: Keep the number of attributes to a minimum. These ones, in particular, seem unnecessary.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,402 @@
...
+        self.sea_level_rise_rate = sea_level_rise_rate
+        self.topset_threshold_x = topset_threshold_x
+        self.topset_threshold_y = topset_threshold_y
+        self.foreset_threshold_x = foreset_threshold_x
+        self.foreset_threshold_y = foreset_threshold_y
+        self.topset_diffusivity_x = topset_diffusivity_x
+        self.topset_diffusivity_y = topset_diffusivity_y
+        self.foreset_diffusivity_x = foreset_diffusivity_x
+        self.foreset_diffusivity_y = foreset_diffusivity_y
+        self.cfl = cfl
```
mcflugen: Do these all need to be public? If so, they should be properties, but I think they should probably be private for now.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,402 @@
...
+    def time_elapsed(self):
+        """Cumulative model time advanced by :meth:`run_one_step`."""
+        return self._time_elapsed
+
+    @property
+    def sea_level(self):
+        """Sea level elevation at the current model time."""
+        return self.sea_level_start + self.sea_level_rise_rate * self._time_elapsed
+
+    def calc_stable_time_step(self):
```
mcflugen: I think this method should be private and should be called by `run_one_step`. It shouldn't be up to the user to remember to call this to get a stable time step.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,402 @@
...
+            Dfore_x=self.foreset_diffusivity_x,
+            Dfore_y=self.foreset_diffusivity_y,
+            dx=self._dx,
+            dy=self._dy,
+            dt=dt,
+        )
+
+        self._thickness[:] = Hnew_xy.T.reshape(-1)
+        self._topo[:] = self._basement + self._thickness
+        self._time_elapsed += dt
```
mcflugen: This should all be wrapped in a time loop that advances the model by `dt` using stable time steps. So something like,
```python
    stable_dt = self._calc_stable_time_step()
    n_steps = math.ceil(dt / stable_dt)
    substep_dt = dt / n_steps

    for _ in range(n_steps):
        ...
```

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
mcflugen: Components should modify fields, even if they're output fields, when they're instantiated. Fields should only be modified as they are advancing in time.

You should only use two of these three as input, with the third being a calculated output. I suggest using `basement__elevation` and `topographic__elevation`, and calculating `sediment__thickness` from those two. In that case, the sediment thickness probably shouldn't be a field. Doing it this was means you won't have to ensure the three arrays are always in sync and your component will be compatible with more components.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,402 @@
...
+        d_max = max(
+            self.topset_diffusivity_x,
+            self.topset_diffusivity_y,
+            self.foreset_diffusivity_x,
+            self.foreset_diffusivity_y,
+            1.0,
+        )
+        return self.cfl / (2.0 * d_max * (1.0 / self._dx**2 + 1.0 / self._dy**2))
+
+    def run_one_step(self, dt, feeder_mask=None):
```
mcflugen: I would reconsider the `feeder_mask` override here. An update method shouldn't silently replace lasting configuration.

mcflugen (changes requested):
Thanks for contributing this component! The underlying model looks valuable, but I think it still needs a bit or work before we merge it.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: Since i still want people to be able to modified or change sediment_flux during a run, I will keep it mutable, but as a property with a setter rather than a bare public attribute: `self._sediment_flux` backing a `@property sediment_flux / @sediment_flux.setter` since `_advance_substep` already rereads `self.sediment_flux` for every substep.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: Okay, that make more sense, in the code I actually use that function inside the `run_one_step()` function. Originally, my intention was to move all the complex cell calculations into that `_enthalpy_diffusion_step` and call that function inside `run_on_step()`, but now I will move that `_advance_substep()` inside the Component because that is where it belongs too.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: - `qy[:, 1:-1] = np.where(face >= 0.0, fac_y[:, :-1] * face, fac_y[:, 1:] * face)`
- `for i in range(nx-1)` loop is too complicated to vectorized because of the candidate cell calculation, lateral_net and available calculations so I will have to keep it as is.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: okay maybe if the user doesn't declare `dt` by themself, make `dt = stable_dt` to be automatically decided by that function, but if the user pass in `dt`, then a condition is need to check and compare, and let them know if their own `dt` is appropriate, because they can still preceed the running but with your suggested internal loop
```python
stable_dt = self._calc_stable_time_step()
if dt is None:
    dt = stable_dt
n_steps = math.ceil(dt / stable_dt)
substep_dt = dt / n_steps

...

for _ in range(n_steps):
    H_xy = self._advance_substep(basement_xy, H_xy, substep_dt)
    self._time_elapsed += substep_dt

```

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: okay now i turned it as property return function
```python
@property
def sediment_thickness(self):
    """Thickness of the mobile sediment deposit at each node.

    Calculated as ``topographic__elevation - basement__elevation``
    rather than tracked as its own field, so it is always consistent
    with the two elevation fields even if another component modifies
    them between calls to :meth:`run_one_step`.
    """
    return (
        self.grid.at_node["topographic__elevation"]
        - self.grid.at_node["basement__elevation"]
    )
```

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
mcflugen: The `dt` that is passed to `run_one_step` is a duration that the component would run for. That duration should be subdivided into stable time steps that are determined by `calc_stable_time_step`.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
mcflugen: While we're at it, all of the inputs should be validated. I'm not sure what that means here but maybe just that they're non-negative?

mcflugen: Perhaps you could do something like the following,
```python
from requireit import require_positive
```
and then,
```python
self._topset_threshold = require_positive(topset_threshold, name="topset_threshold")
```
Remember, though, that you would still have to normalize `self._topset_threshold` to be a tuple of float.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
mcflugen: Since this is doing a calculation, what do you think about turning it into a method rather than a property? If it's a property, someone might not realize that it's actually doing a calculation and use it freely thinking it's just an attribute access. Maybe `def calc_sediment_thickness(self)`?

Is there any need to clamp the result with a lower bound of zero in case the basement happens to be higher than the topography?

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: ope, i saw your comment before this commit. I will address that in the next commit.

vietbuiminh: Please refer to this commit [b319ea](https://github.com/landlab/landlab/pull/2441/changes/b319ea82d5ffe97def170043237861ce75384b80) for that implementation

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: Again, I saw this before my commit so please refer to this next commit [f6f85db](https://github.com/landlab/landlab/pull/2441/changes/f6f85db15c282858c0c12ff32ae328c5a043ddda)

vietbuiminh (commented):
Compress all of the commits that is toward the Address the comments

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    def __init__(
+        self,
+        grid,
+        sediment_flux=0.5,
+        feeder_width=1.0,
+        feeder_y_center=None,
+        sea_level_start=0.0,
+        sea_level_rise_rate=0.0,
```
mcflugen: I think you should remove these parameters. External forcing should be handled through fields. Have a look at the `CarbonateProducer` for an example.

You could add the following to `_info`,
```python
"sea_level__elevation": {
    "dtype": float,
    "intent": "in",
    "optional": False,
    "units": "-",
    "mapping": "grid",
    "doc": "Sea level elevation",
}
```
and then, if you like, the *sea_level* property could just return,
```python
return self.grid.at_grid["sea_level__elevation"]
```

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,410 @@
...
+            "dtype": float,
+            "intent": "in",
+            "optional": False,
+            "units": "-",
+            "mapping": "node",
+            "doc": "Fixed, non-erodible basement elevation",
+        },
+        "topographic__elevation": {
+            "dtype": float,
+            "intent": "out",
```
mcflugen: This should be *inout*.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,410 @@
...
+        topset_diffusivity = require_positive(
+            topset_diffusivity, name="topset_diffusivity"
+        )
+        foreset_diffusivity = require_positive(
+            foreset_diffusivity, name="foreset_diffusivity"
+        )
+        self._topset_threshold = tuple(float(v) for v in topset_threshold)
+        self._foreset_threshold = tuple(float(v) for v in foreset_threshold)
+        self._topset_diffusivity = tuple(float(v) for v in topset_diffusivity)
+        self._foreset_diffusivity = tuple(float(v) for v in foreset_diffusivity)
```
mcflugen: I think you can tidy this up more. How about a helper function that normalizes the x/y components,
```python
def _normalize_as_xy(value):
    return tuple(np.broadcast_to(value, (2,)))
```
That will broadcast scalars to tuples and the validation/normalization would look like,
```python
require_nonnegative(topset_threshold, name="topset_threshold")
self._topset_threshold = _normalize_as_xy(topset_threshold)
```

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,410 @@
...
+            "units": "-",
+            "mapping": "node",
+            "doc": "Land surface topographic elevation",
+        },
+    }
+
+    def __init__(
+        self,
+        grid,
+        sediment_flux=0.5,
```
mcflugen: Because this is external forcing, I would like to see this moved into a field.

Perhaps the sediment flux could be a field. Other components have something like the following for a input field,
```python
"sediment__influx": {
    "dtype": float,
    "intent": "in",
    "optional": False,
    "units": "-",
    "mapping": "node",
    "doc": "Volumetric sediment influx at each node",
}
```
This would remove `sediment_flux`, `feeder_width`,  and `feeder_y_center`. The values of the *sediment__influx* field would represent volumetric sediment supply per node. Values on the west boundary would define both the feeder location and its spatial distribution, while their sum gives total supply. This simplifies the constructor and naturally supports externally varying or nonuniform sediment forcing.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,410 @@
...
+        float
+            A time step that satisfies the explicit stability criterion for
+            the current diffusivities and grid spacing.
+        """
+        d_max = max(*self._topset_diffusivity, *self._foreset_diffusivity, 1.0)
+        return self._cfl / (
+            2.0 * d_max * (1.0 / self.grid.dx**2 + 1.0 / self.grid.dy**2)
+        )
+
+    def _advance_substep(self, basement, thickness, dt):
```
mcflugen: It seems like much of the `_advance_substep` method reimplements *Landlab*’s link-gradient and node-divergence operations using transposed raster arrays.

Could you represent internal sediment transport as flux at *links*, using `calc_grad_at_link` to calculate slopes and `calc_flux_div_at_node` to update thickness. A donor-based limiter could operate on outgoing links at each node, while `sediment__influx` remains a node source.

I think this would remove most manual raster indexing, naturally honor active links and control-volume geometry, and provide a path toward non-raster grids.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,410 @@
...
+                "GeoEnthalpyDelta requires a RasterModelGrid because "
+                "its explicit finite-volume scheme assumes a uniform, "
+                "structured (x, y) grid."
+            )
+
+        super().__init__(grid)
+
+        self.initialize_output_fields()
+
+        grid.at_node["topographic__elevation"][:] = grid.at_node["basement__elevation"]
```
mcflugen: Components shouldn't modify input fields when they're instantiated. This would be very surprising to another component that had, for example, just read in its elevations from a DEM.

I doesn't look to me like the component requires an initial sediment thickness of zero. Is that correct?

mcflugen (changes requested):
Thank you for making those changes, it's getting close. I have a few more comments.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: is it because of the sediment thickness that goes ontop of the basement elevation that makes this inout?

mcflugen: Within `run_one_step`, your component calculates a new sediment thickness and then adjusts the elevation field accordingly. This mean elevation's not just input, but also output. It's input because your initial sediment thickness is the difference between the surface and basement elevations.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: okay, and since we are at this, with this component, I am intent to have the basement_elevation to also be flexible in away that people input DEM into the basement_elevation, and since the topo doesn't have the delta sediment yet so it is getting the same input as basement_elevation

vietbuiminh: I will leave that step for the user to initialize them at the beginning but would raise Error if the input from both are not equal to each other

mcflugen: In the case you describe, the user should set the basement elevation to be the same as the surface elevation. It's up to the caller of this component to do this, though.

The problem before was that your component's state had three fields: basement elevation, surface elevation, and sediment thickness. The issue was that any one of those fields could be calculated from the other two and so the state was overdetermined. To fix this, the component should only use two of those fields. I thought it would be most natural to use basement elevation and surface elevation. The sediment thickness could then be calculated from those two.

vietbuiminh: my usual flow for this component when i was written it in julia is to have in for basement_elevation and out for sediment_thickness. The topographic_elevation or surface_elevation the user can simply add basement_elevation together with sediment_thickness. Would that be better if i go that route?

mcflugen: > I will leave that step for the user to initialize them at the beginning but would raise Error if the input from both are not equal to each other

If you're going to require the initial sediment thickness is zero, then we should do something else. In this case, I suggest dropping the basement elevation as a field entirely. Instead, surface elevation is an input/output field and sediment thickness is an output field.

vietbuiminh: > > I will leave that step for the user to initialize them at the beginning but would raise Error if the input from both are not equal to each other
> 
> If you're going to require the initial sediment thickness is zero, then we should do something else. In this case, I suggest dropping the basement elevation as a field entirely. Instead, surface elevation is an input/output field and sediment thickness is an output field.

Shall i change the name from topographic__elevation to surface__elevation to differentiate this?

mcflugen: No, I don't think so. *Landlab* components use the field called *topographic__elevation* to mean surface elevation. To be compatible with other components, I think you should stick to that name.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: great idea, i checked out those component and i think it would streamline the sediment calculation. I will try to implement that.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
mcflugen: How about this. The component will have two fields: surface elevation and sediment thickness. For surface elevation, the field should be called *topographic__elevation* and be input/output. I'm not sure what the best name for the sediment thickness should be, though. If you could please have a look at the list of all [*Landlab* fields](https://landlab.csdms.io/user_guide/field_definitions.html), maybe you could find something there that's already used by another component and matches what your component needs.

Is there any reason not to have the sediment thickness be both input and output? That is, does the initial sediment thickness have to be zero? I could be wrong, but it doesn't seem like having a non-zero (but positive) initial thickness would be a problem.

mcflugen: Another option. The component only uses a *topographic__elevation* field, which is input/output, and neither the basement elevation nor sediment thickness are fields. The component will still keep track of them, of course, they just won't be fields. Instead, they can be supplied to a user through class properties or methods.

I think I now like this solution the best as it's simpler and we can always expose more arrays as field in the future. Do you think this could work for you?

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: this one is tricky for me. i don't think switching to `calc_grad_at_link/calc_flux_div_at_node` would buys much speed, we already vectorized, and link-indexed math would do the same elementwise work, just addressed by link id instead of [i,j]. 
The downstream x-sweep is the one part that's sequential, but that's because each column's available flux depends on the resolved value from the column west of it (donor-limiter physics as explain from Jorge et al GMD manuscript, not the raster indexing, so it'd still need a loop even with links)
However, right now my raster code doesn't look at `status_at_node` at all, so if someone closes nodes it'd just silently ignore that. Maybe i'd rather not do is rewrite the whole donor-cell logic in link form, since eta[:, :-1] - eta[:, 1:] reads as "node i vs the node east of it" to basically anyone.
Given this is v1.0 raster-only anyway, i'd rather keep the explicit indexing and just add proper handling for inactive/boundary nodes on top of it.

does that seem reasonable, or is there something about the link based approach i'm missing that would matter for implement?

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: Given this change in this new field, shall i add this detail information if the user want to figure out the basement__elevation or just leave it and people can figure out if they want to check for the change in elevation? (I am asking this because the next iteration would make the underlying topography change (with the GIA input of elevation change at each node). 
What i have inside the docstring for initialize the component:

---
The component has no
        ``basement__elevation`` field: a fixed, non-erodible basement is
        computed once at construction as
        ``topographic__elevation - sediment__thickness`` and held
        internally, never updated afterward. This means the initial
        ``topographic__elevation`` you supply (e.g. from a DEM) is only
        ever the basement if ``sediment__thickness`` is zero; if you seed
        a nonzero initial thickness, ``topographic__elevation`` should
        already include it.

vietbuiminh: I think i totally missed your message right above the one i commented about the docstring. i'd actually like to stick with having `sediment__thickness` as a real field rather than going fully private. the main reason is other components (plotting/export code) can just read `grid.at_node["sediment__thickness"]` directly without needing to know anything about `GeoEnthalpyDelta` specifically.

vietbuiminh: but if you think that by having the sediment__thickness not as a field and it would speed up the model run, then i could redo this and have it in the next commit.

mcflugen: I do not think it would speed things up at all. What it would do, though, is simplify the API and make the component less fragile when coupling with other components. If it's a field, there's a greater risk that another component could change it in a way that is unexpected and break your component.

I don't think it would make plotting any more difficult. The following should still work,
```python
grid.imshow(component.sediment_thickness, at="node")
```
Writing output I think might be a bit more difficult. You would have to temporarily add the sediment thickness as a field before writing to a file.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
mcflugen: There are versions of `calc_grad_at_link` and `calc_flux_div_at_node` that are optimized for raster grids. They pretty much do exactly the type of indexing that you do here so the speed should be pretty similar. They, however, are well tested, idiomatic *Landlab*, a useful first step to support non-raster grids, and, as you say, deal with boundary conditions. Grid generalization would still require reconsidering the directional, sequential flux limiter.

We can put that larger rewrite on hold for now. In the meantime, please factor the distinct stages of this function into smaller helper methods (e.g. calculating lateral flux, limiting lateral flux, calculating downstream flux, and applying the conservative update). This should make the algorithm easier to read and test. As it is now, it can effectively be tested only end to end.

vietbuiminh: I turned that into `def _calc_lateral_flux():`, `_limit_lateral_flux():`, and `_calc_downstream_flux()` so inside that _advance_substep() it just contains 
```python
eta = basement + thickness
qy = self._calc_lateral_flux(eta)
qy = self._limit_lateral_flux(qy, thickness, dt)
qx = self._calc_downstream_flux(eta, thickness, qy, influx, dt)
```

vietbuiminh (commented):
Compact all the changes from the review suggestions

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,423 @@
...
+
+        This is external forcing: update ``grid.at_grid["sea_level__elevation"]``
+        (or set this property) between calls to :meth:`run_one_step` if you
+        want sea level to vary with time.
+        """
+        return self.grid.at_grid["sea_level__elevation"]
+
+    @sea_level.setter
+    def sea_level(self, sea_level):
+        self.grid.at_grid["sea_level__elevation"] = float(sea_level)
```
mcflugen: Because `sea_level__elevation` is declared as an input field and represents external forcing, this component shouldn't expose a setter that modifies it.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,423 @@
...
+
+    @property
+    def sediment_thickness(self):
```
mcflugen, line 291: This property returns the component’s ***mutable*** internal array, so callers can modify sediment thickness without validation. Please expose a read-only view instead.

I think the following should work,
```python
thickness = self._thickness.view()
thickness.flags.writeable = False
return thickness
```

Related to this: in `run_one_step`, `_thickness` should be updated in-place during each step. This ensures that previously returned read-only views continue to reflect the component’s current state.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,423 @@
...
+
+        Parameters
+        ----------
+        dt : float, optional
+            Time step duration. If not given, a single CFL-stable substep
+            is taken.
+        """
+        stable_dt = self._calc_stable_time_step()
+        if dt is None:
+            dt = stable_dt
```
mcflugen, line 450: A user-supplied `dt` needs validation here. It would be positive and finite. Maybe,
```python
require_between(dt, 0.0, np.inf, inclusive_min=False, inclusive_max=False, name="dt")
```

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,423 @@
...
+        basement_xy = self._basement.reshape(self.grid.shape).T
+        influx_xy = self.grid.at_node["sediment__influx"].reshape(self.grid.shape).T
+
+        # Internal loop for running with user input dt
+        for _ in range(n_steps):
+            H_xy = self._advance_substep(basement_xy, H_xy, influx_xy, substep_dt)
+            self._time_elapsed += substep_dt
+
+        self._thickness = H_xy.T.reshape(-1)
+        self.grid.at_node["topographic__elevation"][:] = (
```
mcflugen: The `topographic__elevation` field is in/out, but you reset it here each time step. Instead, you should read in the new elevations, calculate the new basement, calculate the change in thickness and then update the elevations.

I think something like the following could work,
```python
old_thickness = self._thickness.reshape(self.grid.shape)
elevation = self.grid.at_node["topographic__elevation"].reshape(self.grid.shape)

basement = elevation - old_thickness
new_thickness = old_thickness.copy()
```
The time loop would then update `new_thickness`, and then you could update the elevations and thicknesses,
```python
thickness_change = new_thickness - old_thickness
self.grid.at_node["topographic__elevation"] += thickness_change.reshape(-1)
self._thickness[:] = new_thickness.reshape(-1)
```

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,423 @@
...
+        """
+        Ctop_x = self._topset_threshold[0]
+        Cfore_x = self._foreset_threshold[0]
+        Dtop_x = self._topset_diffusivity[0]
+        Dfore_x = self._foreset_diffusivity[0]
+        Z = self.sea_level
+
+        nx, ny = eta.shape
+        qx = np.zeros((nx + 1, ny))
+        qx[0, :] = influx[0, :] / self.grid.dy
```
mcflugen: The sediment influx is documented as being provided at every node, however you only consider the first row and ignore the rest. Are you able to apply this field to every node?

I might also be useful to validate that values are >= 0.0 and finite.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,423 @@
...
+        if dt is None:
+            dt = stable_dt
+        n_steps = math.ceil(dt / stable_dt)
+        substep_dt = dt / n_steps
+
+        # _advance_substep takes 2D arrays, so reshape the flat internal
+        # state into explicit H_xy, basement_xy parameters.
+        H_xy = self._thickness.reshape(self.grid.shape).T.copy()
+        basement_xy = self._basement.reshape(self.grid.shape).T
+        influx_xy = self.grid.at_node["sediment__influx"].reshape(self.grid.shape).T
```
mcflugen: Please remove the internal (x, y) array convention and the associated transposes. *Landlab* raster fields reshape as (row, column) == (y, x), and retaining that convention throughout will make the code easier to follow and reduce the risk of applying dx, dy, or directional parameters along the wrong axis.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,423 @@
...
+        inherently sequential, but each row is vectorized over grid-y.
+        """
+        Ctop_x = self._topset_threshold[0]
+        Cfore_x = self._foreset_threshold[0]
+        Dtop_x = self._topset_diffusivity[0]
+        Dfore_x = self._foreset_diffusivity[0]
+        Z = self.sea_level
+
+        nx, ny = eta.shape
+        qx = np.zeros((nx + 1, ny))
```
mcflugen: The boundary behavior is hard-coded here, regardless of the grid's boundary conditions.

Please either honor the `RasterModelGrid` boundary configuration or explicitly document and validate the one boundary configuration supported by this implementation. Boundary behavior should not silently disagree with the grid.

mcflugen (changes requested):

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: `sea_level` is read-only now, users write `grid.at_grid["sea_level__elevation"]` directly.

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: `self._basement` isn't a stored attribute anymore, it's derived fresh from `elevation - old_thickness` at the top of every `run_one_step`

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: yeah we did talk about this for other case usages

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
vietbuiminh: construction now raises if the grid has any closed nodes, since this component runs transport on every node including the perimeter and would otherwise silently ignore that. I went with the document and validate

vietbuiminh (commented):
Compact two commits for another review

src/landlab/components/geoenthalpy_delta/geoenthalpy_delta.py
```diff
@@ -0,0 +1,470 @@
...
+            the interval (0, 1].
+        """
+        if not isinstance(grid, RasterModelGrid):
+            raise TypeError(
+                "GeoEnthalpyDelta requires a RasterModelGrid because "
+                "its explicit finite-volume scheme assumes a uniform, "
+                "structured (x, y) grid."
+            )
+
+        if np.any(grid.status_at_node == grid.BC_NODE_IS_CLOSED):
```
mcflugen: This check isn't quite right. Since you're updating topography at all nodes, I think you need to ensure that all the nodes are "core". A helper method might be useful,

```python
def _validate_boundary_conditions(grid):
    if np.any(grid.status_at_node != grid.BC_NODE_IS_CORE):
        raise ValueError(...)
```

Also, this check isn't in the right place. Boundary conditions may not have been finalized when this component is instantiated and boundary conditions are mutable so could change between calls to this component. This check should be done within `run_one_step`.

tests/components/geoenthalpy_delta/test_geoenthalpy_delta.py
```diff
@@ -0,0 +1,378 @@
...
+from numpy.testing import assert_array_equal
+from requireit import ValidationError
+
+from landlab import HexModelGrid
+from landlab import RasterModelGrid
+from landlab.components import GeoEnthalpyDelta
+from landlab.field.errors import FieldError
+
+
+def make_grid(nrows=20, ncols=30, dx=0.5, dy=0.5, sea_level=0.0):
```
mcflugen, line 22: I think we should test the case where `dx != dy`.

mcflugen (changes requested):
@vietbuiminh I think these should be the last round of changes.

tests/components/geoenthalpy_delta/test_geoenthalpy_delta.py
vietbuiminh: This one is interesting because in RasterModelGrid i can only use when dx=dy, because right now the RMG only accepts a float value. Should I out error if dx != dy just for now? This might go into the RMG modification.

vietbuiminh (commented):
Ready for "final" review

mcflugen:
@vietbuiminh This looks ready to merge. Please update the branch with current master and resolve any conflicts so CI can run against the latest base. I plan to squash merge, so there is no need to clean up the intermediate commit history.

mcflugen:
@vietbuiminh One of your tests is missing coverage. In `set_west_boundary_influx`, you don't run the branch when `y_center` is not `None`. You'll need to fix that before merging. If you don't need to test that case, you could remove that keyword argument from `set_west_boundary_influx`.

mcflugen:
>  you could remove that keyword argument from `set_west_boundary_influx`.

@vietbuiminh Thanks for fixing that so quickly. You now just need to update the branch with master and resolve any conflicts (there shouldn't be any).

vietbuiminh:
> > you could remove that keyword argument from `set_west_boundary_influx`.
> 
> @vietbuiminh Thanks for fixing that so quickly. You now just need to update the branch with master and resolve any conflicts (there shouldn't be any).

Perfect, I am/was(still maybe) sleep deprived in Tokyo while waiting for my flight, so i rush myself to get that done before the long flight

mcflugen (approved):
Thanks @vietbuiminh! This is ready to merge.

I'm also going to add you to the Landlab team as a reviewer so that you can do reviews of future pull requests.

Issue #2470, opened 2026-08-27
"Develop your own component or utility" documentation points to unavailable files locations


sdavilao00:
The "Getting your component into the documentation" points to generating our own .rst file, however that file location for a template no longer exists. There is a new way to generate the documentation within Github so this section may be outdated. Thanks!

PR #2475, opened 2026-08-28
Add SoilDepthEvolver component

sdavilao00:
## Pull Request Checklist
- [x] The PR is opened in draft mode
- [ ] Code has been linted and is free of style issues (`nox -s lint`)
- [ ] All tests pass locally (`pytest`, or `nox -s test`)
- [ ] Documentation builds without errors (`nox -s docs-build`)
- [ ] A news fragment describing the change has been added
- [ ] Related issues or discussions are linked in the description
---

##Description

This PR adds a new `SoilDepthEvolver` component for modeling changes in soil depth through depth-dependent soil production and hillslope sediment transport.

`SoilDepthEvolver` is designed to separate soil-depth accounting from the choice of hillslope-transport formulation. Rather than implementing a specific transport law internally, the component accepts an externally supplied Landlab transport component that operates on the same grid. Transport-driven changes in surface elevation are then combined with soil production to update `soil__depth`.

The component explicitly limits transport-driven soil removal by the soil thickness available at the beginning of the timestep. If modeled surface lowering exceeds the available soil thickness, the pre-existing soil is treated as exhausted rather than allowing soil depth to become negative. This provides a relatively simple soil-availability constraint without requiring soil thickness to be incorporated directly into the transport law.

This design allows the soil-depth accounting to be used with different compatible Landlab transport components and on Landlab grids initialized from real, DEM-derived topography. The component itself remains independent of DEM I/O, boundary-condition setup, and application-specific spatial masks.

In preparing this contribution, I recognized overlap with the existing weathering, `DepthDependentDiffuser`, and `DepthDependentTaylorDiffuser` components. I’d appreciate feedback on whether the soil-depth bookkeeping and externally supplied transport approach fills a useful gap, or whether this functionality would be better achieved through composition of existing components.

Tests are included for parameter validation, soil-production calculations, soil-depth updating and erosion behavior, timestep advancement, returned diagnostics, and integration with `TaylorNonLinearDiffuser`.

The `docs/index.toml` file was also updated based on maintainer feedback to correct the component index configuration

## Related Issues

Issue #2470


src/landlab/components/soil_depth_evolver/soil_depth_evolver.py
```diff
@@ -0,0 +1,295 @@
...
+    with an externally supplied Landlab hillslope-transport component.
+    The component tracks changes in soil thickness through time while
+    allowing the user to choose the transport formulation independently.
+
+    Soil is produced according to an exponential soil-production function,
+
+    .. math::
+
+        P = \\frac{\\rho_r}{\\rho_s} P_0
+            \\exp\\left(-\\frac{H}{H_0}\\right),
```
mcflugen, line 22: The *TeX* equation looks good when rendered in HTML, but it's hard to read as text. Consider the following instead,
```
Soil production decreases exponentially with soil depth::

    production_rate = rock_to_soil_density_ratio
                      * maximum_production_rate
                      * exp(-soil_depth / decay_depth)
```

src/landlab/components/soil_depth_evolver/soil_depth_evolver.py
```diff
@@ -0,0 +1,295 @@
...
+    only soil produced during that timestep remains.
+
+    The hillslope-transport component is supplied externally rather than
+    constructed internally. This allows different compatible Landlab
+    transport components to be used without modifying
+    SoilDepthEvolver. The supplied component must operate on
+    the same grid, update ``topographic__elevation``, and provide a
+    ``run_one_step(dt)`` method.
+
+    ## Notes
```
mcflugen, line 50: Use *numpydoc* style for the notes section,
```
Notes
-----
```

src/landlab/components/soil_depth_evolver/soil_depth_evolver.py
```diff
@@ -0,0 +1,295 @@
...
+            "mapping": "node",
+            "doc": "Current soil thickness.",
+        },
+    }
+
+    def __init__(
+        self,
+        grid,
+        *,
+        diffuser,
```
mcflugen, line 117: I think we should have a default of `None` for the diffuser. A value of `None` would mean no sediment transport.

src/landlab/components/soil_depth_evolver/soil_depth_evolver.py
```diff
@@ -0,0 +1,295 @@
...
+            Bulk density of produced soil, in kilograms per cubic meter.
+            Must be positive. Default is 1600 kg/m3.
+        """
+
+        super().__init__(grid)
+
+        soil_production_rate = require_nonnegative(
+            soil_production_rate,
+            name="soil_production_rate",
+        )
```
mcflugen, line 168: You can simplify parameter validation and assignment by combining them,
```python
self._maximum_production_rate = float(
    require_nonnegative(
        soil_production_rate,
        name="soil_production_rate",
    )
)
```

src/landlab/components/soil_depth_evolver/soil_depth_evolver.py
```diff
@@ -0,0 +1,295 @@
...
+        soil_density = require_positive(
+            soil_density,
+            name="soil_density",
+        )
+
+        self._maximum_production_rate = float(soil_production_rate)
+        self._production_decay_depth = float(soil_production_decay_depth)
+
+        self._rock_density = float(rock_density)
+        self._soil_density = float(soil_density)
```
mcflugen, line 189: No need to save these, since they aren't used again.

src/landlab/components/soil_depth_evolver/soil_depth_evolver.py
```diff
@@ -0,0 +1,295 @@
...
+
+        self._density_ratio = self._rock_density / self._soil_density
```
mcflugen, line 191: Maybe the input parameter should be the density ratio rather than rock density and soil density. This would simplify the API. The new parameter could be called `rock_to_soil_density_ratio`, or `density_ratio`, or something else.

src/landlab/components/soil_depth_evolver/soil_depth_evolver.py
```diff
@@ -0,0 +1,295 @@
...
+
+        # Diffuser is created in the driver and passed in.
+        self._diffuser = diffuser
+
+        # Landlab fields
+        self._elevation = self.grid.at_node["topographic__elevation"]
+
+        self._soil_depth = self.grid.at_node["soil__depth"]
```
mcflugen, line 199: Don't cache aliases for field arrays. Another component might change a reference to these fields which would cause these values to be stale. Instead, they shouldn't be class attributes and you should get them every time you run `run_one_step`.

At the top of `run_one_step` you could have,
```python
elevation = self.grid.at_node["topographic__elevation"]
soil_depth = self.grid.at_node["soil__depth"]
```

src/landlab/components/soil_depth_evolver/soil_depth_evolver.py
```diff
@@ -0,0 +1,295 @@
...
+
+        if np.any(self._soil_depth < 0.0):
+            raise ValueError("soil__depth cannot contain negative values.")
```
mcflugen, line 202: This validation should be done in `run_one_step`.

src/landlab/components/soil_depth_evolver/soil_depth_evolver.py
```diff
@@ -0,0 +1,295 @@
...
+
+        self._current_time += dt
+
+        return {
+            "soil_depth": self._soil_depth.copy(),
+            "elevation_change": elevation_change,
+            "soil_depth_change": soil_depth_change,
+            "soil_produced": soil_produced,
+            "production_rate": production_rate,
+        }
```
mcflugen, line 256: `run_one_step` should return `None`. If you want to make these diagnostic arrays available, that should be done through properties. For example,
```python
@property
def elevation_change(self):
    array = self._elevation_change
    array.flags["WRITEABLE"] = False
    return array
```
You probably want it to be read-only so that someone can't change array values. You wouldn't want to do this for `soil_depth` since that's available as a field.

mcflugen (changes requested):
@sdavilao00 Thank you for this contribution! I've made some initial comments.

src/landlab/components/soil_depth_evolver/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .soil_depth_evolver import SoilDepthEvolver
+
+__all__ = ["SoilDepthEvolver"]
```
mcflugen, line 3: If possible, I would like to see this file empty and then have `landlab.components` import the `SoilDepthEvolver` directly from `soil_depth_evolver.py` module.

